PR #129, opened 2026-01-28
fix: Create channel file eagerly for tailf compatibility

btucker:
## Summary
- Creates empty `channel.jsonl` file on `Channel::new()` if it doesn't exist
- Fixes tailf failing silently when channel file hasn't been created yet
- Adds test to verify channel file exists for tailf compatibility

## Root cause
The `tailf` crate wraps `tail -f`, which fails when the file doesn't exist. Previously, `Channel::new()` only created the directory structure, and the channel file was created lazily on first `send()`. This caused tailf to fail silently (error swallowed by `.ok()`) and fall back to polling.

## Test plan
- [x] Added `test_channel_file_exists_for_tailf` test
- [x] All 213 tests pass
- [x] Verified fix by confirming test fails before fix and passes after

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -67,13 +67,22 @@ pub struct Channel {
 impl Channel {
     /// Create a new channel at the specified base directory
     ///
-    /// Creates the directory structure if it doesn't exist.
+    /// Creates the directory structure and channel file if they don't exist.
+    /// The channel file is created eagerly so that file watchers (tailf) can
+    /// immediately start monitoring it.
     pub fn new(base_dir: impl Into<PathBuf>) -> Result<Self> {
         let base_dir = base_dir.into();
         fs::create_dir_all(&base_dir)?;
         fs::create_dir_all(base_dir.join("cursors"))?;
 
         let channel_file = base_dir.join("channel.jsonl");
+
+        // Create the channel file if it doesn't exist.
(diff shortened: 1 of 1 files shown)


btucker (commented):
## Code Review

**Overall**: This is a well-reasoned fix. The root cause analysis is accurate - `tail -f` requires the file to exist before it can watch it. The change is minimal and focused.

### Concern: TOCTOU race condition

The current implementation has a time-of-check-to-time-of-use (TOCTOU) race:

```rust
if !channel_file.exists() {
    File::create(&channel_file)?;
}
```

`File::create()` truncates existing files. In a race between two processes:
1. Process A: checks, file doesn't exist
2. Process B: checks, file doesn't exist
3. Process B: creates empty file
4. Process B: writes message "hello"
5. Process A: creates (truncates!) file - message "hello" is **lost**

**Suggested fix** - use `OpenOptions` to atomically create without truncating:

```rust
use std::fs::OpenOptions;

OpenOptions::new()
    .create(true)
    .write(true)
    .open(&channel_file)?;
```

This creates the file if it doesn't exist but leaves it untouched if it does.

### Everything else looks good
- Doc comment update is clear
- Tests are appropriate
- Root cause analysis in PR description is excellent

Minor fix needed, then LGTM! 👍

PR #132, opened 2026-01-28
fix: Use list-windows instead of has-session in window_exists

btucker:
## Summary
- Fixed `window_exists` to actually check for the window, not just the session
- The bug was that `tmux has-session -t session:window` only verifies the SESSION exists, not the WINDOW
- Now uses `tmux list-windows` to properly check if the window is present
- Also handles windows with status suffixes (e.g., "lexington:dev#3")

## Root Cause
The daemon logs showed "Respawned coworker successfully" but then "Failed to nudge X to resume: Failed to send keys to tmux window". This was because:
1. `spawn_claude` creates a window with a command
2. If the command fails, the window immediately dies
3. `window_exists` was using `has-session` which only checks if the SESSION exists (always true since Lead uses it)
4. So `spawn_claude` reported success even when the window was gone

## Test plan
- [x] All 129 existing tests pass
- [ ] Manual test: daemon correctly detects failed window creation

### DIFF
--- a/src/tmux.rs
+++ b/src/tmux.rs
@@ -536,15 +536,31 @@ pub fn list_windows(session: &str) -> crate::Result<Vec<String>> {
 }
 
 /// Check if a window exists in the session.
+///
+/// Note: Window names may include a status suffix (e.g., "lexington:dev#3"),
+/// so this checks if any window's base name matches the given name.
 pub fn window_exists(session: &str, name: &str) -> crate::Result<bool> {
-    let target = format!("{}:{}", session, name);
-
-    let status = Command::new("tmux")
-        .args(["has-session", "-t", &target])
-        .status()
+    // Use list-windows to get actual window names
+    // has-session only checks if the SESSION exists, not the WINDOW
+    let output = Command::new("tmux")
+        .args(["list-windows", "-t", session, "-F", "#{window_name}"])
+        .output()
         .map_err(Error::Io)?;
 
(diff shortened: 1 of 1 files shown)


btucker:
## Code Review by lexington

**Overall: LGTM** ✅

### The Bug
The root cause analysis is spot-on. `tmux has-session -t session:window` only validates the *session* exists, not the *window*. Since the Lead always uses the session, this was always returning true even when coworker windows had died.

### The Fix
Using `list-windows` with `-F #{window_name}` is the correct approach - it actually enumerates the windows in the session.

### Minor Observations

1. **Case-insensitive comparison**: The code lowercases both names for comparison. This is more lenient than tmux's default case-sensitive matching. Probably fine given coworker names are always lowercase, but worth noting as a behavioral change.

2. **Status suffix parsing**: Splitting on `:` to handle suffixes like `lexington:dev#3` makes sense. Just want to confirm tmux actually uses this format for status suffixes (vs some other delimiter).

3. **No test code**: The comment explaining the bug scenario is helpful. Integration tests for tmux are hard to write, so the comment documenting the fix is a reasonable compromise.

### Verified
- Logic correctly handles session-not-found case (returns `Ok(false)`)
- Error propagation via `map_err(Error::Io)?` is consistent with the rest of the codebase
- Usage sites in `coworker.rs:155` and `tmux.rs:705` will benefit from this fix

Good catch and clean fix!

PR #141, opened 2026-01-28
feat: Add e2e tests for daemon startup and API endpoints

btucker:
## Summary
- Adds `tests/daemon_e2e.rs` with 10 smoke tests for daemon lifecycle
- Tests cover: startup, socket creation, ping/version/status/coworker.list/channel.post endpoints
- Tests verify error handling for unknown methods
- Tests verify multiple sequential requests on same connection
- Tests verify graceful shutdown via SIGTERM
- Tests verify PID file creation

Uses an isolated test fixture with a fake git repo to avoid affecting real daemon instances.

## Test plan
- [x] All 10 tests pass with `cargo test --test daemon_e2e -- --include-ignored`

### DIFF
--- a/tests/daemon_e2e.rs
+++ b/tests/daemon_e2e.rs
@@ -251,10 +251,7 @@ impl Drop for DaemonFixture {
 fn test_daemon_starts_and_creates_socket() {
     let mut fixture = match DaemonFixture::new() {
         Some(f) => f,
-        None => {
-            eprintln!("Skipping test: could not create test fixture");
-            return;
-        }
+        None => return, // Skip silently if fixture creation fails
     };
 
     assert!(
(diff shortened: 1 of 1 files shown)


btucker:
Thanks for the review @vernon!

Addressed the feedback:

1. **Quieter logging** - Removed the `eprintln!("Skipping test: ...")` messages. Tests now return silently with inline comments explaining why. The actual error messages in `start_daemon()` (build failures, spawn failures) remain since those indicate real infrastructure problems.

2. **Test timeout** - Created follow-up task #17 to add global test timeouts. This is a good enhancement but would benefit from a bit more design consideration (e.g., should it use `#[timeout]` attribute, a wrapper, or process-level timeout).

PR should be ready to merge now!

PR #160, opened 2026-01-28
feat: Add nudge verification to detect and retry failed Enter

btucker:
## Summary
- Adds verification after sending nudge to detect if Enter key failed
- Captures pane content with `tmux capture-pane` after each Enter attempt
- Detects stuck nudges by looking for text after the prompt symbol (❯)
- Retries Enter up to 3 times if nudge appears stuck
- Adds debug logging when retries are needed

This catches cases where Enter doesn't work (timing issues, race conditions) and automatically retries, improving nudge reliability.

## Implementation
- `capture_pane()`: Captures current tmux pane content
- `is_nudge_stuck()`: Detects nudge text sitting after prompt symbol
- Modified `send_keys()` to verify after each Enter attempt

## Test plan
- [x] 5 new unit tests for `is_nudge_stuck()` detection logic
- [x] All existing tmux tests pass (28 total)
- [x] clippy and fmt pass
- [ ] Manual test: nudge a coworker and verify messages submit

### DIFF
--- a/src/tmux.rs
+++ b/src/tmux.rs
@@ -392,6 +392,50 @@ pub fn kill_window(session: &str, name: &str) -> crate::Result<()> {
     Ok(())
 }
 
+/// Capture the current content of a tmux pane.
+///
+/// Returns the pane content as a string, or None if capture fails.
+fn capture_pane(target: &str) -> Option<String> {
+    let output = Command::new("tmux")
+        .args(["capture-pane", "-t", target, "-p"])
+        .output()
+        .ok()?;
+
+    if !output.status.success() {
+        return None;
+    }
+
+    Some(String::from_utf8_lossy(&output.stdout).to_string())
+}
+
+/// Check if the nudge text is still sitting in the input line (not submitted).
+///
+/// Returns true if the nudge appears to be stuck (Enter didn't work).
+/// Looks for the text after the prompt symbol (❯) on any recent line.
+fn is_nudge_stuck(pane_content: &str, nudge_text: &str) -> bool {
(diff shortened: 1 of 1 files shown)


btucker:
## Code Review by madison

### Summary
This PR adds nudge verification to detect and retry when Enter key fails. Good improvement to nudge reliability.

### Analysis

**`capture_pane()` function**
- Simple and correct - uses `tmux capture-pane -t <target> -p` to get pane content
- Returns Option<String> to handle capture failures gracefully

**`is_nudge_stuck()` function**
- Checks if nudge text appears after the ❯ prompt symbol on recent lines
- Looks at last 5 lines to handle cases where prompt isn't the very last line
- Uses first 20 characters of long messages to avoid line wrapping false negatives
- This is a reasonable heuristic for detecting stuck nudges

**Modified `send_keys()` flow**
- After sending Enter, waits 100ms then verifies the nudge was submitted
- If text is stuck after prompt, retries Enter (up to 3 times total)
- Falls back to assuming success if pane capture fails
- Returns an error after all retries exhausted

**Test coverage**
- 5 unit tests covering: detection, submitted text, empty prompt, no prompt, partial match
- All tests exercise the `is_nudge_stuck()` function correctly

### Minor considerations

1. The 20-character prefix match is a pragmatic choice. Edge case: if the first 20 chars of two different nudges are identical, it could match the wrong one. In practice this is unlikely.

2. The 100ms wait after Enter before verification adds slight latency but is reasonable for reliability.

Overall this is a clean implementation that addresses a real reliability issue. ✅ LGTM

PR #164, opened 2026-01-28
fix: Pass repo_name to get_busy_coworkers in daemon context

btucker:
## Summary
- Fixed idle shutdown incorrectly killing coworkers with active tasks
- Added `get_busy_coworkers_for_repo(repo_name)` to tasks.rs that explicitly accepts the repo name
- Updated daemon's `get_busy_coworkers()` to accept repo_name parameter and use the new function
- Both call sites (idle shutdown check and reviewer assignment) now pass `state.repo_name`

## Root Cause
The daemon was calling `get_busy_coworkers()` which internally relied on `detect_repo_name()` to find the current repository. However, from the daemon's async context, `detect_repo_name()` may not work correctly since it runs git commands that depend on the current working directory.

The daemon already stores `repo_name` in `DaemonState`, so we just needed to pass it through to the task lookup functions.

## Test plan
- [x] All unit tests pass
- [x] Code compiles without warnings
- [x] Clippy passes

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -733,7 +733,7 @@ async fn check_and_shutdown_idle_coworkers(state: &DaemonState) {
     }
 
     // Get in_progress tasks to determine who is busy
-    let busy_coworkers = get_busy_coworkers();
+    let busy_coworkers = get_busy_coworkers(&state.repo_name);
 
     // Get coworkers with open PRs - they should NEVER be auto-killed
     let coworkers_with_open_prs = get_coworkers_with_open_prs();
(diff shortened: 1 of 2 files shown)


btucker:
## Code Review by broadway



### Summary
This is a well-structured fix for a bug where idle shutdown was incorrectly killing coworkers with active tasks.

### Analysis

**Root Cause**: The daemon calls `get_busy_coworkers()` from an async context where the current working directory may not be the repo directory. The existing function relied on `detect_repo_name()` which runs git commands that depend on the working directory.

**Solution**: The fix correctly:
1. Adds `get_busy_coworkers_for_repo(repo_name)` that takes an explicit repo name parameter
2. Updates the daemon's wrapper to accept and pass through `repo_name`
3. Updates both call sites (`check_and_shutdown_idle_coworkers` and `find_available_reviewer`) to pass `state.repo_name`

### What I like
- **Minimal and focused** - only changes what's needed to fix the bug
- **Non-breaking** - keeps the original `get_busy_coworkers()` for CLI contexts where `detect_repo_name()` works correctly
- **Good documentation** - the doc comment on the new function clearly explains when to use it
- **Complete** - both call sites in daemon.rs are updated

### Verification
- The signature of `read_tasks_for_repo(repo_name: Option<&str>)` correctly uses the provided repo name when `Some(repo_name)` is passed, falling back to detection only when `None` is passed

**LGTM** ✓ - Ready to merge

PR #165, opened 2026-01-28
fix: Pass repo_name to get_busy_coworkers in daemon

btucker:
## Summary
- Fixed idle shutdown incorrectly killing coworkers who had in_progress tasks
- Root cause: `get_busy_coworkers()` called `detect_repo_name()` which runs git commands that may fail in daemon's background process context
- Solution: Pass the repo_name from `DaemonState` (already known) through to task reading functions

## Changes
- Add `get_busy_coworkers_for_repo(repo: &str)` in tasks.rs that takes explicit repo name
- Update daemon's `get_busy_coworkers()` wrapper to accept repo_name parameter
- Update both call sites (idle check, reviewer finder) to pass `state.repo_name`
- Add test for busy coworkers filtering logic

## Test plan
- [x] All existing tests pass (282 tests)
- [x] New test verifies filtering logic for busy coworkers

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -733,7 +733,7 @@ async fn check_and_shutdown_idle_coworkers(state: &DaemonState) {
     }
 
     // Get in_progress tasks to determine who is busy
-    let busy_coworkers = get_busy_coworkers();
+    let busy_coworkers = get_busy_coworkers(&state.repo_name);
 
     // Get coworkers with open PRs - they should NEVER be auto-killed
     let coworkers_with_open_prs = get_coworkers_with_open_prs();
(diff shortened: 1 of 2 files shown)


btucker:
## Code Review by amsterdam



**Verdict: LGTM** ✓

This is a clean, well-targeted fix for the idle shutdown bug where coworkers with in_progress tasks were getting killed.

### What the PR does well

1. **Root cause addressed correctly** - The issue was `detect_repo_name()` failing in daemon background processes. Passing `repo_name` explicitly from `DaemonState` is the right solution.

2. **Clean API design** - The new `get_busy_coworkers_for_repo(repo: &str)` function has a clear purpose and good documentation explaining why it exists.

3. **Both call sites updated** - Both the idle coworker check (line 736) and reviewer finder (line 1623) now pass `state.repo_name`.

4. **Test coverage added** - The new test verifies the filtering logic for busy coworkers.

### Minor observations (non-blocking)

- The test duplicates the filtering logic (`filter(|t| t.status == TaskStatus::InProgress).filter_map(|t| t.owner)`) rather than calling `get_busy_coworkers_for_repo()` directly. This is acceptable for testing the logic in isolation, but a follow-up could add an integration test that calls the actual function.

- The original `get_busy_coworkers()` function in tasks.rs is preserved but no longer used by daemon. This is fine for backwards compatibility, though if nothing else uses it, it could eventually be removed.

Good fix!

PR #167, opened 2026-01-28
fix: Read enabledPlugins instead of plugins from settings.json

btucker:
## Summary
- Fixed `read_user_plugins()` in `src/tmux.rs` to read the correct key from user's settings
- Changed `settings.get("plugins")` to `settings.get("enabledPlugins")`

## Root Cause
The user's `~/.claude/settings.json` uses `enabledPlugins` as the key for configured plugins, but the code was reading `plugins`. This prevented coworkers from receiving the user's configured plugins.

## Test plan
- [x] Code compiles without warnings
- [x] Clippy passes
- [x] All 171 unit tests pass

### DIFF
--- a/src/tmux.rs
+++ b/src/tmux.rs
@@ -660,7 +660,7 @@ fn read_user_plugins() -> Option<serde_json::Value> {
         .join("settings.json");
     let content = std::fs::read_to_string(settings_path).ok()?;
     let settings: serde_json::Value = serde_json::from_str(&content).ok()?;
-    settings.get("plugins").cloned()
+    settings.get("enabledPlugins").cloned()
 }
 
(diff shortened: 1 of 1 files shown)


btucker:
## Code Review by riverside



### Summary
Simple 1-line fix to read the correct settings key for user plugins.

### Analysis
- **Issue**: Code was reading `plugins` key but `~/.claude/settings.json` uses `enabledPlugins`
- **Fix**: Change `settings.get("plugins")` to `settings.get("enabledPlugins")`

### Verification
Task #53 "Fix plugin replication to read enabledPlugins instead of plugins" confirms this is the correct key name.

### Result
**LGTM** ✓ - Straightforward fix, ready to merge.

PR #168, opened 2026-01-28
fix: Atomically assign task ownership before spawning coworker

btucker:
## Summary
- Fixes race condition where multiple coworkers could claim the same task (#51 was worked on by both lexington and park simultaneously)
- Now assigns task ownership BEFORE spawning the coworker, not after
- Makes `next_available_name()` public so daemon can get a name before spawning

## Changes
- **coworker.rs**: Renamed `next_name()` to `next_available_name()` and made it public with documentation
- **daemon.rs**: Reordered `spawn_for_pending_tasks()` Case 2 to:
  1. Get available name first
  2. Atomically assign task ownership
  3. Spawn coworker with pre-assigned name

## Test plan
- [x] All existing tests pass
- [x] Code compiles with no clippy warnings
- [ ] Manual verification: spawn multiple coworkers with pending tasks to confirm no duplicates

### DIFF
--- a/src/coworker.rs
+++ b/src/coworker.rs
@@ -100,7 +100,10 @@ impl CoworkerManager {
     ///
     /// Randomly selects from available primary avenue names first.
     /// Falls back to overflow names only when all primary names are in use.
-    fn next_name(&self) -> Option<String> {
+    ///
+    /// This can be used to get a name before spawning, e.g., to assign
+    /// task ownership atomically before the coworker starts.
+    pub fn next_available_name(&self) -> Option<String> {
         let coworkers = self.coworkers.read().unwrap();
 
         // Collect available primary avenue names
@@ -142,10 +145,12 @@ impl CoworkerManager {
     /// If `resume` is true, passes `--continue` to claude to resume the previous
     /// session from this worktree (useful for recovering orphaned tasks).
     pub fn spawn(&self, resume: bool) -> crate::Result<String> {
-        let name = self.next_name().ok_or_else(|| crate::Error::Rpc {
-            code: -32603,
-            message: "No available coworker slots (all avenue names in use)".to_string(),
-        })?;
+        let name = self
+            .next_available_name()
+            .ok_or_else(|| crate::Error::Rpc {
+                code: -32603,
+                message: "No available coworker slots (all avenue names in use)".to_string(),
+            })?;
 
         // Try to create an isolated worktree for this coworker
         let worktree_path = match self.worktree_manager.create(&name) {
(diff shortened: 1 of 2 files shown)


btucker:
## Code Review by broadway



### Summary
This PR fixes a race condition where multiple coworkers could claim the same task, as seen with task #51 being worked on by both lexington and park simultaneously.

### Analysis

**Root Cause**: The previous order was `spawn()` → `update_task_owner()`, meaning the coworker could start reading TaskList before ownership was assigned.

**Solution**: Reorder to assign ownership atomically before spawning:
1. `next_available_name()` - get the name first
2. `update_task_owner()` - assign ownership in task storage
3. `spawn_with_name()` - then start the coworker

### What I like
- **Correct fix** - the root cause (race between spawn and ownership assignment) is addressed
- **Clean refactor** - `next_name()` → `next_available_name()` with `pub` visibility is well-documented
- **Proper error handling** - if spawn fails after assignment, the task is still owned and will be retried on next tick
- **Updated log messages** - reflect the new "pre-spawn" assignment flow
- **Test renames** - keep tests consistent with the new function name

### One consideration
If spawn fails after task assignment, the task remains assigned to a coworker that doesn't exist. The current handling ("next daemon tick will see the assigned task and try to spawn again") seems reasonable, but might want to eventually clear the owner on spawn failure if the coworker never comes up. Not a blocker for this PR.

**LGTM** ✓ - Ready to merge

btucker:
## Code Review by vernon



### Summary
This PR fixes a race condition where multiple coworkers could claim the same task. The issue occurred because the old flow was `spawn() → update_task_owner()`, meaning a coworker could start reading TaskList before its ownership was recorded.

### Changes Reviewed
1. **coworker.rs**: Made `next_available_name()` public so daemon can reserve a name before spawning
2. **daemon.rs**: Reordered Case 2 in `spawn_for_pending_tasks()`:
   - Before: `spawn()` → `update_task_owner()`
   - After: `next_available_name()` → `update_task_owner()` → `spawn_with_name()`

### Analysis

**The fix is correct.** By assigning ownership BEFORE spawning, the task is claimed in storage before the coworker can even start. This eliminates the race window.

**Failure recovery is also correct.** If `spawn_with_name()` fails after `update_task_owner()` succeeds:
- Task becomes owned but coworker doesn't exist
- Case 1 ("pending tasks with owners but coworker not running") will pick this up on the next daemon tick and retry the spawn

**No race between Case 1 and Case 2.** They run sequentially, and `next_available_name()` checks the coworkers map which is updated during spawns.

### Minor Observations

1. The log message in the spawn failure case uses `warn!` level but the comment says "that's okay" - slight tone mismatch. This is cosmetic.

2. The test renames (`test_next_name_*` → `test_next_available_name_*`) are good - tests should match function names.

### Verdict

**Approve** - The logic is sound and the fix correctly addresses the race condition. Good work on the clear commit message explaining the before/after flow.

PR #174, opened 2026-01-28
fix: Use enabledPlugins key in coworker settings

btucker:
## Summary
- Changed the settings key from `"plugins"` to `"enabledPlugins"` in `coworker_settings_json()` so Claude Code actually reads the plugins

This fixes an issue where plugins weren't being loaded for coworkers because the wrong key name was being used.

## Test plan
- [x] Existing tests pass

### DIFF
--- a/src/tmux.rs
+++ b/src/tmux.rs
@@ -676,7 +676,7 @@ fn coworker_settings_json(bin_command: &str) -> serde_json::Value {
 
     serde_json::json!({
         "editorMode": "normal",
-        "plugins": user_plugins,
+        "enabledPlugins": user_plugins,
         "hooks": {
             "Stop": [{
                 "hooks": [{


btucker:
## Code Review by prince



**Verdict: Approve** ✅

### Summary
This is a clean, minimal fix that completes the work started in PR #167. The change from `"plugins"` to `"enabledPlugins"` ensures consistency between how we read user plugins and how we write coworker settings.

### Analysis

**The Fix:**
- Changes the settings key from `"plugins"` to `"enabledPlugins"` in `coworker_settings_json()` at line 679

**Why This is Correct:**
1. PR #167 fixed `read_user_plugins()` to read from `settings.get("enabledPlugins")` (line 663)
2. Claude Code expects the `enabledPlugins` key in `settings.json`, not `plugins`
3. This PR aligns the write side to match the read side, so coworkers get their plugins loaded correctly

**Testing:**
- Existing tests pass (they don't check the specific key name, but verify the structure)
- The fix is purely a key rename with no logic changes

### Minor Suggestion (non-blocking)
The tests for `coworker_settings_json` don't verify the plugins key name. A future enhancement could add an assertion like:
```rust
assert!(settings.get("enabledPlugins").is_some());
```
But this is not necessary for this PR.

LGTM - ready to merge.

PR #249, opened 2026-01-29
feat: Add TUI chat input and rename mobile sender to user

btucker:
## Summary
- Added a text input box at the bottom of the chat TUI — press `i` to type, Enter to send, Esc to cancel (vim-like UX)
- Renamed "mobile" sender to "user" in web.rs and daemon.rs since messages can now come from either the web app or the TUI
- Input box shows cursor position and visual focus state (cyan border when active)

## Test plan
- [ ] `cargo test` passes (67 bin tests + 15 doc tests)
- [ ] `cargo clippy` and `cargo fmt` pass
- [ ] Open the chat TUI — input box appears at bottom with "Press 'i' to type"
- [ ] Press `i` — border turns cyan, cursor appears, title shows key hints
- [ ] Type a message and press Enter — message posts to channel as "user"
- [ ] Press Esc — exits input mode without sending
- [ ] Arrow keys, Home, End, Backspace, Delete work for cursor movement
- [ ] Web app messages now show as "user" instead of "mobile"
- [ ] Daemon nudges Lead when "user" messages arrive

### DIFF
--- a/src/bin/midtown/cli/chat/app.rs
+++ b/src/bin/midtown/cli/chat/app.rs
@@ -115,6 +115,12 @@ pub struct App {
     repo_status_receiver: Option<Receiver<RepoStatus>>,
     /// Selection mode - when true, mouse capture is disabled for text selection
     pub selection_mode: bool,
+    /// Input mode - when true, keyboard input goes to the text input
+    pub input_mode: bool,
+    /// Current text in the input field
+    pub input_text: String,
+    /// Cursor position within input_text
+    pub input_cursor: usize,
 }
 
 /// Interval between kanban data refreshes (30 seconds)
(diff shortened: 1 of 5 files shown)


btucker:
## Code Review by canal



### Code review

Found 1 issue:

1. **Cursor position does not account for multi-byte or wide Unicode characters.** The input cursor is positioned using `app.input_cursor as u16` directly as a column offset, which assumes every character occupies exactly one terminal column. Emoji, CJK characters, and other wide Unicode glyphs occupy 2+ columns, so the visual cursor will desync from the actual edit position when such characters are present. A crate like `unicode-width` would be needed to compute the correct display width.

https://github.com/btucker/midtown/blob/a66e7e3bf8efeff9a25f271f5ae41597aacfa99f/src/bin/midtown/cli/chat/ui.rs#L731-L735

PR #258, opened 2026-01-29
fix: Route TUI input through daemon to enable Lead nudge

btucker:
## Summary
- Route TUI chat input through daemon RPC instead of direct channel write
- Add `channel_post_as()` method to DaemonClient for explicit sender control
- Fall back to direct write if daemon is unavailable

## Context
The daemon already had nudge logic for `from=="user"` messages (added in e5621b3 for mobile web UI), but the TUI was bypassing it by writing directly to `channel.jsonl`.

This change makes TUI input go through the same path as mobile, so the Lead gets nudged about new user messages.

## Test plan
- [x] Build passes
- [x] All tests pass
- [x] Verified daemon nudge code handles `from=="user"` at daemon.rs:2295-2302

### DIFF
--- a/src/bin/midtown/cli/chat/app.rs
+++ b/src/bin/midtown/cli/chat/app.rs
@@ -7,6 +7,8 @@ use std::time::{Duration, Instant};
 use chrono::{DateTime, Utc};
 use midtown::{Channel, Message};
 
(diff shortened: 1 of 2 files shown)


btucker:
### Code review

PR #282, opened 2026-01-29
fix: Use shared task list path in reset_task_to_pending_for_repo

btucker:
## Summary
- Fix `reset_task_to_pending_for_repo()` to use `task_list_id_for_repo()` (returns `midtown-<repo>`) instead of reading the Lead's session UUID. The old code wrote to `~/.claude/tasks/<UUID>/` while tasks actually live in `~/.claude/tasks/midtown-<repo>/`.
- This path mismatch caused orphan task recovery to fail silently — tasks would stay stuck `in_progress` with a dead owner because the reset wrote to the wrong directory.
- Investigation for task #556 also found that "task ownership not showing in Claude Code UI" is a Claude Code product limitation — the built-in todo panel doesn't render the `owner` field. Task data on disk is correct.

## Test plan
- [x] Added `test_reset_task_uses_shared_task_list_path` verifying the path convention
- [x] All 221 lib tests pass, clippy clean, fmt clean

### DIFF
--- a/src/tasks.rs
+++ b/src/tasks.rs
@@ -407,22 +407,15 @@ pub fn reset_task_to_pending_for_repo(task_id: &str, repo_name: &str) -> Result<
         return Err("Could not determine home directory".to_string());
     };
 
-    // Get the lead session ID
-    let lead_session_file = crate::paths::lead_session_file_for_repo(repo_name);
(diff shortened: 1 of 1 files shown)


btucker:
### Code review

PR #312, opened 2026-01-29
fix: Deduplicate orphaned worktree warnings and detect squash-merged PRs

btucker:
## Summary
- **Deduplication**: Orphaned worktree warnings no longer repeat every 5-second poll cycle. A `warned_orphans` HashSet tracks already-warned names, resetting on daemon restart.
- **@lead notification**: Warnings now @mention lead and consolidate all orphan names into a single message instead of per-worktree messages.
- **Squash-merge detection**: New `is_branch_pr_merged()` method uses `gh pr list --state merged` to detect when a branch's PR was squash-merged. This prevents false "unmerged commits" warnings (since squash-merge creates different SHAs on main vs the branch).

## Test plan
- [x] All 300 existing tests pass
- [x] `cargo clippy` clean
- [x] `cargo fmt` clean
- [x] Verified all 5 "orphaned" worktrees (riverside, prince, spring, bleecker, houston) have merged PRs on GitHub, confirming the squash-merge false positive diagnosis

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -4123,6 +4123,15 @@ fn get_in_progress_tasks_with_owners() -> Vec<(String, String, String)> {
 fn cleanup_orphaned_worktrees(state: &DaemonState) {
     let flagged = state.coworkers.cleanup_orphaned_worktrees();
 
+    // Prune warned_orphans: remove entries for worktrees that are no longer
+    // flagged (they were cleaned up or manually deleted). This ensures that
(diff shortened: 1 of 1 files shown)


btucker:
Good catch riverside — addressed in f974846. The `warned_orphans` set is now pruned each poll cycle: entries for worktrees that are no longer flagged (cleaned up or manually deleted) are removed via `retain`. This ensures reused coworker names trigger new warnings if they become orphaned again.

PR #321, opened 2026-01-29
fix: route user @mentions directly to coworkers as nudges

btucker:
## Summary
- Fixed bug where user @mentions of coworkers were only nudging the lead instead of the mentioned coworker
- Moved `route_mentions` call into `handle_channel_post` for user messages, so both the RPC and mobile message paths route mentions correctly
- Removed the now-redundant external `route_mentions` call from the mobile_rx handler
- Added test for user @mention extraction behavior

## Test plan
- [x] New unit test `test_user_mentions_coworker_should_be_extracted` passes
- [x] All 314 unit tests pass
- [x] Clippy clean, `cargo fmt` clean

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -1574,7 +1574,7 @@ async fn pr_poll_task(
 // ============================================================================
 
 /// Senders to skip when routing mentions (loop protection).
-const SKIP_SENDERS: &[&str] = &["midtown", "system", "github"];
+const SKIP_SENDERS: &[&str] = &["midtown", "system", "github", "user"];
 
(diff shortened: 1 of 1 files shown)


btucker:
Good catch york — elementary oversight on my part! The `chat_monitor_loop` would indeed double-route user @mentions since `"user"` wasn't in `SKIP_SENDERS`.

Fixed in 511df80: added `"user"` to `SKIP_SENDERS`, mirroring the pattern used for `"github"` where the handler routes mentions directly and the chat monitor skips them. Updated the test too.

PR #344, opened 2026-01-30
fix: prevent reviewers from being sent on break due to expired PR assignment tracking

btucker:
## Summary
- Reviewers were being sent on break with "no PR assignment found" because PR assignment tracking expired after 10 minutes even when the reviewer coworker was still actively running
- Added `cleanup_preserving()` methods to both `PrReviewTracker` and `GitHubState` that skip removal of assignments for active coworkers and refresh their timestamps
- Updated idle checker to consult both in-memory and persistent state for reviewer assignments, surviving daemon restarts

## Root cause
Two issues combined to cause the bug:
1. **Assignment timeout expiry**: Both in-memory and persistent trackers clean up assignments older than `PR_REVIEW_ASSIGNMENT_TIMEOUT_SECS` (10 min). Code reviews often take longer. Once expired, the reviewer lost idle-check protection AND the shutdown path couldn't find their PR assignment.
2. **Missing persistent state check**: The idle checker only checked in-memory `active_reviewers()`. After a daemon restart, the in-memory tracker is empty, so all surviving reviewers immediately lost protection — even though the persistent `github_state.json` still had their assignments.

## Changes
- `PrReviewTracker::cleanup_preserving(active_coworkers)` — retains + refreshes expired assignments for still-running coworkers
- `GitHubState::cleanup_expired_preserving(active_coworkers)` — same for persistent state
- `poll_prs_for_issues` — passes active coworker names to both cleanup methods
- `check_and_shutdown_idle_coworkers` — merges persistent state reviewers into `active_reviewers` set

## Test plan
- [x] 4 new unit tests covering preserving and removal behavior for both trackers
- [x] All 327 existing tests pass
- [x] Clippy + fmt clean

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -418,6 +418,29 @@ impl PrReviewTracker {
         self.assigned.retain(|_, (_, t)| t.elapsed() < timeout);
     }
 
+    /// Clean up stale assignments, but preserve assignments for active coworkers.
+    ///
+    /// This prevents reviewers from losing their PR assignment tracking while
+    /// they are still actively running (e.g., a review taking longer than the
+    /// timeout). Assignments for inactive coworkers are cleaned up normally.
+    /// Active coworkers' assignments are refreshed so timeout-based lookups
+    /// (active_reviewers, pr_for_coworker) continue to work.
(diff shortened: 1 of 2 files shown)


btucker:
### Code review

PR #347, opened 2026-01-30
feat: randomize daemon spawn/shutdown channel messages

btucker:
## Summary
- Extracts all hardcoded daemon channel messages (6 spawn call-ins, 3 break notifications) into a new `daemon_messages` module
- Each event type has a pool of 5 template variants with randomized selection via `fastrand`
- Normal personality always uses the canonical first template; Fun and Wild pick randomly from the full pool
- Adds comprehensive unit tests verifying determinism for Normal and valid output for Fun/Wild

## Test plan
- [x] `cargo test daemon_messages` — all 3 tests pass (normal determinism, fun validity, wild validity)
- [x] `cargo test` — full suite passes (335+ tests)
- [x] `cargo clippy` — clean
- [x] `cargo fmt` — clean

I don't review code that arrives without a test suite. This isn't some side street.

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -2058,7 +2058,7 @@ fn route_mentions(state: &DaemonState, msg: &Message) {
                     // Post to channel about the call-in
                     let spawn_msg = Message::text(
                         "midtown",
-                        daemon_messages::called_in_mention(&name, config::get_personality()),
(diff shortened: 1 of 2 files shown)


btucker:
This all seems overly complicated. I don't think we need so many different lists. Can we keep this simpler?

PR #367, opened 2026-01-30
feat: add personality to coworker idle status messages

btucker:
## Summary
- Added `idle_waiting()` function to `daemon_messages.rs` with personality-flavored idle message pool (all variants contain the required `waiting` keyword for daemon status parsing)
- Updated idle hook in `hooks.rs` to load personality config and use the new pool instead of hardcoded `"waiting for input"`
- Strengthened coworker prompt (`agents/coworker.md`) with explicit DO/DON'T examples showing that status messages should express personality

## Test plan
- [x] All 456 library tests pass
- [x] New `idle_waiting_always_contains_waiting_keyword` test verifies all personality variants include `waiting` keyword (50 iterations per variant)
- [x] Existing daemon_messages tests extended for idle_waiting in Normal/Fun/Wild paths
- [x] Clippy clean, fmt clean

### DIFF
--- a/src/bin/midtown/cli/hooks.rs
+++ b/src/bin/midtown/cli/hooks.rs
@@ -181,7 +181,7 @@ fn handle_idle_hook() -> Result<Response, String> {
     let agent = std::env::var("MIDTOWN_AGENT").unwrap_or_else(|_| "coworker".to_string());
     let personality = midtown::config::get_personality();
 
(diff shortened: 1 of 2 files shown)


btucker:
Fixed in 70c83bc. Removed the `name` parameter and the no-op `.replace()` — the idle text is used as `Message::action()` content where the sender name is already in the `from` field, so embedding it in the message body would be redundant.

PR #391, opened 2026-01-30
feat: trigger immediate coworker spawn on task creation

btucker:
## Summary
- Detects "created task" messages in channel posts and immediately calls `spawn_for_pending_tasks`, eliminating the 0-5 second delay from waiting for the next orphan check interval
- Two detection points: `handle_channel_post` (direct RPC path from lead CLI) and `chat_monitor_loop` (file-tail path, catches all sources including system messages)
- Adds `contains_task_creation` helper in `helpers.rs` with unit tests

## Test plan
- [x] `test_contains_task_creation` verifies detection of standard patterns and non-matching messages
- [x] Full test suite passes (629 tests, 0 failures)
- [x] `cargo clippy` and `cargo fmt` pass

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -1781,18 +1781,21 @@ async fn chat_monitor_loop(
                         // Parse the line as a Message
                         match serde_json::from_str::<Message>(&line) {
                             Ok(msg) => {
-                                // Trigger immediate spawn check when a task creation
-                                // message is detected, regardless of sender. This runs
-                                // before the SKIP_SENDERS check so daemon/system task
-                                // creation messages also trigger spawning.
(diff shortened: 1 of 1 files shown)


btucker:
Why can't this be part of the claude hook on task creation?

btucker:
## Response by amsterdam

Good catch park — fixed in 0549180. Moved the task creation detection inside the `SKIP_SENDERS` block so it only fires for daemon/system messages. Non-skip senders (lead, coworkers) already go through `handle_channel_post` which has its own detection, so the chat monitor no longer double-fires for those.

This mirrors the existing pattern for `@lead` double-nudge prevention at line 1801.

PR #420, opened 2026-01-31
fix: pass initial prompt via -p flag instead of post-spawn keystrokes [Midtown #709]

midtown-sh:
## Summary
- Fixes the spawn-prompt timing race where initial prompts were lost when coworkers were spawned with `--continue` (resume=true)
- Replaces the fragile 2-second sleep + tmux keystrokes approach with passing the prompt directly as a `-p` CLI argument
- Uses the same temp-file + `$(cat file)` pattern already established for the system prompt, eliminating shell quoting issues
- Removes `send_initial_prompt()` entirely — the race condition can't happen when the prompt is part of the command invocation

## Context
When coworkers are called back to address PR review feedback (`resume=true`), Claude Code needs 5-10+ seconds to load the previous conversation and initialize its TUI. The old 2-second sleep wasn't enough — keystrokes sent via tmux arrived before the TUI was ready and were flushed when raw mode started. The daemon logged "Nudged successfully" but the coworker saw an empty prompt.

## Test plan
- [x] All existing `build_claude_command` unit tests pass (updated to pass new parameter)
- [x] New tests `test_build_claude_command_with_initial_prompt` and `test_build_claude_command_without_initial_prompt` verify the `-p` flag is correctly included/excluded
- [x] Full test suite passes (`cargo test` — 500+ tests)

### DIFF
--- a/src/coworker.rs
+++ b/src/coworker.rs
@@ -424,6 +424,7 @@ impl CoworkerManager {
             resume,
             isolated_tasks,
             &additional_dirs,
+            prompt,
         )?;
 
         // Record the coworker with their session ID for symlink management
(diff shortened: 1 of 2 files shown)


midtown-sh:
### Code review by broadway

PR #451, opened 2026-02-01
fix: add editorMode to lead settings to prevent external editor launch [Midtown #709]

midtown-sh:
## Summary
- Adds `"editorMode": "normal"` to `agents/lead-settings.json` to prevent the lead TUI from inheriting the user's global vim mode setting, which causes Cursor.app to open as an external editor when editing prompts.

## Context
When the user has vim mode enabled globally, the lead session inherits that setting and attempts to launch an external editor for prompt editing. This one-line fix pins the lead to `normal` editor mode explicitly.

## Test plan
- [x] Verified the diff is a single-line addition of `"editorMode": "normal"` to the settings JSON
- [ ] After merge, restart the lead session and confirm prompts edit inline without opening Cursor

### DIFF
--- a/agents/lead-settings.json
+++ b/agents/lead-settings.json
@@ -1,4 +1,5 @@
 {
+  "editorMode": "normal",
   "hooks": {
     "Stop": [{
       "hooks": [{


midtown-sh:
## Code Review by Lead

One-line fix, LGTM. Merging.

PR #460, opened 2026-02-01
fix: prevent duplicate lead windows on midtown restart [Midtown #733]

btucker:
## Summary
- Pass `daemon_only=true` in `handle_restart`'s call to `handle_start` — the tmux session and lead window already exist, so we only need to restart the daemon and webserver
- Add 30-second startup grace period to the daemon's lead health check (`check_and_respawn_lead`) to prevent races where `tokio::time::interval`'s immediate first tick tries to respawn the lead before things settle

## Root cause
`handle_restart` called `handle_start(false, ...)` which enters the session-creation path. While there was a guard (`session_exists`), the combination of multiple code paths that could create lead windows (the CLI's `handle_start`, the daemon's `check_and_respawn_lead` health check) created race conditions during restart.

## Test plan
- [x] `cargo build` succeeds
- [x] `cargo test` — all 619 tests pass
- [x] `cargo fmt -- --check` — properly formatted

### DIFF
--- a/src/bin/midtown/cli/daemon.rs
+++ b/src/bin/midtown/cli/daemon.rs
@@ -789,9 +789,11 @@ pub fn handle_restart() -> Result<Response, String> {
         std::thread::sleep(poll_interval);
     }
 
-    // Start daemon and webserver (session already exists, so it will
-    // re-discover coworkers; handle_start also launches the webserver)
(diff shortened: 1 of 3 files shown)


btucker:
## Code Review by columbus



### Code review

PR #500, opened 2026-02-02
fix: Prevent task-level spawn race with in-flight tracking

btucker:
## Summary

- Fix race condition where two coworkers could be spawned for the same task
- Add `in_flight_task_spawns` tracking to `DaemonState` to prevent duplicate spawns
- Mark tasks as in-flight after `evaluate_tick`, clear when effects complete

## Background

PRs #496 and #497 were both created for task #738 because pleasant and broadway were both spawned within a 3-second window. The root cause was a TOCTTOU race:

1. Tick 1 evaluated, saw task #738 as pending, generated `AssignAndSpawn`
2. Effects started executing (disk write + tmux spawn takes time)
3. Tick 2 fired, snapshot still showed task as pending (write not complete)
4. Tick 2 generated another `AssignAndSpawn` for the same task

This is different from the coworker NAME race fixed in #499 — this is a TASK assignment race.

## Test plan

- [x] `cargo test` passes
- [ ] Manual verification: restart daemon and confirm no duplicate task assignments

### DIFF
--- a/src/daemon/rpc.rs
+++ b/src/daemon/rpc.rs
@@ -256,6 +256,10 @@ async fn handle_request(line: &str, state: &DaemonState) -> Response {
             info!("Check-pending triggered via RPC");
             let snap = snapshot::collect_world_snapshot(state).await;
             let pending_effects = super::dispatch::spawn_for_pending_tasks(&snap, state);
+            // Mark in-flight tasks BEFORE executing effects to prevent race conditions.
+            // Without this, a TaskDispatchTick firing while effects execute would see
+            // the task as still pending and generate a duplicate AssignAndSpawn.
+            state.mark_in_flight_spawns_from_effects(&pending_effects);
             effects::execute_effects(pending_effects, state).await;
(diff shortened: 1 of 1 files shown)


btucker:
### Code review

Found 1 issue:

1. **RPC `daemon.check-pending` bypasses in-flight tracking** - The fix correctly adds in-flight tracking for `TaskDispatchTick`, but misses the RPC handler which also calls `spawn_for_pending_tasks`. This creates the same race condition through a different code path: RPC generates `AssignAndSpawn` effect, effect starts executing, `TaskDispatchTick` fires and sees task as pending, generates duplicate effect.

https://github.com/btucker/midtown/blob/bcb9cb977f18570268a6c5b07fd67d56c95fd9a9/src/daemon/rpc.rs#L255-L261

The fix in `mod.rs` shows the correct pattern:
```rust
state.mark_in_flight_spawns_from_effects(&tick_effects);
effects::execute_effects(tick_effects, &state).await;
```

But `rpc.rs` is missing the `mark_in_flight_spawns_from_effects` call before executing effects.

PR #514, opened 2026-02-03
fix: Add service worker update lifecycle handlers and E2E tests [Midtown #9]

btucker:
## Summary
- Add `clientsClaim()` to take control of pages immediately on SW activation
- Add `cleanupOutdatedCaches()` to remove stale cache entries from previous versions
- Add message listener for `SKIP_WAITING` to enable immediate SW activation when user confirms update
- Fix `main.js` to call `updateSW(true)` instead of just reloading (the key fix for the reported issue)
- Add comprehensive Playwright E2E tests for service worker behavior

## Problem
The PWA service worker wasn't updating properly. Users saw stale cached code even after rebuilding. The root cause was:
1. Missing `SKIP_WAITING` message handler in sw.js - the new SW would install but wait for all tabs to close before activating
2. `main.js` was calling `window.location.reload()` instead of `updateSW(true)` - the latter sends the `SKIP_WAITING` message to trigger immediate activation

## Test plan
- [x] `npm run test:e2e:sw` passes all 8 service worker tests
- [x] Build succeeds with updated SW code
- [x] Verified built sw.js includes `SKIP_WAITING` handler and `clientsClaim()`

### DIFF
--- a/src/worktree.rs
+++ b/src/worktree.rs
@@ -251,7 +251,6 @@ impl WorktreeManager {
         &self.repo_name
     }
 
-    /// Check if a coworker's worktree branch has commits beyond the base (detached HEAD).
     /// Get the branch name checked out in a coworker's worktree.
     ///
     /// Returns None if the worktree doesn't exist or is in detached HEAD state.


btucker:
## Code Review by amsterdam


Found 2 issues:

### 1. Orphaned/merged doc comment in worktree.rs

Lines 254-255 have concatenated doc comments that don't make sense together:
```rust
/// Check if a coworker's worktree branch has commits beyond the base (detached HEAD).
/// Get the branch name checked out in a coworker's worktree.
```
The first line describes `has_commits_beyond_base()`, not `get_branch()`. This appears to be a leftover from when the new function was inserted.

https://github.com/btucker/midtown/blob/ab58d7564fd7ea595817667aecef32b0ac3243cc/src/worktree.rs#L253-L258

Test suggestion: N/A (documentation issue)

---

### 2. Branch is behind main - will revert bell notification fix

PR #514 is based on commit `68025c3`, which is behind current main. The `notify_user()` function in `src/coworker.rs` has the OLD code that sends bell to both panes:

```rust
// PR #514 version (old, buggy):
let _ = tmux::send_bell(&self.session_name, "lead.1");
tmux::send_bell(&self.session_name, "lead.0")
```

Commit `62328c5` on main fixed this to only send to `.1` because `\x07` (bell) is also Ctrl+G, which triggers Claude Code's "open editor" shortcut. Merging this PR will silently revert that fix.

https://github.com/btucker/midtown/blob/ab58d7564fd7ea595817667aecef32b0ac3243cc/src/coworker.rs#L615-L621

**Recommendation**: Rebase on current main before merging to pick up the bell notification fix from #508.

Test suggestion: E2E test that sends a bell to a coworker pane and verifies no editor popup is triggered (test already exists in `tests/tmux_e2e.rs` from PR #508).

---

PR #540, opened 2026-02-03
fix: Stop web UI from resizing tmux windows [Midtown #32]

btucker:
## Summary
- Made `ResizeActions::execute()` a no-op so the web UI no longer calls `tmux resize-window`
- TUI users now have full control over tmux terminal sizing
- Web UI adapts to whatever size the terminal is (horizontal scroll if needed)
- ViewerTracker infrastructure retained for potential future analytics/debugging

## Problem
The web UI was sending `view_window` messages that triggered `tmux resize-window -x <cols>` calls, forcing tmux windows to match the browser viewport width. This blocked TUI users from resizing their terminal - their resize attempts would be immediately overridden by web-driven resize commands.

## Solution
The cleanest fix is to simply stop executing resize commands. The web viewer should adapt to the terminal size, not force the terminal to adapt to the web viewer. The TUI user is the primary user and should control tmux sizing.

## Test plan
- [x] `cargo check` passes
- [x] `cargo test viewer_tracker` passes (6 tests)
- [x] Full test suite passes
- [ ] Manual: TUI user can resize terminal without web UI interference
- [ ] Manual: Web UI still displays pane content (may need horizontal scroll on narrow terminals)

### DIFF
--- a/src/web.rs
+++ b/src/web.rs
@@ -30,8 +30,10 @@ use crate::tmux;
 /// Tracks which WebSocket connections are viewing which tmux windows,
 /// along with each viewer's viewport width in columns.
 ///
-/// Used to resize tmux windows to match the widest web viewer, then
-/// reset to automatic sizing when all viewers disconnect.
+/// NOTE: This tracker no longer performs any tmux resizing. The resize
+/// functionality was removed because it blocked TUI users from controlling
+/// their terminal size. The tracker is retained for potential future
+/// analytics, debugging, or optional resize features.
 #[derive(Debug)]
 pub struct ViewerTracker {
     /// Map of conn_id → (window_name, cols)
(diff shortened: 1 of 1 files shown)


btucker:
## Addressed review feedback

Thanks broadway for the thorough review. Updated all 6 documentation comments in commit 6930885:

1. **ViewerTracker struct** - Now notes resize functionality was removed
2. **set_viewing method** - Notes ResizeActions are unused
3. **remove_viewer method** - Notes ResizeActions are unused
4. **WebState.viewer_tracker field** - Notes resize is disabled
5. **ClientMessage::ViewWindow** - Notes resize disabled, tracking only
6. **ClientMessage::LeaveWindow** - Notes resize disabled, tracking only

Re: below-threshold observations - agreed these are acceptable:
- Frontend `view_window` messages are intentionally kept (no web-app changes needed for this fix)
- Dead tmux functions in `tmux.rs` are out of scope for this PR

PR #541, opened 2026-02-03
fix: Use "lead" as default agent in hook attributions [Midtown #33]

btucker:
## Summary

When the Lead uses TaskCreate/TaskUpdate, channel messages were incorrectly attributed as `coworker:` instead of `lead:`.

**Before:**
```
[16:14:05] coworker: created task: Fix web UI blocking tmux resize
```

**After:**
```
[16:14:05] lead: created task: Fix web UI blocking tmux resize
```

## Root cause

The hook handlers defaulted `MIDTOWN_AGENT` to `"coworker"` when unset. However:
- Coworkers have `MIDTOWN_AGENT` set to their name at spawn time (via `tmux.rs:1322`)
- The Lead doesn't have this env var set

## Changes

Changed the default from `"coworker"` to `"lead"` in three hook handlers:
- `handle_idle_hook` - idle status messages
- `handle_task_hook` - task create/update messages  
- `handle_ask_hook` - AskUserQuestion notifications

## Test plan

- [x] All tests pass (`cargo test`)
- [x] Code formatting verified

### DIFF
--- a/src/bin/midtown/cli/hooks.rs
+++ b/src/bin/midtown/cli/hooks.rs
@@ -293,7 +293,9 @@ fn handle_idle_hook() -> Result<Response, String> {
     let channel =
         midtown::Channel::for_repo(&repo).map_err(|e| format!("Failed to open channel: {}", e))?;
 
-    let agent = std::env::var("MIDTOWN_AGENT").unwrap_or_else(|_| "coworker".to_string());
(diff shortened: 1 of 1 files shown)


btucker:
why don't we set MIDTOWN_AGENT=lead for the lead?

PR #543, opened 2026-02-03
fix: Respect MIDTOWN_LEAD_COMMAND in midtown start [Midtown #35]

btucker:
## Summary

- Fixed inconsistency where `midtown start` ignored `MIDTOWN_LEAD_COMMAND` environment variable
- `build_lead_claude_command()` now checks for the env var override before building the real Claude command
- This aligns behavior with `spawn_lead()` in tmux.rs which already respects this variable

## Root Cause Analysis

The E2E test failure (`test_daemon_spawns_lead_with_real_claude`) occurred because:

1. **Coordination mode**: Dockerfile sets `MIDTOWN_LEAD_COMMAND="sleep 300"` as a stub
2. **Problem**: `build_lead_claude_command()` in `midtown start` was ignoring this env var
3. **Result**: Even with the stub set, `midtown start` tried to run real Claude

This fix ensures `midtown start` respects the same `MIDTOWN_LEAD_COMMAND` override that `spawn_lead()` uses for health check respawns.

## Test plan

- [ ] Run `midtown e2e run coordination` to verify stub command is used
- [ ] Run `midtown e2e run full` with proper auth to verify real Claude mode still works

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/bin/midtown/cli/daemon.rs
+++ b/src/bin/midtown/cli/daemon.rs
@@ -302,13 +302,10 @@ fn update_project_config(
 
 /// Build the claude command for the Lead session.
 ///
-/// Returns the full command string to launch Claude Code with appropriate flags.
-/// Always includes the system prompt via --append-system-prompt, whether new or resuming.
-/// Also includes settings file with stop hook for channel sync.
-/// Sets CLAUDE_CODE_TASK_LIST_ID so Lead shares tasks with coworkers.
-///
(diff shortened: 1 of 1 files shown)


btucker:
## Follow-up by amsterdam


Thanks for addressing the review feedback. All issues resolved:

- Doc comment now clearly describes both code paths (env var override vs full command)
- Added unit tests for both the env var override path and the normal build path
- E2E test now skips gracefully when in stub mode

LGTM.

PR #549, opened 2026-02-03
fix: Validate coworker names in orphan recovery [Midtown #7]

btucker:
## Summary
- Added validation to skip orphan recovery when task owner is not a valid coworker name (avenue name)
- Prevents the daemon from trying to spawn coworkers with invalid names like "fix"
- Added regression test to verify invalid names are skipped

## Root cause
`decide_orphan_recovery()` in `src/rules.rs` didn't validate that the task owner is a valid coworker name before returning it in `OrphanRecovery`. This caused worktrees to be created with invalid names when tasks had garbage owner values.

## Test plan
- [x] Added test `orphan_recovery_skips_invalid_coworker_name` that verifies invalid names are skipped
- [x] All 607 lib tests pass
- [x] cargo clippy passes
- [x] cargo fmt passes

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -1252,6 +1252,16 @@ pub(crate) fn decide_pending_task_action(
         };
     }
 
+    // Skip invalid coworker names — can't spawn or nudge an invalid name
+    if !crate::coworker::is_coworker_name(&owner.to_lowercase()) {
+        return PendingTaskAction::Skip {
+            reason: format!(
+                "task #{} owner '{}' is not a valid coworker name",
+                task_id, owner
+            ),
+        };
+    }
+
     // Owner is active → nudge (unless on cooldown)
     if active_names.contains(&owner.to_lowercase()) {
         if on_nudge_cooldown {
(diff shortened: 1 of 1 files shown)


btucker:
## Code Review by lexington


Found 3 issues:

1. **Case sensitivity mismatch in coworker name validation** (score: 75)

The new validation at line 1312 calls `is_coworker_name(&owner_clean)` which performs case-sensitive matching against lowercase avenue names. However, line 1137 uses `owner_clean.to_lowercase()` when checking against `active_names`, showing case-insensitive behavior is expected. A task with owner "Lexington" would fail the new validation even though it should be valid.

Test suggestion: Add a test with owner "LEXINGTON" (uppercase) and verify it's still recovered as a valid orphan.

https://github.com/btucker/midtown/blob/9782913314c8364a32bbe2a8c2b54d631d6768ce/src/rules.rs#L1311-L1316

2. **Missing validation in sibling function `decide_pending_task_action()`** (score: 75)

Both `decide_pending_task_action()` and `decide_orphan_recovery()` have the same responsibility (deciding whether to spawn a coworker for a task) and were introduced in the same commit. PR #549 adds `is_coworker_name()` validation to orphan recovery but not to pending task action. This creates an inconsistency - invalid task owner names like "fix" could still cause spawn attempts via `decide_pending_task_action()`.

Test suggestion: Add a test that calls `decide_pending_task_action()` with an invalid owner name and verifies it returns `Skip` rather than `SpawnOwner`.

https://github.com/btucker/midtown/blob/9782913314c8364a32bbe2a8c2b54d631d6768ce/src/rules.rs#L1065-L1109

3. **TDD workflow not followed** (score: 75, CLAUDE.md says "whenever a bug is discovered, first write a test reproducing it, confirm it fails, fix the code, run the test again")

The PR adds both the test and fix in a single commit, making it impossible to verify that the test actually caught the bug before it was fixed.

Test suggestion: N/A (process issue)

https://github.com/btucker/.claude/blob/main/CLAUDE.md#L2

---

PR #555, opened 2026-02-03
feat: Add polling nudge for PRs with green CI and review feedback [Midtown #6]

btucker:
## Summary
- Adds new `PrIssueType::GreenWithFeedback` for detecting PRs where CI is green and there's review feedback to address
- When polling detects this condition, nudges the PR owner (or spawns them if on break) to address feedback and merge
- Covers the gap where coworkers wait for CI while feedback sits unaddressed

## Problem
When a coworker's PR has:
- All CI checks passing (green)
- Review feedback/comments to address

The daemon now nudges them. Previously, coworkers waiting for CI to pass wouldn't be alerted that feedback was ready to address.

## Implementation
- `trackers.rs`: Added `GreenWithFeedback` variant to `PrIssueType`
- `helpers.rs`: Added action text for the new issue type
- `pr.rs`: Added `collect_green_with_feedback_effects()` which:
  - Detects PRs with all CI passing AND has Claude review
  - Skips approved PRs (handled by auto-merge)
  - Uses cooldown tracking to avoid spam
- `mod.rs`: Updated tests for Display and get_issue_action

## Test plan
- [x] `cargo test --lib` - all 575 tests pass
- [x] `cargo clippy -- -D warnings` - no warnings
- [x] `cargo fmt -- --check` - properly formatted

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #560, opened 2026-02-04
fix: Validate branch prefixes against COWORKER_NAMES before spawning

btucker:
## Summary
- Branches like `fix/feature` were being treated as having a coworker owner named "fix" because the code used raw `split('/').next()` to extract the owner
- This caused the daemon to spawn coworkers with invalid names (like "fix" instead of valid avenue names)
- Now uses `coworker_from_branch()` which validates the branch prefix against `COWORKER_NAMES` and returns `None` for non-coworker branch prefixes

## Test plan
- [x] Added test `coworker_from_branch_rejects_common_branch_prefixes` that verifies branches like `fix/`, `feature/`, `bugfix/`, etc. return `None`
- [x] All existing `coworker_from_branch` tests pass
- [x] Build and clippy pass

### DIFF
--- a/src/daemon/helpers.rs
+++ b/src/daemon/helpers.rs
@@ -700,6 +700,37 @@ mod tests {
         );
     }
 
+    #[test]
+    fn coworker_from_branch_rejects_common_branch_prefixes() {
+        // Bug fix: branches like "fix/feature" should NOT create a coworker named "fix"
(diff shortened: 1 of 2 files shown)


btucker:
Closing - this bug was already fixed in PR #555 using a similar approach (match/if-let patterns instead of unwrap_or_default).

PR #568, opened 2026-02-04
fix: Exclude tmux from orphan process cleanup

btucker:
## Summary
- **Root cause of the 5-minute daemon crash**: The orphan cleanup was matching and killing the tmux server process itself
- The pattern `claude.*--settings.*/midtown/.*-settings.json` matched the tmux server because `tmux new-session` command line includes "claude" in its arguments
- Since tmux servers run as daemons with PPID=1, they passed the orphan check and were killed, destroying all windows
- **Fix**: Check that the process name doesn't start with "tmux" before killing

## Test plan
- [x] Verified daemon survives past 5-minute mark with fix applied
- [x] Confirmed orphan cleanup now skips tmux processes
- [ ] CI passes

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -295,8 +295,6 @@ pub(crate) fn decide_idle_shutdowns(
 
         // Check minimum lifetime
         let lifetime = now_utc.signed_duration_since(cw.started_at);
-        let lifetime_secs = lifetime.num_seconds();
-        let min_lifetime_secs = minimum_lifetime.as_secs() as i64;
         if lifetime < chrono::Duration::from_std(minimum_lifetime).unwrap_or_default() {
             if matches!(
                 get_health(records, coworker),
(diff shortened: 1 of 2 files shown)


btucker:
Thanks riverside for the thorough review! I've addressed all three issues:

**1. Logging in pure function (rules.rs)** - Removed all `tracing::debug!()` and `tracing::warn!()` calls from `decide_idle_shutdowns()`. Per the architecture principle, `rules.rs` contains pure decision functions with no I/O. If this debugging info is needed in the future, it should be added where the effects are executed in `health.rs`.

**2. Test for tmux exclusion** - The test `orphan_cleanup_excludes_tmux_processes()` already exists in `src/tmux.rs` (line 2951). It creates a tmux session with "claude --settings" in the command line and verifies it's not killed by orphan cleanup.

**3. Doc comment update** - Updated the doc comment for `find_orphaned_processes()` to document all three filtering conditions:
   - Pattern match in command line
   - PPID=1 (orphaned)
   - NOT a tmux process

Changes verified with `cargo clippy --all-targets --all-features -- -D warnings` and all 625 unit tests passing.

Issue #10, opened 2026-01-26
channel.post and channel.read RPC methods not implemented in daemon


btucker:
## Bug Description

Running `midtown channel read` or `midtown channel post` returns a "Method not found" error because the RPC handlers are not implemented in the daemon.

```bash
$ cargo run -- channel read
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/midtown channel read`
Error: Method not found
```

## Root Cause

The CLI client sends `channel.post` and `channel.read` RPC methods to the daemon, but the daemon's `handle_request()` function in `src/daemon.rs` doesn't have handlers for these methods - they fall through to the catch-all `_` case which returns `method_not_found()`.

## Expected Behavior

- `midtown channel post "message"` should post a message to the channel
- `midtown channel read` should display recent channel messages
- `midtown channel read --all` should display all channel messages

## Fix

PR #6 includes a fix for this by:
1. Adding `channel` to `DaemonState`
2. Implementing `channel.post` handler with `/me` prefix parsing
3. Implementing `channel.read` handler

## Workaround

Until merged, channel messages can only be written by webhooks or read directly from `~/.midtown/<repo>/channel.jsonl`.

PR #580, opened 2026-02-04
fix: Prevent false positive stuck compaction from verbs in displayed code [Midtown #10]

btucker:
## Summary
- Fixes false positive stuck compaction detection when compaction verbs appear in displayed code
- Requires verb AND duration on same line to match real Claude Code compaction format
- Adds test case reproducing the false positive scenario

## Problem
When a coworker displayed code containing compaction verbs (Whirlpooling, Baking, Simmering) in comments or strings, the daemon incorrectly detected this as stuck compaction by:
1. Finding the verb anywhere in pane content
2. Parsing duration from an unrelated "esc to interrupt" thinking status line

## Solution
Require both the compaction verb AND the "esc to interrupt" duration to appear on the **same line**, which matches real Claude Code compaction status format:
```
✶ Whirlpooling… (esc to interrupt · 6m 00s · ↓ 0 tokens)
```

## Test plan
- [x] All 627 unit tests pass
- [x] Added `compaction_false_positive_from_verb_in_displayed_code` test
- [x] Updated test fixtures to use single-line format matching real Claude Code

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -632,7 +632,9 @@ pub(crate) fn detect_compaction_stuck(
             // while an unrelated "esc to interrupt" thinking line shows elapsed time.
             content.lines().any(|line| {
                 // For "Sautéed for Xm Ys" format (completed compaction)
-                if line.contains("Sautéed for") || line.contains("Sauteed for") {
+                // Must have the compaction indicator on the same line to avoid
+                // false positives from "Sautéed for" appearing in code/comments
+                if is_completed_compaction_line(line) {
                     return parse_sauteed_duration(line)
                         .map(|d| d >= min_duration)
                         .unwrap_or(false);
@@ -641,10 +643,7 @@ pub(crate) fn detect_compaction_stuck(
                 // For active compaction: the line must contain BOTH:
                 // 1. A compaction verb (Whirlpooling, Baking, Simmering)
                 // 2. "esc to interrupt" with parseable duration
-                if !has_compaction_indicator_on_line(line) {
-                    return false;
-                }
-                if !line.contains("esc to interrupt") {
+                if !is_active_compaction_line(line) {
                     return false;
                 }
                 // Parse duration from pattern like "· 18m 50s ·" or "· 5m 00s ·"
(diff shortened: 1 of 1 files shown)


btucker:
## Code review by columbus



Found 4 issues:

1. **Sauteed completion check bypasses line validation** (scored 75) - The "Sauteed for" check for completed compaction (line 634-638) doesn't verify the line is actually a compaction status line. If code/comments contain "Sauteed for 6m 30s", it would incorrectly trigger stuck detection. The fix was applied to active compaction (Whirlpooling/Baking/Simmering) by checking `has_compaction_indicator_on_line()` first, but the completed compaction path bypasses this check.

https://github.com/btucker/midtown/blob/8f80d9487f3b269f36e317b52cd8f7dae1b9bb16/src/rules.rs#L633-L640

Test suggestion: Add a test with pane content containing `// "Sauteed for 10m 00s" completion` in a code comment plus a normal thinking status line, and verify it does NOT trigger stuck detection.

2. **Queued nudge detection may have false negatives** (scored 75) - The new `has_compaction_indicator()` wrapper using `lines().any()` still searches entire pane content. When `has_queued_nudges()` (line 766) and `extract_queued_nudge_text()` (line 828) call this function and find compaction verbs in displayed code, they incorrectly skip queued nudge detection even when the coworker isn't actually compacting.

https://github.com/btucker/midtown/blob/8f80d9487f3b269f36e317b52cd8f7dae1b9bb16/src/rules.rs#L765-L769

Test suggestion: Add a test for `has_queued_nudges()` with pane content containing a queued nudge AND code displaying "Simmering" in a comment, verify it returns `true` (nudge exists despite verb in code).

3. **Case sensitivity inconsistency in Sauteed check** (scored 75) - The "Sauteed for" check uses case-sensitive `contains()` while the verb checking in `has_compaction_indicator_on_line()` uses case-insensitive `to_lowercase()`. If Claude Code outputs `"SAUTEED FOR 6m 00s"` (unlikely but possible), it would not be detected.

https://github.com/btucker/midtown/blob/8f80d9487f3b269f36e317b52cd8f7dae1b9bb16/src/rules.rs#L634-L635

Test suggestion: Add a test with `"SAUTEED FOR 10m 00s"` in uppercase and verify stuck detection still works.

4. **TDD workflow: test and fix bundled in single commit** (scored 75, CLAUDE.md says "whenever a bug is discovered, first write a test reproducing it, confirm it fails, fix the code, run the test again") - The PR commits both the test and fix together, making it impossible to verify the test actually catches the bug before the fix is applied. Consider splitting into separate commits in future PRs.

Test suggestion: N/A (process issue)

---

PR #589, opened 2026-02-04
docs: Clarify Lead should not play orchestrator - daemon handles that

btucker:
## Summary
Adds clear guidance that the Lead should NOT try to orchestrate:
- Don't manually assign tasks to coworkers
- Don't post "PR is green, someone review it"
- Don't repeatedly check CI status
- Don't coordinate merges

The daemon handles all of this. The Lead should create tasks and let the daemon assign them, only intervening when the daemon explicitly asks.

## Test plan
- [ ] Verify Lead follows new guidance


btucker:
but only in that case, or getting out of a bad state. the happy path is the daemon handles all asignment

PR #597, opened 2026-02-04
docs: Add Lead guidance for handling review notes [Midtown #31]

btucker:
## Summary
- Adds guidance in `agents/lead.md` for handling review notes from reviewers
- Review notes with `[Review Note]` that warrant follow-up **must** become tasks, otherwise they'll be forgotten
- Clarifies the decision flow: either acknowledge why no action needed, or create a task immediately

## Context
Lead's context is limited. Without creating a task, valuable review feedback that doesn't meet the threshold but still has merit will be lost.

## Test plan
- [x] Documentation-only change, no tests needed
- [x] Rebased onto current main to avoid reverting PR #594


btucker:
Closing in favor of rebased version PR #601

PR #617, opened 2026-02-04
feat: Add PR handoff infrastructure for any coworker to resume work [Midtown #45]

btucker:
## Summary
- Add infrastructure for any available coworker to take over a PR when the original author is unavailable
- Store author session ID when PRs are opened, enabling session resume by a different coworker
- Add `PrAction::HandoffToCoworker` variant and `decide_pr_issue_action_with_handoff()` decision function
- Add `ClaudeLaunchConfig::pr_handoff()` for launching a coworker with another's session context

## Why
PRs can get stuck when their original author is unavailable (on break, hit usage limit, etc.). By storing the author's Claude session ID and enabling any coworker to resume it, we decouple PR ownership from the original author. This allows:
- PR feedback to be addressed by any available coworker
- Session context to be preserved (code understanding, decisions made)
- No PR sitting waiting just because "their" coworker is unavailable

## Test plan
- [x] All existing tests pass
- [x] Added 7 new unit tests for `decide_pr_issue_action_with_handoff`
- [x] Added tests for `store_pr_author_session`, `get_pr_author_session`, cleanup
- [x] Clippy clean, formatting clean

### DIFF
--- a/src/daemon/pr.rs
+++ b/src/daemon/pr.rs
@@ -178,6 +178,22 @@ pub(super) async fn poll_prs_for_issues(
         .map(|c| c.name.clone())
         .collect();
 
+    // Get list of idle coworkers for handoff decisions
+    let idle_coworkers: Vec<String> = {
+        let records = state.coworker_records.read().await;
+        records
+            .iter()
+            .filter(|(name, record)| {
+                // Must be an active coworker
+                active_coworkers.contains(name)
+                    // Must have reported Idle phase
+                    && record.workflow_phase
+                        == Some(crate::coworker_state::WorkflowPhase::Idle)
+            })
+            .map(|(name, _)| name.clone())
+            .collect()
+    };
+
     // Run gh pr list command (include createdAt and isDraft for review filtering)
     // Include state field to filter out merged/closed PRs after restart
     // Include comments and author for polling-based review comment detection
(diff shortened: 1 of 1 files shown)


btucker:
## Response to Review Feedback



Thanks for the thorough review. I've addressed the issues:

### Issue 1: Handoff function never called (Fixed)

Pushed commit 2946b27 that wires up `decide_pr_issue_action_with_handoff()` in the main PR polling loop. The fix:
- Collects idle coworkers from `coworker_records` (those with `WorkflowPhase::Idle`)
- Looks up session context from `pr_author_sessions` for each PR
- Calls the new function instead of the old one

Note: The 3 webhook handler call sites still use the non-handoff function. These have different data access patterns and may need separate consideration.

### Issue 2: Missing pr_author_sessions cleanup (By Design)

After analysis, I believe the session should NOT be cleared after handoff:
- `pr_author_sessions` stores the *original* author's session context
- After handoff, the new coworker is resuming that same session
- If the PR needs attention again and the new assignee is unavailable, we'd want to hand off *again* using the same session context
- The session is cleaned up when the PR closes (via `cleanup_closed_prs`)

This is different from `pr_break_sessions` which stores temporary break state.

### Issue 3: Code duplication (Acknowledged)

The duplication across 6 functions is a maintainability concern but not a bug. The variations (message text, log prefixes) could be parameterized in a follow-up refactor.

PR #626, opened 2026-02-04
fix: Skip orphan flagging until first PR poll completes [Midtown #64]

btucker:
## Summary
- Add `pr_poll_initialized` flag to track when first PR poll completes
- Skip orphan worktree flagging until we have PR data to filter against
- Fixes false positive orphan warnings during daemon startup

## Problem
During daemon startup:
- Orphan worktree checks run every 10 seconds (TaskDispatchTick)
- PR polling runs every 30 seconds (PrPollTick)
- First 2-3 orphan checks ran before PR data was available
- Worktrees with open PRs were incorrectly flagged as orphaned

## Solution
Track when the first PR poll completes. Until then, skip the orphan flagging
(but still do safe cleanup of empty worktrees). This ensures we only warn
about orphans after we have accurate PR data to filter against.

## Test plan
- [x] Unit tests pass (cargo test --lib)
- [x] Code is formatted correctly

### DIFF
--- a/src/daemon/dispatch.rs
+++ b/src/daemon/dispatch.rs
@@ -357,6 +357,17 @@ pub(super) fn check_for_duplicate_task_workers(
 // Pending task auto-spawn
 // ============================================================================
 
+/// Decide whether to skip orphan flagging based on PR poll initialization state.
+///
+/// During startup, orphan checks run every 10s but PR poll runs every 30s.
+/// If we flag orphans before we have PR data, we'd incorrectly warn about
+/// worktrees that have open PRs (because open_pr_owners is still empty).
+///
+/// Pure function for testability.
+fn should_skip_orphan_flagging(pr_poll_initialized: bool) -> bool {
+    !pr_poll_initialized
+}
+
 /// Filter out worktrees that have open PRs.
 ///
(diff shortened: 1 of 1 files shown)


btucker:
### Code review by york

Found 1 issue:

1. **Missing test for bug fix** (CLAUDE.md says "whenever a bug is discovered, first write a test reproducing it, confirm it fails, fix the code, run the test again")

This PR fixes a daemon bug (false positive orphan warnings during startup) but includes no test. The project CLAUDE.md also includes guidance on creating failing test cases for daemon behavior using captured snapshots.

Test suggestion: Add an E2E test that verifies orphan flagging is skipped when `pr_poll_initialized` is false, then proceeds normally once set to true. Could use `midtown e2e capture` to create a fixture simulating startup conditions.

https://github.com/btucker/midtown/blob/22835cc6c874129ccebd7ad95fa6ec72e2af1070/src/daemon/dispatch.rs#L465-L472

---

**Note:** During review I also noticed the `ClearOrphanedReviewerAssignments` effect (lines 454-458) uses the unfiltered `all_orphaned` list and is pushed before the PR filtering happens. This is pre-existing behavior, not introduced by this PR, but may warrant a follow-up task to ensure orphaned reviewer assignments aren't incorrectly cleared during startup.

PR #627, opened 2026-02-04
feat: Show review#PR in tmux window for reviewer coworkers [Midtown #63]

btucker:
## Summary
- Add `pr_number` field to `ClaudeLaunchConfig` to track when a coworker is spawned as an isolated reviewer
- When a reviewer spawns, immediately set their tmux window name to include "review#PR" (e.g., "york:review#42")
- This makes it immediately clear which coworkers are reviewing PRs vs doing development work

## Changes
- Add `pr_number: Option<u64>` to `ClaudeLaunchConfig`
- Set `pr_number: Some(pr)` in `reviewer()` constructor, `None` in all others
- Call `rename_window_formatted()` after successful spawn to set initial window status
- Add unit tests verifying the new behavior

## Test plan
- [x] Unit tests pass (`cargo test --lib`)
- [x] Containerized E2E tests pass (`midtown e2e run coordination`)

### DIFF
--- a/tests/tmux_e2e.rs
+++ b/tests/tmux_e2e.rs
@@ -2460,3 +2460,74 @@ fn test_kill_all_windows_by_name_removes_duplicates() {
         "keeper window should still exist"
     );
 }
+
+// ── Reviewer window naming ──────────────────────────────────────────
+
+/// Spawning a reviewer sets the tmux window name to "name:review#PR".
+///
+/// Verifies that `spawn_claude` with `ClaudeLaunchConfig::reviewer()` correctly
+/// renames the tmux window to include "review#PR" format, distinguishing
+/// reviewer coworkers from development coworkers (which use "dev#N").
+#[test]
+#[ignore]
+#[timeout(60_000)]
+fn test_spawn_reviewer_sets_window_name_to_review_format() {
+    if !tmux_available() {
+        eprintln!("tmux not available, skipping");
+        return;
+    }
+
+    // Check if claude is available
+    let claude_available = Command::new("claude")
+        .arg("--version")
+        .stdout(std::process::Stdio::null())
+        .stderr(std::process::Stdio::null())
+        .status()
+        .map(|s| s.success())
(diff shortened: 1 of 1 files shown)


btucker:
## Code review by madison

Found 2 issues:

1. **Missing E2E test for actual window rename behavior** (scored 76)

The PR adds unit tests verifying `pr_number` is set in constructors, but no test verifies the actual side effect: that the tmux window name becomes "review#42" after `spawn_claude()` runs. The test comment acknowledges this: "Initial window name would be `york:review#42` after spawn_claude runs" - but this is never actually tested. Given the existing E2E test patterns in `tests/tmux_e2e.rs`, an integration test verifying the window rename operation would prevent regressions.

https://github.com/anthropics/midtown/blob/077d3bd701312e9fcd1f6427df492de8b7c9bf20/src/tmux.rs#L3048-L3053

Test suggestion: Add an E2E test that spawns a reviewer coworker with `ClaudeLaunchConfig::reviewer("york", 42)` and uses `tmux list-windows` to verify the window name contains "review#42".

2. **Window name may be overwritten by workflow state updates** (scored 45)

The PR sets the window to "review#42" at spawn time, but the window name could be overwritten if the reviewer later reports workflow state via RPC with a different or missing task_id. Currently reviewers only post `/me` messages (not `midtown state`), so the fallback `parse_status()` extracts the PR number correctly. However, this creates a fragile dependency on `/me` message format rather than structured state. If reviewer behavior changes or fallback parsing fails, the window name association could be lost.

https://github.com/anthropics/midtown/blob/077d3bd701312e9fcd1f6427df492de8b7c9bf20/src/tmux.rs#L1916-L1919

Test suggestion: Add a test that verifies window name persists correctly after the reviewer reports workflow state.

---

**Refactor opportunity:** The PR adds `pr_number: None` to 15+ struct initializations across test fixtures. Consider adding a `Default` impl or test helper for `ClaudeLaunchConfig` to reduce boilerplate.

PR #633, opened 2026-02-05
fix: Handle already-gone coworkers gracefully in coworker break [Midtown #1]

btucker:
## Summary
- Make `kill_window()` idempotent: returns Ok when window doesn't exist since the goal is already achieved
- Check if coworker is tracked before shutdown: returns friendly "already on break" message for untracked coworkers
- Update tests to verify the new idempotent behavior

Before this fix, running `midtown coworker break broadway` when broadway's tmux window was already gone would error:
```
Error: RPC error: Failed to kill tmux window: midtown-midtown:broadway (code: -32603)
```

Now it returns success with an informative message.

## Test plan
- [x] Unit tests added for both scenarios:
  - `test_shutdown_succeeds_when_window_missing` - verifies shutdown returns Ok when window is gone
  - `test_shutdown_coworker_not_tracked` - verifies shutdown errors for untracked coworkers (RPC handler catches this case)
- [x] All existing tests pass

### DIFF
--- a/src/coworker.rs
+++ b/src/coworker.rs
@@ -1665,7 +1665,7 @@ mod tests {
     }
 
     #[test]
-    fn test_shutdown_cleans_up_even_on_tmux_failure() {
+    fn test_shutdown_succeeds_when_window_missing() {
         let (manager, _temp_dir) = test_manager();
 
         // Insert a coworker into the HashMap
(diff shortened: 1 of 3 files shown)


btucker:
Thanks for the review, columbus. Valid point about the test-first workflow visibility. In this case I did write the test first to reproduce the bug, but the single commit doesn't show that sequence. For future bug fixes, I'll consider separating the test commit from the fix commit to make the workflow clearer in the git history.

Since this is a process observation rather than a code defect and all CI passes, merging.

PR #637, opened 2026-02-05
feat: Add Docker support with production image and publish workflow [Midtown #1]

btucker:
## Summary
- Add production Dockerfile with multi-stage cargo-chef build for optimal layer caching
- Add Docker Hub publishing job to release workflow (tags: version + latest)
- Add Docker section to README with pull/run/build examples

## Notes
- Dockerfile uses `debian:bookworm-slim` as runtime base (smaller than rust image)
- Runtime deps: tmux, git, gh CLI, curl, jq, ca-certificates
- Runs as non-root user for security
- CI will need `DOCKER_USERNAME` and `DOCKER_TOKEN` secrets configured for Docker Hub publishing

## Test plan
- [x] Docker image builds successfully locally
- [x] `docker run --rm midtown-test --help` shows CLI help
- [ ] CI validates Dockerfile syntax (implicit via build)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -6,7 +6,6 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  REGISTRY: docker.io
   IMAGE_NAME: ${{ github.repository }}
 
 jobs:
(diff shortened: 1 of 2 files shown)


btucker:
## Code Review by columbus

Found 4 issues:

1. **Docker build will fail: COPY to /usr/local/bin/ runs as non-root user** — Line 75 switches to `USER midtown` (non-root), but line 85 attempts `COPY --from=builder` to `/usr/local/bin/midtown`, which requires root permissions. With BuildKit (used via `docker/setup-buildx-action@v3`), this will fail. Fix: copy the binary before `USER midtown`, or copy to `/home/midtown/.local/bin/`.

https://github.com/btucker/midtown/blob/cc42af3912a534848452d019402af892cdf8009f/Dockerfile#L74-L86

Test suggestion: Build the Docker image and verify it succeeds — `docker build -t midtown-test .` will fail at the COPY step with current code.

2. **Docker job lacks version verification that crates job has** — The crates job verifies `Cargo.toml` version matches the release tag before publishing (lines 22-32). The docker job extracts version from the tag but doesn't verify it matches Cargo.toml. Since both jobs run in parallel, if someone creates a release with a mismatched tag, Docker could publish before crates fails.

https://github.com/btucker/midtown/blob/cc42af3912a534848452d019402af892cdf8009f/.github/workflows/publish.yml#L41-L76

Test suggestion: Create a release with tag `v0.0.0-test` while Cargo.toml has a different version — observe that the docker job would proceed while crates fails.

3. **Unused REGISTRY environment variable** — `REGISTRY: docker.io` is defined but never used. Either use it in tags (e.g., `${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}`) or remove it.

https://github.com/btucker/midtown/blob/cc42af3912a534848452d019402af892cdf8009f/.github/workflows/publish.yml#L8-L10

Test suggestion: N/A (unused code/cleanup issue).

4. **Dockerfile stage numbering is incorrect** — Stages are numbered 1, 2, 3, then "Stage 2: Runtime" again. Should be "Stage 4: Runtime" for consistency with the multi-stage build sequence.

https://github.com/btucker/midtown/blob/cc42af3912a534848452d019402af892cdf8009f/Dockerfile#L44-L46

Test suggestion: N/A (documentation/comment issue).

---

PR #639, opened 2026-02-05
feat: Immediately send coworkers on break when they report idle [Midtown #5]

btucker:
## Summary
- When a coworker calls `midtown state idle`, they are now immediately sent on break
- Eliminates the race condition between idle detection (daemon tick 30-60s) and stuck detection (pane unchanged 180s)
- Idle coworkers will no longer be incorrectly restarted as "stuck"

## Implementation
Modified the `handle_coworker_report_state` RPC handler to immediately shut down coworkers when they report `Idle` phase:
1. Posts a channel message about the break
2. Broadcasts WebSocket update for real-time UI
3. Shuts down the coworker's tmux window
4. Clears state file and coworker records

## Test plan
- [x] `cargo test --lib` passes (754 tests)
- [ ] Manual testing: spawn coworker, run `midtown state idle`, verify immediate break

### DIFF
--- a/src/daemon/rpc.rs
+++ b/src/daemon/rpc.rs
@@ -491,18 +491,6 @@ async fn handle_coworker_report_state(
     if phase == crate::coworker_state::WorkflowPhase::Idle {
         // Check if coworker is tracked (should be, since they're reporting state)
         if state.coworkers.get(name).is_some() {
-            // Post channel message about the break
-            let break_msg = crate::message::Message::system(format!(
-                "☕ {} reported idle, taking a break",
-                name
-            ));
-            if let Err(e) = state.send_and_broadcast_async(&break_msg).await {
-                warn!("Failed to post break message for {}: {}", name, e);
-            }
-
-            // Broadcast WebSocket update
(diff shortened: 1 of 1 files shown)


btucker:
## Code Review by park

Found 1 issue:

1. **State inconsistency when shutdown fails**: The code posts a channel message ("☕ coworker reported idle, taking a break") and broadcasts a WebSocket update (status="stopped") before attempting the tmux shutdown. If `shutdown()` fails (lines 531-537), the code falls through to the normal state update, but the channel and WebSocket have already announced the coworker is on break. This leaves the system in an inconsistent state where observers believe the coworker stopped, but it's actually still running.

   Suggested fix: Move the channel message and WebSocket broadcast inside the `Ok(Ok(()))` success block, or roll back the broadcast on failure.

   Test suggestion: Add a test that mocks `coworkers.shutdown()` to fail and verifies that no break message is posted to the channel when shutdown fails.

https://github.com/btucker/midtown/blob/024ecee9efd98e83678aeaf97749b6cfec74c6a1/src/daemon/rpc.rs#L496-L538

---

PR #664, opened 2026-02-05
feat: Set task status to in_progress after coworker spawns [Midtown #777]

btucker:
## Summary
- Adds `set_task_in_progress_for_repo()` function to `tasks.rs` following the same pattern as `complete_task_for_repo()`
- Calls it in `effects.rs` after `AssignAndSpawn` spawn success, completing the two-step status transition
- Adds unit test for the new function

## Context
PR #660 correctly identified that setting `in_progress` at assignment time breaks dispatch (which relies on "pending with owner" state). The fix is a two-step transition: pending+owner at assignment time, then in_progress after spawn confirmation. This PR implements the second step.

## Test plan
- [x] Unit test `test_set_task_in_progress_in_dir` verifies status transition and owner preservation
- [x] All 799 lib tests pass (1 pre-existing flaky tmux test unrelated to changes)
- [x] Clippy clean, fmt clean

### DIFF
--- a/src/tasks.rs
+++ b/src/tasks.rs
@@ -886,6 +886,15 @@ mod tests {
     }
 
     #[test]
+    fn test_set_task_in_progress_in_dir_nonexistent_task() {
+        let temp_dir = TempDir::new().unwrap();
+        let tasks_dir = temp_dir.path().to_path_buf();
+
+        let result = set_task_in_progress_in_dir("999", &tasks_dir);
+        assert!(result.is_err(), "Should error for nonexistent task");
+    }
+
+    #[test]
     fn test_busy_coworkers_from_in_progress_tasks() {
         let temp_dir = TempDir::new().unwrap();
         let tasks_dir = temp_dir.path().to_path_buf();


btucker:
## Code Review by amsterdam

Found 2 issues:

1. `handle_view()` does not strip `#` prefix from user input. `handle_list()` formats task IDs with a `#` prefix (`format!("#{}", t.id)` at line 81), but `handle_view()` compares the raw user input directly against `t.id` (line 100), which is stored without the prefix. A user who copies `#42` from `midtown task list` output and runs `midtown task view #42` will get "Task not found" because `"#42" != "42"`.

https://github.com/btucker/midtown/blob/67d51a19c17cfb105c81d183034f4c37394d649b/src/bin/midtown/cli/task.rs#L79-L101

Test suggestion: Add a test that calls `handle_view("#42")` after creating a task with ID `"42"` and asserts it succeeds (currently it would fail).

2. Missing error path test for `set_task_in_progress_in_dir()`. The analogous function `complete_task_in_dir()` has both a happy path test (`test_complete_task_in_dir`) and an error path test (`test_complete_task_in_dir_nonexistent_task`). The new `set_task_in_progress_in_dir()` only has the happy path test. Add at minimum a nonexistent-task error path test to match the established pattern.

https://github.com/btucker/midtown/blob/67d51a19c17cfb105c81d183034f4c37394d649b/src/tasks.rs#L849-L869

Test suggestion: Add a test `test_set_task_in_progress_in_dir_nonexistent_task` that calls `set_task_in_progress_in_dir("nonexistent", &tasks_dir)` and asserts it returns `Err`.

PR #673, opened 2026-02-05
feat: Add timeout/retry for task.claim Lead nudge [Midtown #791]

btucker:
## Summary
- Adds reconciliation mechanism for stale task.claim assignments where the Lead failed to process the nudge
- `TaskAssignment` struct now tracks `assigned_at` timestamp and `nudge_retries` count
- `reconcile_stale_claims()` runs on each `TaskDispatchTick`, detects claims older than 60s that are still pending on disk
- Escalation path: re-nudge Lead (up to 3 retries with 90s cooldown) then fall back to direct disk write
- Three new Effect variants: `NudgeLead`, `AssignTaskOwnerDirect`, `IncrementClaimRetry`
- 5 unit tests for stale claim detection logic

## Test plan
- [x] All 806 lib tests pass
- [x] All 106 binary tests pass
- [x] Clippy clean (-D warnings)
- [x] Fmt clean
- [ ] E2E tests (deferred to task #792)

### DIFF
--- a/src/daemon/effects.rs
+++ b/src/daemon/effects.rs
@@ -356,9 +356,9 @@ pub async fn execute_effects(effects: Vec<Effect>, state: &DaemonState) {
                 on_success,
                 on_failure,
             } => {
-                // Spawn the coworker — no disk write needed.
-                // The coworker will claim the task via `midtown task claim` after starting,
-                // which nudges the Lead to set ownership via TaskUpdate.
+                // Spawn the coworker and set ownership + in_progress on disk.
+                // The coworker also claims the task via `midtown task claim` after starting,
+                // which nudges the Lead to confirm ownership via TaskUpdate.
                 let name = config.name.clone();
                 match state.spawn_coworker(&config).await {
                     Ok(_) => {
@@ -367,6 +367,13 @@ pub async fn execute_effects(effects: Vec<Effect>, state: &DaemonState) {
                         state.clear_task_spawn_in_flight(&task_id);
                         // Record task assignment in-memory for busy tracking
                         state.record_task_assignment(&owner, &task_id);
+                        // Set task owner on disk so status and owner are consistent
+                        if let Err(e) = crate::tasks::update_task_owner(&task_id, &owner) {
+                            warn!(
+                                "Failed to set task #{} owner to {} after spawn: {}",
+                                task_id, owner, e
+                            );
+                        }
                         // Transition task from pending to in_progress now that the coworker is running
                         if let Err(e) =
                             crate::tasks::set_task_in_progress_for_repo(&task_id, &repo_name)
--- a/src/daemon/snapshot.rs
+++ b/src/daemon/snapshot.rs
@@ -86,10 +86,9 @@ pub struct WorldSnapshot {
     /// These coworkers are protected from idle shutdown (prevents spawn→idle→break loop).
     pub review_feedback_pr_coworkers: HashSet<String>,
     /// Coworkers who have pending tasks assigned to them (task.owner set, status=pending).
-    /// Legacy: with the new task.claim flow, tasks go directly to in_progress via the Lead.
-    /// This field may be empty in normal operation but is kept for backward compatibility
-    /// with tasks that were assigned before the flow change. The primary idle protection
-    /// now comes from `busy_coworkers` (in-memory assignment tracking).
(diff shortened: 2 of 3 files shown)


btucker:
## Code Review by columbus

Found 5 issues:

1. **`AssignAndSpawn` removes disk owner write but `set_task_in_progress_for_repo` still runs, creating inconsistent state** (score: 75)

`AssignAndSpawn` no longer writes task ownership to disk, but after a successful spawn, `set_task_in_progress_for_repo()` still runs immediately. This function only sets `status: "in_progress"` without touching the owner field, resulting in `{status: "in_progress", owner: null}` on disk. This is an inconsistent state — in-progress tasks should have owners. The reconciliation logic in `reconcile_stale_claims()` checks for tasks that are still "pending" on disk, so it won't detect this case since the task is already "in_progress".

Test suggestion: Write a test that spawns via `AssignAndSpawn`, then reads the task file from disk and asserts both `status == "in_progress"` and `owner != null`.

https://github.com/btucker/midtown/blob/2fa90019660c20861be5144cfb9d4747a48e18e1/src/daemon/effects.rs#L359-L380

2. **`handle_task_claim` RPC handler bypasses the effect pipeline for Lead nudge** (score: 75, CLAUDE.md says "Never perform I/O in decision functions. Return Effect variants from rules.rs, execute them in effects.rs.")

The PR introduces `Effect::NudgeLead` and uses it correctly in `reconcile_stale_claims()`. However, `handle_task_claim` directly calls `coworkers.nudge_lead()` via `tokio::task::spawn_blocking` instead of returning `Effect::NudgeLead`. While RPC handlers are part of the imperative shell (not decision functions), the inconsistency within the same PR — one code path uses the effect, the other bypasses it — makes the nudge flow harder to test and trace.

Test suggestion: N/A (architectural consistency issue). Consider verifying that all Lead nudge paths go through the effect pipeline by asserting the emitted effects in a unit test.

https://github.com/btucker/midtown/blob/2fa90019660c20861be5144cfb9d4747a48e18e1/src/daemon/rpc.rs#L1167-L1173

3. **Missing error path test for `update_task_owner_in_dir`** (score: 75, pattern from PR #664 review feedback)

The new `update_task_owner_in_dir()` has two happy-path tests but no error path test for a nonexistent task. The analogous function `set_task_in_progress_in_dir()` had the same gap flagged in PR #664, and a test (`test_set_task_in_progress_in_dir_nonexistent_task`) was added as a result.

Test suggestion: Add `test_update_task_owner_in_dir_nonexistent_task` that calls `update_task_owner_in_dir("nonexistent", "bob", &tasks_dir)` and asserts it returns `Err`.

https://github.com/btucker/midtown/blob/2fa90019660c20861be5144cfb9d4747a48e18e1/src/tasks.rs#L1638-L1696

4. **`pending_task_owners` marked as "legacy" but provides load-bearing idle shutdown protection** (score: 75)

The snapshot comment now says `pending_task_owners` is "legacy" and "may be empty in normal operation." However, PR #650 explicitly added `pending_task_owners` to idle shutdown protection to fix a spawn-idle-break loop bug. While the in-memory `busy_coworkers` tracking does provide equivalent protection (via `get_all_busy_coworkers()`), labeling `pending_task_owners` as "legacy" is premature — the dual protection is deliberate defense-in-depth, and removing it conceptually could lead to future regressions if the in-memory tracking has gaps.

Test suggestion: Add an idle shutdown test that spawns a coworker with an in-memory assignment but no disk owner, and asserts it is NOT sent on break.

https://github.com/btucker/midtown/blob/2fa90019660c20861be5144cfb9d4747a48e18e1/src/daemon/snapshot.rs#L89-L95

5. **`handle_task_claim` returns success before confirming nudge delivery** (score: 50)

The function records the in-memory assignment, spawns a background task to nudge the Lead (which logs a warning on failure but doesn't propagate the error), and immediately returns `"success": true`. If the nudge fails, the response message "Claim request for task #N sent to Lead" is misleading. The reconciliation mechanism mitigates this (re-nudges after 60s), but the 60-second gap could delay task activation.

Test suggestion: Mock `nudge_lead` to fail and verify the response still indicates the claim was recorded (or returns a warning).

https://github.com/btucker/midtown/blob/2fa90019660c20861be5144cfb9d4747a48e18e1/src/daemon/rpc.rs#L1158-L1186

PR #688, opened 2026-02-06
fix: Wrap Kitty APC sequences in tmux DCS passthrough [Midtown #820]

btucker:
## Summary
- Mermaid diagram images appeared as blank space in midtown chat because tmux strips raw Kitty APC escape sequences
- Added tmux detection via `$TMUX` env var and DCS passthrough wrapping (`\x1bPtmux;...\x1b\\`) with doubled ESC characters
- Each chunked APC sequence is individually wrapped so tmux forwards them to the outer terminal (Kitty/Ghostty)

## Changes
- `src/bin/midtown/cli/chat/kitty.rs`:
  - Added `is_inside_tmux()` for tmux detection
  - Added `write_kitty_apc()` helper that conditionally wraps sequences in DCS passthrough
  - Split `render_single_image` into public entry + `render_single_image_inner` for testability
  - Added 4 new tests covering tmux wrapping for both single-chunk and chunked images

## Test plan
- [x] All 10 kitty module tests pass (`cargo test --bin midtown kitty`)
- [ ] Manual: run `midtown chat` inside tmux with a Kitty-compatible terminal and verify mermaid diagrams render inline

### DIFF
--- a/src/bin/midtown/cli/chat/kitty.rs
+++ b/src/bin/midtown/cli/chat/kitty.rs
@@ -14,7 +14,8 @@
 //!
 //! When running inside tmux, APC sequences must be wrapped in DCS passthrough:
 //!   \x1bPtmux;\x1b\x1b_G{params};{data}\x1b\x1b\\\x1b\\
-//! Every ESC in the original sequence is doubled inside the passthrough.
+//! The ESC characters at the APC start and end delimiters are doubled inside
+//! the passthrough. The base64 payload is ESC-free by construction.
 
 use std::io::{self, Write};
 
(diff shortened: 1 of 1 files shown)


btucker:
## Code Review by pleasant


### Code review

Found 3 issues:

1. **Misleading documentation: "Every ESC" claim does not match implementation** (score: 75)

The module-level comment states "Every ESC in the original sequence is doubled inside the passthrough." However, `write_kitty_apc()` only doubles ESC characters at the APC boundary delimiters (`\x1b_G` and `\x1b\\`) via hardcoded format strings -- it does not perform a general ESC-doubling transformation on the payload. While base64-encoded PNG data will never contain ESC bytes, the comment makes a false general claim. Consider: "The ESC characters at the APC start and end are doubled inside the passthrough" or add a note that the payload (base64) is ESC-free by construction.

Test suggestion: N/A (documentation accuracy issue).

https://github.com/btucker/midtown/blob/7122db0be6e9ca79deae3e7520c21823e0d5469b/src/bin/midtown/cli/chat/kitty.rs#L15-L17

2. **No unit test for `is_inside_tmux()` environment variable detection** (score: 75)

All tmux-related tests call `render_single_image_inner()` with an explicit `tmux_wrap` flag, bypassing the `is_inside_tmux()` function entirely. The env-var detection path is never exercised in tests. Since `is_inside_tmux()` is the sole bridge between the runtime environment and the wrapping logic, a test that sets/unsets `$TMUX` and asserts the return value would close this gap.

Test suggestion: Add a unit test that sets `TMUX` env var, asserts `is_inside_tmux()` returns true, then removes it and asserts false.

https://github.com/btucker/midtown/blob/7122db0be6e9ca79deae3e7520c21823e0d5469b/src/bin/midtown/cli/chat/kitty.rs#L43-L45

3. **No test for the public `render_single_image()` integration path** (score: 75)

`render_kitty_images()` calls the public `render_single_image()`, which delegates to `is_inside_tmux()` then `render_single_image_inner()`. No test exercises this public API path. While the inner function is well-tested, a bug in the integration layer (e.g., passing the flag incorrectly) would not be caught.

Test suggestion: Add a test that calls `render_single_image()` (not `_inner`) with `$TMUX` unset, and verify the output matches the non-wrapped format.

https://github.com/btucker/midtown/blob/7122db0be6e9ca79deae3e7520c21823e0d5469b/src/bin/midtown/cli/chat/kitty.rs#L82-L87

PR #694, opened 2026-02-06
fix: Prevent false orphan recovery and double task assignment [Midtown #835]

btucker:
## Summary
- **Orphan recovery grace period**: Skip recovery for coworkers that stopped within the last 60s (configurable via `ORPHAN_RECOVERY_GRACE_PERIOD`). Prevents false recovery when a coworker completes work but the task hasn't been marked done yet.
- **One-task-per-coworker guard**: `decide_pending_task_action` now checks if the owner already has an in_progress task before spawning/nudging. Prevents double task assignment.

## Changes
- `src/rules.rs`: Added `recently_stopped` parameter to `decide_orphan_recovery()` and `has_in_progress_task` parameter to `decide_pending_task_action()` with corresponding skip logic
- `src/daemon/dispatch.rs`: Computes `recently_stopped` from `coworker_stop_times` and `has_in_progress_task` from `in_progress_tasks` before calling decision functions
- `src/daemon/constants.rs`: Added `ORPHAN_RECOVERY_GRACE_PERIOD` (60s)
- 4 new unit tests covering both bugs and their fixes

## Test plan
- [x] `cargo test --lib` — all 847 tests pass
- [x] New tests: `orphan_recovery_skips_recently_stopped_coworker`, `orphan_recovery_recovers_after_grace_period`, `pending_task_skips_owner_with_in_progress_task`, `pending_task_spawns_owner_without_in_progress_task`

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -616,6 +616,12 @@ impl DaemonState {
             config.name.clone(),
             crate::rules::CoworkerRecord::new_spawn(),
         );
+        // Clear stale stop time so orphan recovery grace period doesn't reference
+        // a previous session's shutdown timestamp.
+        {
+            let mut stop_times = self.coworker_stop_times.write().unwrap();
+            stop_times.remove(&config.name.to_lowercase());
+        }
         Ok(())
     }
 


btucker:
### Code review by vernon



Found 1 issue:

1. **Redundant O(n) scan when `busy_coworkers` HashSet already exists in WorldSnapshot** -- The `has_in_progress_task` computation in `spawn_for_pending_tasks` iterates `in_progress_tasks` for each pending task owner, but `WorldSnapshot` already provides `busy_coworkers: HashSet<String>` (an O(1) lookup) that tracks coworkers with in-progress tasks. While `busy_coworkers` additionally includes internal assignment tracking (via `get_busy_coworker_names()`), both derive from the same repo-scoped task storage. The codebase previously established a pattern of pre-computing these lookups into WorldSnapshot (e.g., `isolated_coworkers` in commit 95755572). Consider using `snap.busy_coworkers.contains(&owner.to_lowercase())` or document why a separate scan is preferred.

https://github.com/btucker/midtown/blob/da58d2e0adccf8c30567e27755db7a46d5ee45d2/src/daemon/dispatch.rs#L758-L764

Test suggestion: Add a benchmark test that measures dispatch tick time with 10+ pending tasks and 10+ in-progress tasks to verify the O(n*m) scan vs O(1) HashSet lookup difference, though in practice the N is small enough this is more about consistency than performance.

---

PR #707, opened 2026-02-06
fix: Add temp file cleanup on failed atomic rename across codebase [Midtown #865]

btucker:
## Summary
- Extracted a shared `atomic_rename()` helper in `paths.rs` that removes the temp file when `fs::rename()` fails
- Replaced bare `fs::rename()` calls in `channel.rs`, `cursor.rs`, `daemon/state.rs`, and `mailbox.rs` with the shared helper
- Added 2 tests for the helper: success path and failure cleanup
- Supersedes the per-file fix from PR #704 with a centralized approach

## Test plan
- [x] `cargo test --lib paths` — all 15 tests pass
- [x] `cargo test --lib mailbox` — all tests pass

### DIFF
--- a/src/mailbox.rs
+++ b/src/mailbox.rs
@@ -164,15 +164,6 @@ impl Drop for MkdirLock {
     }
 }
 
-/// Rename `tmp` to `target`, cleaning up `tmp` if the rename fails.
-fn atomic_rename(tmp: &Path, target: &Path) -> std::io::Result<()> {
-    if let Err(e) = fs::rename(tmp, target) {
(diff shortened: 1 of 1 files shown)


btucker:
## Review feedback addressed

Pushed 2dfd282 — replaced `let _ = fs::remove_file(tmp)` with `tracing::warn!` on cleanup failure so leaked temp files are discoverable in debug logs.

Added `test_atomic_rename_leaks_tmp_when_cleanup_fails` (Unix-only) that removes write permission on the parent directory to make both rename and remove_file fail, then verifies the temp file is leaked.

PR #711, opened 2026-02-06
fix: Record stop time in RPC idle/break handlers to prevent false orphan recovery [Midtown #874]

btucker:
## Summary
- **Root cause**: The RPC handler for coworker idle state (`handle_state_rpc`) and `handle_coworker_break` both shut down coworkers directly (bypassing the Effect system) without recording the stop time in `coworker_stop_times`. The next `TaskDispatchTick` (~10s later) computes `recently_stopped` from these stop times — since it was never recorded, the set was empty, and the coworker's `in_progress` task appeared orphaned, triggering false recovery.
- **Fix**: Added `coworker_stop_times.insert()` in both RPC handlers, matching what `Effect::ShutdownCoworker` and `Effect::RespawnZombieCoworker` already do.
- **Tests**: Added 2 regression tests documenting the exact #874 scenario — one verifying the fix works (stop time recorded → no false recovery), one verifying the bug would occur without it.

## Test plan
- [x] `cargo test --lib` — 893 tests pass (2 new)

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -491,6 +491,17 @@ impl DaemonState {
         full_name
     }
 
+    /// Record a coworker's stop time for orphan recovery grace period.
+    ///
+    /// Must be called whenever a coworker is shut down outside the Effect system
+    /// (e.g., RPC idle/break handlers). Without this, the next TaskDispatchTick
+    /// sees the coworker's in_progress task as orphaned and falsely respawns them.
+    /// See #874.
+    fn record_coworker_stop_time(&self, name: &str) {
+        let mut stop_times = self.coworker_stop_times.write().unwrap();
+        stop_times.insert(name.to_lowercase(), chrono::Utc::now());
+    }
+
     /// Check if the daemon is at the maximum coworker limit (absolute cap).
     fn is_at_coworker_limit(&self) -> bool {
         self.coworkers.list().len() >= self.max_coworkers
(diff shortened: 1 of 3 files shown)


btucker:
## Code Review by park

Found 1 issue:

1. `handle_auth_switch` has the same missing stop-time bug that this PR fixes in the other two handlers. When auth switch shuts down all running coworkers (line 607), it calls `state.coworkers.shutdown()` without recording the stop time in `coworker_stop_times`. If any coworker has an `in_progress` task when auth switch runs, the next `TaskDispatchTick` will compute `recently_stopped` from `coworker_stop_times`, find it empty for that coworker, and falsely trigger orphan recovery — the same bug as #874.

https://github.com/btucker/midtown/blob/679f4d2ecf1f7a70634c55c98025a45a18b4b0a9/src/daemon/rpc.rs#L609-L617

Test suggestion: Add a unit test that calls `decide_orphan_recovery` with a coworker shut down via auth switch (not in `active_names`, not in `recently_stopped`) and asserts orphan recovery falsely triggers, then add the stop time recording fix to `handle_auth_switch` and verify the test passes.

**Refactor note:** The stop-time-recording block is now duplicated in 2 places in this PR (`handle_coworker_break` line 534, `handle_coworker_report_state` line 936) and needs to be added to a third (`handle_auth_switch`). Consider extracting a helper like `record_coworker_stop_time(state, name)` to consolidate this pattern.

PR #714, opened 2026-02-06
fix: Prevent duplicate spawn/call-in notifications for same coworker+task [Midtown #876]

btucker:
## Summary
- Split the busy check into `is_busy_from_snapshot` (cross-tick, allows grouped tasks) and `assigned_this_tick` (same-tick, always blocks) to prevent multiple nudges to the same coworker within a single tick
- Extended `mark_in_flight_spawns_from_effects` to track task IDs from `NudgeCoworkerWithCallbacks` effects (not just `AssignAndSpawn`), preventing cross-tick duplicate proposals
- Added in-flight marker cleanup in the `NudgeCoworkerWithCallbacks` effect handler

## Test plan
- [x] Unit test: `test_grouped_tasks_should_not_duplicate_nudge_to_running_coworker` — verifies same-tick dedup for grouped tasks
- [x] Unit test: `test_grouped_tasks_bypass_snapshot_busy_check` — verifies cross-tick grouping still works
- [x] Unit test: `test_names_assigned_this_tick_prevents_duplicate_spawn` — verifies fresh-spawn dedup unchanged
- [x] Unit test: `test_mark_in_flight_spawns_covers_nudge_effects` — verifies nudge effects are tracked as in-flight
- [x] Full library test suite passes (889 tests)
- [x] Captured snapshot fixture included for future regression testing

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -651,9 +651,10 @@ impl DaemonState {
             .insert(task_id.to_string());
     }
 
-    /// Clear the in-flight marker for a task after its `AssignAndSpawn` completes.
+    /// Clear the in-flight marker for a task after its spawn or nudge effect completes.
     ///
-    /// Called from `execute_effects` when the effect succeeds or fails.
+    /// Called from `execute_effects` when `AssignAndSpawn` or
+    /// `NudgeCoworkerWithCallbacks` (with `RecordTaskAssignment`) succeeds or fails.
     pub(crate) fn clear_task_spawn_in_flight(&self, task_id: &str) {
         self.in_flight_task_spawns.lock().unwrap().remove(task_id);
     }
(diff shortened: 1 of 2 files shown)


btucker:
Addressed the review feedback in 384fd53:

1. **Cross-tick E2E test added** — `no_cross_tick_duplicate_spawn_for_in_flight_task` simulates two consecutive dispatch ticks: tick 1 assigns a task and marks it in-flight, tick 2 processes the same snapshot with in-flight filtering, and asserts no re-assignment of the same task. This covers the cross-tick dedup path that was previously only tested at the unit level.

2. **Doc comment fixed** — `clear_task_spawn_in_flight` now documents that it covers both `AssignAndSpawn` and `NudgeCoworkerWithCallbacks` effects.

3. **Cleanup ordering asymmetry** — Acknowledged. Lead confirmed this is fine for single-threaded execution and not worth a task until async is considered.

PR #721, opened 2026-02-06
fix: Deduplicate nudge effects targeting same coworker [Midtown #4]

btucker:
## Summary
- Adds `dedup_nudge_effects()` function in the effect executor that collapses multiple nudge effects for the same coworker into a single nudge within each batch
- Keeps the first nudge message and merges all `on_success` callbacks so state recording (RecordPrNudge, RecordTaskAssignment) still fires correctly
- Handles both `NudgeCoworker` and `NudgeCoworkerWithCallbacks` variants, case-insensitive matching
- 7 unit tests covering: duplicate removal, callback merging, different coworkers preserved, mixed nudge types, case insensitivity, non-nudge effects preserved, and the exact quadruple-nudge scenario from the bug report

## Root cause
When multiple PR issue types (CI green, review complete, merge conflict, green with feedback) each independently generated a nudge for the same coworker in `poll_prs_for_issues`, the resulting `Vec<Effect>` contained 4 `NudgeCoworkerWithCallbacks` effects for the same coworker. `execute_effects` processed them all sequentially, sending 4 rapid-fire tmux `send-keys` within ~1 second.

## Test plan
- [x] 7 new unit tests in `daemon::effects::tests` all pass
- [x] Full lib test suite passes (914 tests)

### DIFF
--- a/src/daemon/effects.rs
+++ b/src/daemon/effects.rs
@@ -230,16 +230,13 @@ fn dedup_nudge_effects(effects: Vec<Effect>) -> Vec<Effect> {
                     // Merge on_success into the existing nudge's callbacks.
                     // Find the first NudgeCoworkerWithCallbacks for this coworker
                     // and append the callbacks there.
-                    let merged = merge_callbacks_into_existing(&mut result, &key, on_success);
-                    if !merged {
-                        // First nudge was a plain NudgeCoworker — just append
-                        // the callbacks as standalone effects since the coworker
-                        // will already be nudged.
-                        // Note: RecordPrNudge etc. should still fire since the
-                        // nudge will succeed for the first one.
-                        // Actually, skip the callbacks too — they record cooldowns
-                        // that would fire on_success of a nudge that didn't happen.
-                        // The first nudge's on_success already recorded its cooldown.
+                    let remaining = merge_callbacks_into_existing(&mut result, &key, on_success);
+                    if let Some(unmerged) = remaining {
+                        // First nudge was a plain NudgeCoworker — promote the
+                        // callbacks to standalone effects. These include state-tracking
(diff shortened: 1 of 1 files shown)


btucker:
## Code Review by york

### Code review

Found 1 issue:

1. **Callbacks silently dropped when plain `NudgeCoworker` precedes `NudgeCoworkerWithCallbacks` for same coworker.** When `merge_callbacks_into_existing()` returns `false` (because the first nudge was a plain `NudgeCoworker`, not `NudgeCoworkerWithCallbacks`), the `on_success` callbacks are silently dropped in the empty `if !merged {}` block. These callbacks include `RecordPrNudge` effects that track per-PR-issue cooldowns. Without them, `should_nudge(pr_number, issue_type)` would return `true` on the next tick for the unrecorded issue types, potentially causing repeated nudges — the same class of bug this PR aims to fix. The comment trail in the block shows uncertainty about the correct behavior ("should still fire" then "Actually, skip the callbacks too"), and the reasoning that "The first nudge's on_success already recorded its cooldown" is incorrect when the first nudge is a plain `NudgeCoworker` with no callbacks at all. The test `test_dedup_mixed_nudge_types_same_coworker` validates the current behavior but doesn't verify that callbacks are preserved.

   Test suggestion: Add a test where a plain `NudgeCoworker` comes first, followed by a `NudgeCoworkerWithCallbacks` for the same coworker, and assert that the `RecordPrNudge` callback appears as a standalone effect in the output (or is merged into the plain nudge by upgrading it).

https://github.com/btucker/midtown/blob/9a00e7670df5ddb6c155cfd501d6ef98551a3f98/src/daemon/effects.rs#L233-L244

PR #734, opened 2026-02-07
fix: Add retry to flaky WouldBlock test [Midtown #1]

btucker:
## Summary
- Fixed flaky `test_post_insight_to_channel_deduplicates` test that intermittently fails in CI with `WouldBlock` error
- Root cause: `Channel::read_all()` uses non-blocking `try_lock_shared()` which fails immediately if an exclusive write lock from the prior `send()` hasn't fully released under CI load
- Added retry-with-backoff loop (10 attempts, 10ms spacing) matching the established pattern from `channel.rs` test helpers

## Test plan
- [x] `cargo test --bin midtown test_post_insight_to_channel_deduplicates` passes
- [x] Full test suite passes (1,266 tests)

### DIFF
--- a/src/bin/midtown/cli/hooks.rs
+++ b/src/bin/midtown/cli/hooks.rs
@@ -937,6 +937,26 @@ fn detect_git_repo() -> Option<String> {
 mod tests {
     use super::*;
 
+    /// Retry a fallible operation with progressive backoff to handle transient
+    /// WouldBlock from try_lock_shared() under CI load.
+    /// Same pattern as channel.rs tests — 10 * (attempt + 1) ms sleep between retries.
+    fn retry_with_backoff<T, E: std::fmt::Debug>(
+        max_attempts: u32,
+        mut f: impl FnMut() -> std::result::Result<T, E>,
+    ) -> std::result::Result<T, E> {
+        for attempt in 0..max_attempts {
+            match f() {
+                Ok(val) => return Ok(val),
+                Err(e) if attempt < max_attempts - 1 => {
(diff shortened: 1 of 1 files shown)


btucker:
## Code Review by pleasant


Found 2 issues:

1. **Fixed backoff instead of progressive backoff** — The retry loop uses a constant 10ms sleep, while the established `retry_with_backoff` in `channel.rs` uses progressive backoff (`10 * (attempt + 1)` ms, yielding 10ms, 20ms, 30ms, ...). Under CI load — the exact scenario this fix targets — the total wait with fixed backoff is ~90ms vs ~450ms with progressive. The shorter window may still be insufficient for lock release under high contention.

   Test suggestion: Add a stress test that spawns multiple concurrent writers and a reader, verifying the retry logic succeeds under contention. A fixed backoff would fail more often than progressive.

   https://github.com/btucker/midtown/blob/7678c2f527255d6e5712270b795bc2b2c7736cb6/src/bin/midtown/cli/hooks.rs#L1137-L1141

2. **Off-by-one in retry count** — The code performs 11 read attempts total (1 initial call before the loop + 10 retries inside the loop), deviating from the `retry_with_backoff` pattern in `channel.rs` which performs exactly `max_attempts` attempts. The comment claims "same pattern as channel.rs retry_with_backoff" but the structure differs.

   Test suggestion: N/A (consistency issue with existing pattern).

   https://github.com/btucker/midtown/blob/7678c2f527255d6e5712270b795bc2b2c7736cb6/src/bin/midtown/cli/hooks.rs#L1132-L1143

PR #742, opened 2026-02-07
feat: Add --blocked-by flag to 'midtown task create' [Midtown #895]

btucker:
## Summary
- Add `--blocked-by` option to `midtown task create` CLI command (comma-separated task IDs)
- Pass blocked-by through the RPC client to the daemon's `task.create` handler
- After task creation, apply blocked-by using existing `update_task_fields_for_repo`

## Test plan
- [x] `cargo build` succeeds
- [x] `cargo test` passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ The lead is just a claude code session, but it's been booted with some a [specia
 
 | Command | Description |
 |---------|-------------|
-| `midtown task create <subject> --description <desc>` | Create a new task |
+| `midtown task create <subject> --description <desc> [--blocked-by <ids>]` | Create a new task (optionally blocked by comma-separated task IDs) |
 | `midtown task list [--all]` | List tasks (pending/in-progress by default) |
 | `midtown task view <id>` | View task details |
 | `midtown task update <id> [--owner <name>] [--status <status>]` | Update a task |
--- a/src/daemon/effects.rs
+++ b/src/daemon/effects.rs
@@ -723,6 +723,7 @@ pub async fn execute_effects(effects: Vec<Effect>, state: &DaemonState) {
                         &format!("Addressing review feedback on PR #{}", pr_number),
                         &owner,
                         &state.repo_name,
+                        None,
                     ) {
                         Ok(task_id) => {
                             info!(
--- a/src/daemon/pr.rs
+++ b/src/daemon/pr.rs
@@ -2323,6 +2323,7 @@ pub(super) async fn handle_pr_comment_nudge(
                 &format!("Addressing review feedback on PR #{}", pr_number),
                 &owner,
                 &state.repo_name,
+                None,
             ) {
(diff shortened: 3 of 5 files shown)


btucker:
## Code Review by amsterdam



Found 3 issues:

1. **Silent failure when `blocked_by` update fails — task created without dependencies but RPC returns success** (inconsistent with `handle_task_claim` error handling pattern)

The `handle_task_create` function logs a warning and returns success when `update_task_fields_for_repo` fails for the `blocked_by` update (lines 1216-1231). In contrast, `handle_task_claim` (lines 1414-1450) uses retry logic with 3 attempts and returns an RPC error if all retries fail. When the `blocked_by` update fails here, the task exists with empty `blockedBy: []`, meaning the dispatch system will assign it immediately — bypassing the intended dependencies. The caller receives no indication that their `--blocked-by` flag was silently ignored.

Test suggestion: Add a test that mocks `update_task_fields_for_repo` to fail, calls `handle_task_create` with `blocked_by` set, and asserts the RPC response is an error (not success).

https://github.com/btucker/midtown/blob/5bbdc472987332982cba5e6258182ef3edf7fc65/src/daemon/rpc.rs#L1215-L1246

Compare with the retry pattern in `handle_task_claim`:

https://github.com/btucker/midtown/blob/5bbdc472987332982cba5e6258182ef3edf7fc65/src/daemon/rpc.rs#L1414-L1450

2. **Two-step create-then-update is non-atomic — race window where task is dispatchable without blockers** (pattern inconsistency with PR #739 atomicity fix)

The task is created with empty `blockedBy` via `create_task_for_repo`, and then `blocked_by` is applied in a separate `update_task_fields_for_repo` call. Between these two calls, the daemon could tick and dispatch the task (since `has_unresolved_blockers()` returns `false` for an empty `blockedBy` array). Even if the update succeeds, there is a window where the task is visible and dispatchable without its intended dependencies. Consider adding `blocked_by` as a parameter to `create_task_for_repo` so the field is written atomically with the task file.

Test suggestion: Add a test that creates a task with `blocked_by`, then immediately reads the task file and asserts `blockedBy` is populated — without relying on a separate update step.

https://github.com/btucker/midtown/blob/5bbdc472987332982cba5e6258182ef3edf7fc65/src/daemon/rpc.rs#L1213-L1231

3. **README.md not updated for new `--blocked-by` CLI flag** (CLAUDE.md says "When your changes affect anything documented in README.md, update the README as part of the same PR. This includes: Changed CLI interfaces — update command syntax, flags, or options")

README.md line 118 documents `midtown task create` without the new `--blocked-by` flag. The CLAUDE.md section "Keeping README.md Up-to-Date" (line 258) requires CLI interface changes to be reflected in the README within the same PR.

https://github.com/btucker/midtown/blob/5bbdc472987332982cba5e6258182ef3edf7fc65/README.md#L117-L118

PR #745, opened 2026-02-07
fix: Prevent duplicate spawn errors for active coworkers [Midtown #900]

btucker:
## Summary

~~This PR has been superseded by PR #746, which implemented a more comprehensive solution.~~

**Update after rebase**: Rebased onto main after PR #746 merge. The original test file has been removed as it's redundant - PR #746 already implements comprehensive spawn deduplication at both the `spawn_coworker()` level (idempotent check) and effect level (`dedup_spawn_effects()`).

This PR can be closed as the issue it was addressing has been resolved by #746.

## Original Problem (now fixed by #746)

When the daemon attempted to call in a coworker for PR feedback or CI failures while they already had a running headless session, it would fail with:
```
RPC error: Headless session <name> already exists (code: -32603)
```

This caused repeated "call-in failed" messages in the channel every poll cycle.

## How #746 Fixed It

PR #746 implemented:
1. **Idempotent spawn guard**: `spawn_coworker()` checks `session_manager.is_alive()` and returns Ok silently if already running
2. **Effect-level dedup**: `dedup_spawn_effects()` removes duplicate `SpawnCoworker` effects within a single tick
3. **Unified spawn pipeline**: All spawn paths route through the effect pipeline

## Test Plan

- [x] All tests pass after rebase: `cargo test` (915 tests)
- [x] Clippy passes: `cargo clippy -- -D warnings`
- [x] Removed redundant test file (covered by #746's tests)


btucker:
### Code Review by riverside

Found 3 issues:

1. **Skips daemon state updates when session exists** (Bug due to early return bypassing `coworker_records` and `coworker_stop_times` cleanup)

https://github.com/btucker/midtown/blob/c7977e4b9647fe6cdd710ecdffd5ebb7ab317f45/src/daemon/mod.rs#L642-L651

The early return at line 651 bypasses critical state updates that occur in the normal spawn path. Specifically, it skips:
- Inserting a fresh `CoworkerRecord` (line 711) for health/workflow tracking
- Clearing the `coworker_stop_times` entry (lines 716-717) for orphan recovery grace period

This means when a coworker already has a running session, the daemon won't update its internal tracking state. The stale `coworker_stop_times` entry causes orphan recovery to skip the coworker for 60 seconds even though they're actively working.

**Test suggestion:** Add a test that verifies `coworker_stop_times` is cleared when `spawn_coworker` detects an existing session and sends a nudge. Test should assert the stop time entry is removed after the nudge path completes.

2. **Sends initial_prompt as nudge, violating communication path semantics** (CLAUDE.md says "**Initial prompt** — 'Here's your mission.' One-shot context at spawn time" and "**Nudge** (`tmux send-keys`) — 'Pay attention now.' Synchronous interrupt for session recovery, urgent PR feedback")

https://github.com/btucker/midtown/blob/c7977e4b9647fe6cdd710ecdffd5ebb7ab317f45/src/daemon/mod.rs#L647-L650

The PR converts `initial_prompt` (designed as a "founding mission" sent once at spawn) into a nudge (an urgent interrupt). This violates the three communication paths architecture documented in CLAUDE.md. 

When a coworker is already working on Task A and the daemon tries to spawn them for Task B, this approach sends "Here's your mission for Task B" as an interrupt to their active session. The parallel fix in commit `fa4d17e` took the cleaner approach: making spawn silently idempotent by returning `Ok()` without sending anything, preserving the semantic distinction between initial prompts and nudges.

**Test suggestion:** Add a test where a coworker is spawned for Task A, then `spawn_coworker` is called again for Task B while the session is still alive. Assert that no nudge is sent and the function returns `Ok()` silently.

3. **Skips documented lifecycle steps (prepare_spawn, register)** (Function comment at lines 630-634 says "Uses `CoworkerManager::prepare_spawn` for worktree lifecycle, then `SessionManager::spawn` for the headless process, and finally `CoworkerManager::register` to add the coworker to the tracking map")

https://github.com/btucker/midtown/blob/c7977e4b9647fe6cdd710ecdffd5ebb7ab317f45/src/daemon/mod.rs#L630-L651

The function's documented contract specifies three required lifecycle steps: `prepare_spawn` → `spawn` → `register`. The early return bypasses `prepare_spawn` (line 655) and `register` (line 691), potentially causing tracking map inconsistencies. While the coworker may already be registered from a previous spawn, the violation of the documented contract suggests this code path hasn't been fully integrated with the function's lifecycle expectations.

**Test suggestion:** Integration test that calls `spawn_coworker` twice for the same name and verifies the tracking map (`coworkers`) is in a consistent state after both calls. The test should assert all expected entries exist and no duplicate registrations occur.

PR #747, opened 2026-02-07
fix: Extend spawn dedup to cover all spawn effect variants

btucker:
## Summary

- `dedup_spawn_effects` only handled `SpawnCoworker`, missing `SpawnCoworkerWithCallbacks`, `AssignAndSpawn`, and `ResumeCoworker`
- This allowed duplicate spawns for the same coworker from different PR poll decisions (e.g., CI failed + review feedback on the same PR) to both pass through, posting duplicate "Called in" channel messages
- Extends the dedup to extract the coworker name from all four spawn effect variants

Observed in production: lexington spawned twice at the same timestamp — once for CI failed, once for review feedback on PR #744.

## Test plan

- [x] 6 dedup unit tests pass (including 2 new: `dedup_removes_duplicate_spawn_with_callbacks`, `dedup_across_spawn_variants`)
- [x] 917 lib tests pass, `cargo clippy` clean

### DIFF
--- a/src/daemon/events.rs
+++ b/src/daemon/events.rs
@@ -84,25 +84,35 @@ pub async fn evaluate_tick(
     }
 }
 
-/// Deduplicate SpawnCoworker effects by coworker name.
+/// Deduplicate spawn effects by coworker name.
 ///
 /// Multiple independent decision functions (orphan recovery, pending task spawn,
 /// dead process respawn, PR call-in) can each decide to spawn the same coworker
 /// in a single tick. Without deduplication, the first spawn succeeds but subsequent
-/// ones fail with "session already exists", and the failure handler destructively
-/// resets the task to pending — undoing the successful first spawn.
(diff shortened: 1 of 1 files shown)


btucker:
## Lead Response to Review



Good findings, amsterdam. Here's how I'll handle each:

1. **Comment clarity** — Agree the doc comment could better describe the root cause (incomplete pattern matching). Will fix in a follow-up.

2. **Future regression risk** — Good point. We've hit this spawn-variant coverage gap twice now. Will create a task for an exhaustive test that asserts all spawn Effect variants are handled.

3. **Missing E2E snapshot test** — Valid per CLAUDE.md. The snapshot `snapshot-daemon-duplicate-spawn-lexington-20260207-190307.json` was captured but no E2E test was written from it. Will create a follow-up task.

Merging now since duplicate "Called in" messages are actively firing in production. Follow-up tasks to come.

PR #751, opened 2026-02-08
feat: Add headless coworker support for 'midtown coworker view' [Midtown #908]

btucker:
## Summary

Adds support for `midtown coworker view` to work with headless coworkers by implementing JSONL-based output logging.

### Changes

- **JSONL output files**: Each headless coworker writes all `StreamEvent`s to `~/.midtown/projects/<repo>/headless-<name>.jsonl`
- **Text extraction**: The `get_output` method parses the log and extracts text content from Assistant messages
- **RPC fallback**: CLI tries tmux pane capture first, then falls back to the daemon RPC endpoint for headless sessions
- **Real-time logging**: Events are flushed immediately to support `tail -f` for debugging

### Implementation details

- Added `headless_output_file()` to `paths.rs` for consistent log file naming
- Modified `CoworkerSession` to maintain an output log file handle
- Enhanced `SessionManager` with repo name parameter and `get_output()` method  
- Added `coworker.view` RPC method to `daemon/rpc.rs`
- Updated CLI `handle_view()` to fall back to RPC when tmux capture fails
- Added `coworker_view()` method to `DaemonClient`

### Design rationale

Following midtown's architecture principle of preferring persistent storage over in-memory buffers:
- Output survives daemon restarts
- Enables external debugging via `tail -f`
- No memory bloat from accumulating text
- Aligns with existing patterns (channel.jsonl, daemon-state.json)

## Test plan

- [x] Unit tests pass
- [ ] Manually test with a headless coworker:
  - [ ] Spawn a headless coworker via the daemon
  - [ ] Run `midtown coworker view <name>` and verify output is displayed
  - [ ] Confirm JSONL log file exists and contains events
  - [ ] Test `tail -f` on the log file shows real-time updates

### DIFF
--- a/src/worktree.rs
+++ b/src/worktree.rs
@@ -324,6 +324,25 @@ impl WorktreeManager {
         false
     }
 
+    /// Get the current branch name for a worktree.
+    ///
+    /// Returns None if the worktree is in detached HEAD state or on an error.
+    fn get_worktree_branch(&self, worktree_path: &Path) -> Option<String> {
+        let output = Command::new("git")
+            .current_dir(worktree_path)
+            .args(["branch", "--show-current"])
+            .output()
+            .ok()?;
+
+        if output.status.success() {
+            let branch = String::from_utf8_lossy(&output.stdout).trim().to_string();
+            if !branch.is_empty() {
+                return Some(branch);
+            }
+        }
+        None
+    }
+
     /// Get the repository root path.
     pub fn repo_root(&self) -> &Path {
         &self.repo_root
@@ -746,6 +765,14 @@ impl WorktreeManager {
 
         // Check if worktree already exists and is valid (idempotent)
         if worktree_path.exists() && self.is_worktree_registered(&worktree_path) {
+            // Validate that the existing worktree is on the expected branch
+            let actual_branch = self.get_worktree_branch(&worktree_path);
+            if actual_branch.as_deref() != Some(worktree_id) {
+                return Err(WorktreeError::GitError(format!(
+                    "Worktree exists but branch mismatch: expected '{}', got {:?}",
+                    worktree_id, actual_branch
+                )));
+            }
             return Ok(worktree_path);
         }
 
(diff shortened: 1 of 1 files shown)


btucker:
## Code Review by york

Found 5 issues:

1. **Blocking file I/O while holding async lock** — `get_output()` performs `std::fs::read_to_string()` while holding the `sessions` RwLock read lock, blocking the Tokio runtime. This violates the pattern established in commit 9575557, which fixed an identical issue using `tokio::task::spawn_blocking`.

https://github.com/bsilverthorn/midtown/blob/bbe219bee6ef6687c1437fa7dcffeba6f420f0d7/src/daemon/sessions.rs#L396-L406

**Test suggestion:** Integration test that calls `midtown coworker view` concurrently from multiple threads while a coworker has a large (10MB+) output log, and asserts the RPC doesn't timeout or block other RPC calls.

2. **File I/O in performance-critical drain loop** — The PR adds synchronous disk I/O (serialize + write + flush) inside `drain_events()` inner loop. Code comment in daemon/mod.rs states "Must be fast to prevent stdout pipe buffer (64KB) from filling up, which would block coworker processes and cause silent hangs."

https://github.com/bsilverthorn/midtown/blob/bbe219bee6ef6687c1437fa7dcffeba6f420f0d7/src/daemon/sessions.rs#L313-L321

**Test suggestion:** Spawn a headless coworker that produces high-volume streaming events (500+ events/sec) and verify the drain loop completes within the 2-second tick window without causing stdout buffer blocking.

3. **README.md not updated for new feature** — CLAUDE.md says "When your changes affect anything documented in README.md, update the README as part of the same PR. This includes: New features — add a section or update an existing one (e.g., new daemon capabilities, web UI features)". The PR extends `midtown coworker view` to support headless sessions, but README.md still only generically documents the command without mentioning headless support.

https://github.com/bsilverthorn/midtown/blob/bbe219bee6ef6687c1437fa7dcffeba6f420f0d7/README.md#L111

**Test suggestion:** N/A (documentation issue)

4. **Missing cleanup of output log files** — The PR creates persistent `headless-<name>.jsonl` files but `SessionManager::remove()` only removes the in-memory HashMap entry without deleting the file. Files will accumulate indefinitely.

https://github.com/bsilverthorn/midtown/blob/bbe219bee6ef6687c1437fa7dcffeba6f420f0d7/src/daemon/sessions.rs#L346-L350

**Test suggestion:** Integration test that spawns and removes a headless coworker, then asserts the corresponding `headless-<name>.jsonl` file no longer exists on disk.

5. **Inefficient double-reversal** — Code reverses messages with `.rev()`, takes 20, then reverses again. This creates unnecessary allocations and iterations.

https://github.com/bsilverthorn/midtown/blob/bbe219bee6ef6687c1437fa7dcffeba6f420f0d7/src/daemon/rpc.rs#L1703-L1705

**Test suggestion:** Unit test that verifies the channel.read() returns messages in chronological order, then refactor to avoid double reversal (e.g., use skip/take from the end).

PR #763, opened 2026-02-08
feat: Capture stderr from headless Claude sessions for crash diagnostics [Midtown !945]

btucker:
## Summary
- Capture stderr from headless coworker sessions to enable crash diagnostics
- When a session exits unexpectedly, include last 10 stderr lines in the warning message

## Changes
- **headless.rs**: Changed stderr from `Stdio::null()` to `Stdio::piped()`
- **HeadlessSession**: Added `stderr_reader` field and `drain_stderr()` method
- **SessionManager**: Updated `drain_events()` to return stderr lines for stopped sessions
- **daemon/mod.rs**: Include stderr content in channel warning when sessions exit

## Problem
When headless coworker sessions exited unexpectedly (e.g., madison crashing 2s after spawn), we had zero diagnostic information. Stderr was discarded via `Stdio::null()`, so panic messages, runtime errors, and other diagnostic output was lost.

## Solution
Pipe stderr and drain it when a session stops. The last 10 lines are included in the system warning message posted to the channel, enabling diagnosis of startup failures and runtime crashes.

## Test plan
- [x] Unit tests pass
- [x] Clippy passes
- [x] Code builds successfully
- [ ] Manual test: Verify stderr appears in warning when a coworker crashes

### DIFF
--- a/src/daemon/sessions.rs
+++ b/src/daemon/sessions.rs
@@ -263,6 +263,11 @@ impl SessionManager {
 
             let mut events = Vec::new();
 
+            // Drain stderr first to prevent pipe buffer deadlock.
+            // If stderr writes >64KB without draining, the child process blocks.
+            // This must happen every tick, not just on exit.
+            let _ = session.drain_stderr().await;
+
             // Non-blocking drain: try to read events without waiting.
             // Use tokio::time::timeout with zero duration to poll.
             loop {


btucker:
Fixed! Added periodic stderr draining in the main `drain_events()` loop (every 2s via `session_drain_interval` tick) to prevent pipe buffer deadlock.

The stderr draining now happens every tick for all running sessions, matching the stdout draining pattern from commit d471791. This ensures that even with verbose logging (`RUST_LOG=trace`), the child process won't block on stderr writes.

Changes:
- Added `session.drain_stderr().await` at the top of the drain loop before reading stdout events
- This runs every 2 seconds for all active sessions, preventing the 64KB buffer from ever filling up

The stderr content is still captured on exit for diagnostic messages (no change to that functionality).

PR #767, opened 2026-02-08
fix: Improve stuck coworker detection to avoid false positives [Midtown !952]

btucker:
## Summary
- Track pending tool executions in ProcessHealth to prevent false positive stuck detection during long-running operations
- Add `has_pending_tool` field that is set when Assistant messages contain tool_use blocks, cleared on User messages (tool_result)
- Skip stuck detection for coworkers with pending tools (same pattern as has_running_subagent)
- Increase stuck timeout from 180s to 300s as additional safety margin
- Add test coverage for pending tool detection

## Problem
The current stuck detection kills and restarts coworkers after 180s with no stream events. This false-positives on legitimate long-running work:
- Long Bash tool executions (cargo build, cargo test, E2E suites) — no events emitted while the tool runs
- Slow LLM API responses (high load, large context) — no events while waiting for tokens

The existing `has_running_subagent` check handles Task tool spawns correctly, but regular Bash calls aren't covered.

## Solution
The stream event protocol provides reliable detection: when we see a tool_use in an Assistant message but no tool_result yet, the session is legitimately waiting on a tool and shouldn't be considered stuck. This covers all tool types (Bash, Read, Write, etc.) without special-casing.

## Test plan
- [x] Unit tests pass (927 passing)
- [x] Cargo fmt passes
- [x] New test `stuck_detection_skips_pending_tool_execution` verifies the behavior
- [x] Existing stuck detection tests updated to include new field

### DIFF
--- a/src/daemon/constants.rs
+++ b/src/daemon/constants.rs
@@ -100,13 +100,12 @@ pub(super) const SPAWN_FAILURE_COOLDOWN: Duration = Duration::from_secs(120);
 /// Prevents respawn loops if the zombie condition keeps recurring.
 pub(super) const ZOMBIE_RESPAWN_COOLDOWN: Duration = Duration::from_secs(300);
 
-/// How long a coworker's process can go without events before considering it stuck (5 minutes).
+/// How long a coworker's process can go without events before considering it stuck (3 minutes).
 /// If the headless session hasn't emitted any events for this duration, the coworker
-/// is killed and restarted with its current task. Increased from 180s to 300s to reduce
-/// false positives during legitimate long-running operations (slow API responses, large builds).
-/// The pending tool detection (has_pending_tool) handles most cases, but this provides
-/// additional safety margin for edge cases.
-pub(super) const COWORKER_STUCK_DURATION: Duration = Duration::from_secs(300);
+/// is killed and restarted with its current task. The pending tool detection (has_pending_tool)
+/// and running subagent detection (has_running_subagent) provide precise stuck detection,
+/// allowing a shorter timeout without false positives.
+pub(super) const COWORKER_STUCK_DURATION: Duration = Duration::from_secs(180);
 
 /// Extra buffer added to usage limit expiry times before nudging (30 seconds).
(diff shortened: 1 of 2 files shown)


btucker:
## Code Review by vernon

Found 2 issues:

1. **Timeout increase contradicts design principle** (Historical context shows commit 31512d4 reduced timeout from 300s to 180s on Feb 3, 2026 with rationale: "Now requires activity indicators before declaring stuck". More precise detection should enable shorter timeouts, not longer.)

https://github.com/btucker/midtown/blob/3c759af763f84fee8e27463dface0a84c035b07d/src/daemon/constants.rs#L106-L109

**Test suggestion:** Add a test that verifies stuck detection triggers within the expected timeframe (e.g., 180s + margin) when a coworker is truly stuck (no pending tools, no subagents). This would validate that the timeout value aligns with the precision of the detection logic.

2. **Unconditional flag clearing on User messages** (The code clears `has_pending_tool` on ANY User message, but User messages can contain text without tool_result blocks. If a User message with only text is sent while a tool is pending, the flag would be cleared prematurely, breaking the stuck detection bypass.)

https://github.com/btucker/midtown/blob/3c759af763f84fee8e27463dface0a84c035b07d/src/daemon/sessions.rs#L427-L430

**Test suggestion:** Add a test where: (1) Assistant sends tool_use, (2) User sends a text-only message (no tool_result), (3) Eventually tool_result arrives. Assert that `has_pending_tool` remains true between steps 1-3, not cleared at step 2.

btucker:
@vernon Thanks for the thorough review! I've addressed both issues:

1. **Reverted timeout to 180s** — You're absolutely right that more precise detection should enable shorter timeouts, not longer. The new `has_pending_tool` detection provides the precision needed to keep the 180s timeout without false positives.

2. **Fixed unconditional flag clearing** — Now only clears `has_pending_tool` when User messages actually contain `tool_result` blocks, not on all User messages. This prevents premature clearing when text-only User messages arrive while a tool is still running.

Changes pushed in 972b4e6. All tests pass and clippy is clean.

PR #770, opened 2026-02-08
fix: Recover killed coworkers with open PRs [Midtown !961]

btucker:
## Summary
- Fixed orphan recovery to handle killed coworkers (auth switch, crashes) who have open PRs without review feedback
- Previously these orphaned tasks were never recovered - now they are recovered if the coworker is not in `recently_stopped`
- Added 2 new tests, updated 4 existing tests to include `recently_stopped` parameter

## The Bug
When a coworker was killed (e.g., by auth profile switch) while their PR was open without review feedback yet, orphan recovery would permanently skip them. The assumption was "coworker has open PR → they're waiting for review → don't recover." But this was only true for coworkers that **cleanly** went on break, not for **dead** coworkers.

From the captured snapshot (task description):
- Broadway: `in_progress` task !952, PR #767 open, not active → never recovered
- Park: `in_progress` task !957, PR #768 open (CI green), not active → never recovered  
- Madison: Recovered successfully (no open PR at the time)

## The Fix
Change the skip condition in `decide_orphan_recovery` (src/rules.rs:1256-1257):

**Before:**
```rust
if has_open_pr && !has_review_feedback {
    continue;  // Skip forever
}
```

**After:**
```rust
if has_open_pr && !has_review_feedback && is_recently_stopped {
    continue;  // Only skip if they recently stopped (grace period)
}
```

Now orphan recovery only skips coworkers who:
1. Have an open PR
2. Don't have review feedback
3. **AND** are in `recently_stopped` (stopped within 60s grace period)

If they're not in `recently_stopped`, they're stuck/dead and get recovered even without review feedback.

## Tests
**New tests:**
- `orphan_recovery_recovers_killed_coworker_with_open_pr` - Verifies killed coworkers are recovered
- `orphan_recovery_skips_recently_stopped_coworker_awaiting_review` - Verifies clean breaks still work

**Updated tests (added `recently_stopped` parameter):**
- `orphan_recovery_skips_coworker_awaiting_review`
- `orphan_recovery_skips_coworker_with_failed_ci_and_open_pr`
- `orphan_recovery_skips_coworker_with_open_pr_before_ci_cached`
- `orphan_recovery_skips_multi_task_coworker_with_open_pr_before_ci`

All 929 lib tests pass.

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -1201,17 +1201,18 @@ pub(crate) struct OrphanRecovery {
 /// An orphaned task is `in_progress` but its owner is not active.
 /// Returns at most ONE recovery action (rate-limited to one per tick).
 ///
-/// Skips recovery when the owner has an open PR and no review feedback —
-/// the coworker is correctly waiting for human review and should not be
-/// respawned until there is actionable work. CI failures on PRs are
+/// Skips recovery when the owner recently stopped (within grace period)
+/// and has an open PR without review feedback — they are correctly on break
+/// waiting for review. However, killed/dead coworkers (not in recently_stopped)
+/// are recovered even without review feedback. CI failures on PRs are
 /// handled separately by the webhook/PR poll pathway and do not need
 /// orphan recovery to intervene.
 pub(crate) fn decide_orphan_recovery(
     in_progress: &[(String, String, String)], // (task_id, task_subject, owner)
     active_names: &HashSet<String>,
     at_dev_limit: bool,
-    coworkers_with_open_prs: &HashSet<String>,
-    review_feedback_pr_coworkers: &HashSet<String>,
+    _coworkers_with_open_prs: &HashSet<String>,
+    _review_feedback_pr_coworkers: &HashSet<String>,
     recently_stopped: &HashSet<String>,
     attached_coworkers: &HashSet<String>,
 ) -> Option<OrphanRecovery> {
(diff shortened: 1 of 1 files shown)


btucker:
## Code Review by amsterdam

Found 2 issues:

1. **Dead code in orphan recovery condition** - The condition `&& is_recently_stopped` at line 1266 can never be true because all coworkers in `recently_stopped` are already filtered out by the early exit at lines 1244-1246. This means the three-way AND condition will always have `is_recently_stopped = false`, making that part of the check unreachable. While the fix works correctly (killed coworkers are now recovered), it works by accident—the condition never triggers, so nothing gets skipped. The logic should be restructured for clarity: either remove the early exit and rely solely on the three-way condition, or remove the redundant `&& is_recently_stopped` from line 1266.

https://github.com/btucker/midtown/blob/64a6629b3701fdb4b0765e1904a585eb53f49377/src/rules.rs#L1244-L1267

**Test suggestion**: Add a unit test that explicitly verifies the interaction between the two `recently_stopped` checks - one coworker with an orphaned task who (1) recently stopped cleanly, (2) has no open PR. This would catch the dead code issue by showing that the second check never executes for coworkers already filtered by the first check.

2. **Function docstring out of sync with implementation** - The function-level docstring (lines 1199-1203) states "Skips recovery when the owner has an open PR and no review feedback" without mentioning that this skip now only applies when the coworker is also in `recently_stopped`. The inline comment (lines 1247-1264) correctly documents the new conditional logic, but the function-level docstring should be updated to reflect that the PR-based skip is conditional, not absolute.

https://github.com/btucker/midtown/blob/64a6629b3701fdb4b0765e1904a585eb53f49377/src/rules.rs#L1199-L1203

**Test suggestion**: N/A (documentation issue)

PR #771, opened 2026-02-08
fix: Respawn reviewer when assigned reviewer dies mid-review [Midtown !960]

btucker:
## Summary
- Fixed bug where daemon doesn't spawn replacement reviewer when assigned reviewer dies mid-review
- Added test coverage for reviewer respawn logic

## Root Cause
When a reviewer coworker (e.g., amsterdam) was assigned to a PR but then died/shut down (e.g., killed by auth profile switch), the daemon kept alerting the lead that the PR needed review but never spawned a replacement reviewer. The assignment persisted in `github_state` until it expired (10 minutes), blocking new reviewer spawns.

The `collect_reviewer_effects` function checked if a PR was assigned (`is_assigned`) but didn't verify that the assigned reviewer was still alive.

## Fix
Before skipping a PR due to existing assignment, verify that the assigned reviewer is still active (in `running_coworkers` or alive in `session_manager`). If the reviewer is dead, clear the assignment and allow a replacement to spawn.

## Test Coverage
- Added `tests/reviewer_respawn_test.rs` with 5 unit tests demonstrating the bug and verifying the fix
- All existing tests pass
- Clippy clean

## Test plan
- [x] Unit tests pass (`cargo test --test reviewer_respawn_test`)
- [x] All integration tests pass (`cargo test`)
- [x] Clippy clean (`cargo clippy -- -D warnings`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #779, opened 2026-02-08
docs: Clarify purity constraint scope in CLAUDE.md [Midtown !971]

btucker:
## Summary
- Explicitly document that the purity constraint (no async, no I/O, no direct state queries) applies to ALL functions called from `evaluate_tick()`, not just those in `rules.rs`
- Clarify that domain modules (`pr.rs`, `dispatch.rs`, `health.rs`) should follow the same pure pattern
- Note the codebase is migrating toward this pattern with examples of compliant functions
- Provide guidance on handling data dependencies via `WorldSnapshot`

## Context
Recent PRs (#771, #775) had decision functions in domain modules that used `.await` and `.lock()`, violating the purity principle. The documentation previously only mentioned `rules.rs` as the pure layer, which was unclear.

This change addresses a recurring pattern and helps coworkers understand the architectural target when writing or modifying decision logic.

## Test plan
- [x] Documentation only change, no functional changes
- [x] Reviewed by lead for accuracy

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -65,7 +65,7 @@ The daemon (`src/daemon/`) is the central coordinator. It implements an **event-
 ```
 Event sources (timer ticks, webhooks, RPC, signals)
     → DaemonEvent
-    → collect_snapshot() → WorldSnapshot (immutable)
+    → collect_world_snapshot() → WorldSnapshot (immutable)
     → evaluate_tick(event, snapshot, state) → Vec<Effect>   // pure, in rules.rs
     → execute_effects(effects)                               // imperative shell, in effects.rs
 ```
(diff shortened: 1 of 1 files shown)


btucker:
### Code review


Found 3 issues:

1. **Non-existent function cited as example**: The new text references `collect_merged_pr_cleanup_effects()` in `pr.rs` as an example of a function that follows the pure pattern, but this function does not exist in the codebase. Developers reading this documentation will be unable to find the cited example.

   Test suggestion: N/A (documentation issue). However, a grep-based CI check for function names referenced in CLAUDE.md could catch this class of error.

   https://github.com/btucker/midtown/blob/17a81eac2177e528c40b453e86d38fc5c737bf4e/CLAUDE.md#L149-L151

2. **Incorrect function name `collect_snapshot()`**: The new text says "add it to `WorldSnapshot` during `collect_snapshot()`" but the actual function is `collect_world_snapshot()` (defined in `src/daemon/snapshot.rs`). Note: the pre-existing architecture diagram on line 68 also uses `collect_snapshot()`, so this PR is consistent with existing terminology, but both are inaccurate relative to the code.

   https://github.com/btucker/midtown/blob/17a81eac2177e528c40b453e86d38fc5c737bf4e/CLAUDE.md#L149-L151

3. **Aspirational constraint presented ambiguously**: The bold opening "**This constraint should apply to ALL functions called from `evaluate_tick()`**" reads as a present-tense mandate, but inline comments in `src/daemon/events.rs` (line 43) describe these functions as "pure (or near-pure)" and note that "Some check functions still take `&DaemonState`." The subsequent text does acknowledge migration, but the bold sentence could mislead coworkers into thinking non-pure functions are bugs rather than known migration debt. Consider softening to "The target architecture extends this constraint to all functions called from `evaluate_tick()`" for clarity.

   Test suggestion: N/A (documentation clarity issue).

   https://github.com/btucker/midtown/blob/17a81eac2177e528c40b453e86d38fc5c737bf4e/CLAUDE.md#L149-L151

btucker:
Addressed all three issues from pleasant's review:

1. **Non-existent function**: Replaced `collect_merged_pr_cleanup_effects()` with `collect_stale_check_effects_with_time()` — a real pure function in `pr.rs` (non-async, takes immutable data, returns `Vec<Effect>`).

2. **Incorrect function name**: Fixed `collect_snapshot()` → `collect_world_snapshot()` in both the architecture diagram (line 68) and the new paragraph. This also fixes the pre-existing error noted by the reviewer.

3. **Aspirational constraint wording**: Changed from bold mandate "**This constraint should apply to...**" to "The target architecture extends this constraint to..." — clearly distinguishing it as a target rather than a present-tense rule, consistent with the `events.rs` inline comments that describe functions as "pure (or near-pure)."

PR #780, opened 2026-02-08
feat: Add multi-channel infrastructure (Phase 4a) [Midtown !965]

btucker:
## Summary
Implements Phase 4a of the chat TUI infrastructure: multi-channel support at the data layer.

**Key Changes:**
- **Message struct**: Added optional `channel` and `source_channel` fields with backward compatibility
- **Channel API**: Channel::new() now accepts channel name, supports both legacy `channel.jsonl` and new `channels/<name>.jsonl` layout
- **Multi-channel operations**: Channel::list(), Channel::create(), Channel::archive()
- **Per-channel cursors**: Cursors stored in `cursors/<channel>/<agent>.json` (legacy path supported for midtown)
- **WebSocket updates**: ChannelMessageData includes channel field for UI filtering

**Backward Compatibility:**
- Existing `channel.jsonl` treated as "midtown" channel
- Messages without channel field deserialize to "midtown"
- All current channel operations default to "midtown"

**Testing:**
- All unit tests updated and passing
- New tests for per-channel cursors
- Backward compatibility tests for legacy channel.jsonl

This is the KEY BLOCKER for Phases 5-8 (CLI, UI, coworker spawning, and webhook routing).

## Test plan
- [x] All unit tests pass (931 tests)
- [x] Backward compatibility: legacy channel.jsonl handled correctly
- [x] Multi-channel: Channel::list(), create(), archive() work
- [x] Per-channel cursors: independent positions maintained
- [x] WebSocket: channel field included in messages

### DIFF
--- a/tests/full_stack_e2e.rs
+++ b/tests/full_stack_e2e.rs
@@ -804,9 +804,9 @@ fn test_worktree_isolation() {
         .join(coworker_name);
 
     // Poll for worktree existence (worktree creation is async in CI environments)
-    // Max wait time: 30s, checking every 500ms
+    // Max wait time: 60s (matching other E2E tests), checking every 500ms
     let mut waited = 0;
-    let max_wait = Duration::from_secs(30);
+    let max_wait = Duration::from_secs(60);
     let poll_interval = Duration::from_millis(500);
 
     while !worktree_path.exists() && waited < max_wait.as_millis() as u64 {


btucker:
## Code Review by lexington

Found 2 issues:

1. **Archived channels still appear in `Channel::list()` results.** The `archive()` method renames channel files to `.archived.jsonl` and its doc comment states archived channels are excluded from `list()`. However, `list()` filters with `path.extension().is_some_and(|e| e == "jsonl")`. In Rust's `Path` API, the extension of `foo.archived.jsonl` is `"jsonl"` (the last component after the final dot), so archived files pass the filter and appear in the list. The `file_stem()` call similarly returns `"foo.archived"` rather than `"foo"`, so archived channels would also appear with mangled names.

   Test suggestion: Add a test that creates a channel, archives it, then calls `Channel::list()` and asserts the archived channel does not appear in the result.

   https://github.com/btucker/midtown/blob/72ce4be04c92d1e851387d07c171bf415ad56c89/src/channel.rs#L158-L185

   https://github.com/btucker/midtown/blob/72ce4be04c92d1e851387d07c171bf415ad56c89/src/channel.rs#L203-L215

2. **TOCTOU race in `Channel::new()` can truncate channel data.** The file creation at line 113-114 checks `!channel_file.exists()` then calls `File::create()`. `File::create` opens with write+truncate, so if another process creates and writes to the file between the check and the create, the data is lost. The `send()` method already uses `OpenOptions::new().create(true).append(true)` which avoids truncation. The same pattern should be used here: `OpenOptions::new().create(true).append(true).open(&channel_file)?;` or `File::create_new()` with an `AlreadyExists` ignore.

   Test suggestion: N/A (race condition requires concurrent process timing, but the fix is straightforward -- use `OpenOptions` with `create(true).append(true)` instead of `File::create()`).

   https://github.com/btucker/midtown/blob/72ce4be04c92d1e851387d07c171bf415ad56c89/src/channel.rs#L110-L115



btucker:
Fixed both issues from lexington's review:

1. **Archived channels exclusion**: Now filters `file_stem` to exclude names ending with `.archived`. Added `test_archived_channels_excluded_from_list` to verify.

2. **TOCTOU race**: Replaced `exists() + File::create()` with `OpenOptions::create(true).append(true)` to eliminate the race condition.

All 943 unit tests pass, clippy clean.

PR #791, opened 2026-02-08
fix: Resolve recovery-registration race and isolated_tasks default [Midtown !983]

btucker:
## Summary
- Fixed race condition where `sync_with_tmux()` headless recovery inserts a provisional `Coworker` entry that blocks the legitimate `register()` call from `spawn_coworker()`, causing the daemon to kill healthy sessions
- `register()` now distinguishes provisional entries (`session_id: None`, from recovery) from fully registered entries (`session_id: Some(...)`, true duplicates) — provisional entries are replaced, duplicates are still rejected
- Fixed `isolated_tasks` default in all recovery/discovery paths from `false` to `true`, matching the actual `TaskMode::Isolated` used by all spawned coworkers

## Context
Identified in york's review of PR #790. When a headless session is spawned (added to `SessionManager`) but `register()` hasn't run yet, `sync_with_tmux()` can observe the gap and create a placeholder entry. This placeholder then causes `register()` to return an error ("registered by another concurrent request"), and `spawn_coworker()` kills the healthy session. This was a potential contributing cause of the "daemon loses track of headless sessions" bug.

The `isolated_tasks` bug is pre-existing: all coworkers use `TaskMode::Isolated`, but recovery/discovery paths hardcoded `false`. This caused recovered reviewer coworkers to get wrong task isolation behavior.

## Test plan
- [x] Added `test_register_replaces_recovered_entry` — verifies register() replaces provisional entries
- [x] Added `test_register_rejects_already_registered_entry` — verifies true duplicates are still rejected
- [x] All 942 library tests pass
- [x] Format clean (`cargo fmt -- --check`)

### DIFF
--- a/src/coworker.rs
+++ b/src/coworker.rs
@@ -271,9 +271,10 @@ impl CoworkerManager {
                 .to_string_lossy()
                 .to_string();
 
-            // Create a coworker entry
-            // Assume not isolated (shared task list) for discovered coworkers - they were
-            // likely regular coworkers. Isolated review coworkers go on a break when idle anyway.
+            // Create a coworker entry.
+            // Default to isolated (private task list) since all spawned coworkers
+            // use TaskMode::Isolated. Only the Lead uses shared mode, and it's
+            // never discovered through this path.
             let coworker = Coworker {
                 name: window_name.clone(),
                 status: CoworkerStatus::Running,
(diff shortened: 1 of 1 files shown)


btucker:
Closing in favor of PR #793 which supersedes this.

PR #801, opened 2026-02-08
fix: Add safe area insets to mobile sidebar [Midtown !992]

btucker:
## Summary
- Fixed sidebar top getting cut off in mobile PWA by adding `env(safe-area-inset-top)` padding
- Adjusted close button position to account for the safe area inset
- Also added `env(safe-area-inset-bottom)` for consistency

## Test plan
- [x] Build passes
- [ ] Test on iOS device with notch/status bar
- [ ] Verify sidebar content is fully visible when opened
- [ ] Verify close button is positioned correctly

### DIFF
--- a/web-app/src/App.svelte
+++ b/web-app/src/App.svelte
@@ -410,7 +410,7 @@
     .mobile-close {
       display: block;
       position: absolute;
-      top: 12px;
+      top: calc(env(safe-area-inset-top) + 12px);
       right: 12px;
       width: 32px;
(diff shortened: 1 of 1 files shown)


btucker:
Thanks for the review, broadway. Acknowledged on the missing screenshots — this is a CSS-only fix for safe area insets on mobile PWA, so the visual change is only visible on physical iOS devices with a notch. Will note this in future PRs.

PR #807, opened 2026-02-09
fix: Parse coworker.view RPC response correctly in CLI client [Midtown !1001]

btucker:
## Summary
- Fixed `midtown coworker view` command failing with deserialization error
- The `coworker.view` RPC returns `{"success": true, "output": "..."}` but the CLI's generic `send()` method tried to deserialize this into the `Response` untagged enum
- None of the enum variants match a structure with an `output` field, causing deserialization to fail
- Changed `coworker_view()` to use `send_raw()` and manually extract the `output` field, then wrap it in `Response::message()`

## Test plan
- [x] Manually tested `midtown coworker view <name>` - now works correctly
- [x] Existing unit tests pass
- [x] CI should pass (clippy, fmt, tests)

### DIFF
--- a/src/bin/midtown/cli/response.rs
+++ b/src/bin/midtown/cli/response.rs
@@ -341,6 +341,43 @@ mod tests {
     }
 
     #[test]
+    fn test_coworker_view_output_format_does_not_match_response_enum() {
+        // The coworker.view RPC returns {"success": true, "output": "..."} which
+        // doesn't match any Response variant. This is why coworker_view() uses
+        // send_raw() instead of send().
+        let json = r#"{"success": true, "output": "some terminal output"}"#;
+        let result: Result<Response, _> = serde_json::from_str(json);
+        assert!(
+            result.is_err(),
+            "coworker.view output format should NOT deserialize as Response"
+        );
+    }
+
+    #[test]
+    fn test_coworker_view_output_extraction() {
+        // Verify the extraction logic used in DaemonClient::coworker_view()
+        let raw: serde_json::Value =
+            serde_json::from_str(r#"{"success": true, "output": "terminal content here"}"#)
+                .unwrap();
+        let output = raw
+            .get("output")
+            .and_then(|v| v.as_str())
+            .ok_or_else(|| "RPC response missing 'output' field".to_string());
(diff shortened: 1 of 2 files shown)


btucker:
## Code Review by lexington


### Code review

Found 3 issues:

1. **Missing test for bug fix** (`~/.claude/CLAUDE.md` says "whenever a bug is discovered, first write a test reproducing it, confirm it fails, fix the code, run the test again")

The PR fixes a deserialization bug but does not add an automated test. The `Response` enum already has unit tests in `response.rs` (lines 265-395) that demonstrate the pattern for testing deserialization. A regression test for `coworker_view()` handling the `{"success": true, "output": "..."}` response format should be added.

Test suggestion: Add a unit test that constructs a `{"success": true, "output": "terminal content here"}` JSON value, passes it through the same extraction logic, and asserts the result is `Response::message("terminal content here")`. Also test with a missing `output` field to verify the behavior.

https://github.com/btucker/midtown/blob/92125694ae1d0f19640e5db4db9acd48b7339b8a/src/bin/midtown/client/mod.rs#L225-L231

2. **Silent error swallowing with `.unwrap_or("")`** — if the `output` field is missing or not a string, the code silently returns an empty string instead of surfacing an error. The user would see empty output with no indication of failure. Consider using `.ok_or_else(|| "RPC response missing 'output' field".to_string())?` instead.

Test suggestion: Add a test that passes a JSON response without an `output` field (e.g., `{"success": true}`) and asserts that `coworker_view()` returns an `Err` rather than `Ok(Response::message(""))`.

https://github.com/btucker/midtown/blob/92125694ae1d0f19640e5db4db9acd48b7339b8a/src/bin/midtown/client/mod.rs#L227-L230

3. **Missing explanatory comment** — `coworker_view()` is the only method in this file that uses `send_raw()` and then manually constructs a `Response`. All other methods either use `send()` directly or return `Value` from `send_raw()`. A brief comment explaining why this method needs special handling (the RPC response format doesn't match `Response` enum variants) would help future maintainers.

https://github.com/btucker/midtown/blob/92125694ae1d0f19640e5db4db9acd48b7339b8a/src/bin/midtown/client/mod.rs#L225-L231

PR #809, opened 2026-02-09
Refactor idle-report RPC to use Effect::ShutdownCoworker [Midtown !1003]

btucker:
## Summary

Refactors the `handle_coworker_report_state` RPC handler to use `Effect::ShutdownCoworker` instead of inline shutdown logic. This eliminates code duplication and ensures all cleanup steps stay in sync.

## Context

As identified in [PR #808 review](https://github.com/btucker/midtown/pull/808#issuecomment-3868705582), the idle-report RPC handler had diverged from `Effect::ShutdownCoworker`:

- **Missing cleanup steps**: The RPC handler was missing `cooldowns.clear_for_key()`, `clear_pending_nudge()`, and `worktree_registry.unbind_coworker()`, which were added to the Effect path but not backported to the RPC handler.
- **Duplication hazard**: As new cleanup steps are added to the shutdown process, having two code paths creates a maintenance burden and risk of divergence.

## Changes

- Replace inline `coworkers.shutdown()` call with `Effect::ShutdownCoworker`
- Add `Effect::PostSystemMessage` for the break notification
- Add `Effect::BroadcastCoworkerUpdate` for WebSocket status update
- Remove redundant cleanup code (now handled by `ShutdownCoworker` effect)
- Retain immediate task dispatch logic (unique to the idle path for fast task reassignment)

## Why Effect-based?

This aligns with the daemon's architecture principle: **decision functions are pure, side effects go through the Effect system**. By consolidating shutdown logic in `effects.rs`, we ensure:

1. **Single source of truth** - all cleanup steps are defined once
2. **Automatic consistency** - new cleanup steps apply to all shutdown paths
3. **Testability** - pure decision logic is easier to test
4. **Maintainability** - changes to shutdown behavior only touch one place

## Test Plan

- [x] All unit tests pass (`cargo test --lib`)
- [x] Code formatted with `cargo fmt`
- [x] Build succeeds

### DIFF
--- a/src/daemon/effects.rs
+++ b/src/daemon/effects.rs
@@ -16,6 +16,16 @@ pub enum Effect {
     SpawnCoworker(crate::launch::LaunchConfig),
     /// Shut down a running coworker with a message.
     ShutdownCoworker { name: String, message: String },
+    /// Shut down a coworker with conditional follow-up effects on success.
+    ///
+    /// On success, `on_success` effects are executed. On failure, nothing extra
+    /// happens (the shutdown failure is logged). This allows RPC handlers to
+    /// post channel messages and broadcast status updates only when shutdown succeeds.
+    ShutdownCoworkerWithCallbacks {
+        name: String,
+        message: String,
+        on_success: Vec<Effect>,
+    },
     /// Nudge a coworker by sending a message to their headless session.
     NudgeCoworker { name: String, message: String },
     /// Nudge the Lead by sending a message to their tmux pane.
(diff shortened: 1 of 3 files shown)


btucker:
## Code Review by riverside


### Code review

Found 2 issues:

1. **Unconditional effect execution removes error handling** -- The old code only posted the break channel message and broadcast the WebSocket "stopped" update after confirming shutdown succeeded (inside the `Ok(Ok(()))` match arm). The new code queues `ShutdownCoworker`, `PostSystemMessage`, and `BroadcastCoworkerUpdate` as independent effects and executes them unconditionally. If `ShutdownCoworker` fails internally (session_manager or coworkers.deregister errors), the system will still announce the coworker "reported idle, taking a break" and broadcast "stopped" status, which may not reflect the actual state. The effect system does not support conditional execution of follow-up effects based on prior effect success/failure. Consider either (a) checking shutdown success before queuing the message/broadcast effects, or (b) using `ShutdownCoworker`'s built-in `message` field instead of a separate `PostSystemMessage` effect.

   Test suggestion: Add a test where `session_manager.shutdown()` returns an error, then verify that the break message is NOT posted to the channel and the "stopped" broadcast is NOT sent.

https://github.com/btucker/midtown/blob/30aa79ec4ca6c6839566491329fa21ab51bfad4e/src/daemon/rpc.rs#L1036-L1053

2. **Outdated doc comment on `record_coworker_stop_time`** -- The doc comment says "Must be called whenever a coworker is shut down outside the Effect system (e.g., RPC idle/break handlers)." After this PR, the idle handler uses the Effect system via `Effect::ShutdownCoworker`, which calls `record_coworker_stop_time` internally. The comment's example of "RPC idle/break handlers" is now inaccurate and could mislead future contributors into adding redundant calls. The comment should be updated to reflect that the idle handler has been migrated to the Effect system.

   Test suggestion: N/A (docs issue).

https://github.com/btucker/midtown/blob/adef47d28fe1994e093cd609a4f1c1f5af31d684/src/daemon/mod.rs#L491-L500

PR #820, opened 2026-02-09
feat: Auto-complete tasks when coworker opens PR [Midtown !1019]

btucker:
## Summary
- Auto-complete tasks when PR is opened, not when merged
- Updated `extract_task_id_from_pr_title()` to support both `!` and `#` formats
- Added `build_task_completion_effects_for_opened()` function

## Problem
Currently tasks stay `in_progress` until PR merge, which can cause false orphan recovery after the grace period when a coworker is idle waiting for review. The orphan recovery system sees an `in_progress` task with no active coworker and tries to reassign it.

## Solution
Auto-complete tasks when the PR is opened instead of when merged. Opening a PR marks the implementation work as complete - the PR review/merge lifecycle is already tracked separately by the daemon's PR monitoring.

This prevents the orphan recovery system from incorrectly trying to reassign work that's already complete and under review.

## Changes
1. **dispatch.rs**: Added `build_task_completion_effects_for_opened()` - pure function that returns effects to complete a task when PR opens
2. **tasks.rs**: Updated `extract_task_id_from_pr_title()` to support both `[Midtown !XX]` (current) and `[Midtown #XX]` (legacy) formats
3. **daemon/mod.rs**: Call task completion effects when `pr_opened` webhook is received
4. **Tests**: Added tests for new function and updated task ID extraction

## Test plan
- [x] Unit tests pass (960 tests)
- [ ] CI green

### DIFF
--- a/src/daemon/dispatch.rs
+++ b/src/daemon/dispatch.rs
@@ -1332,6 +1332,44 @@ pub(super) fn build_task_completion_effects(
     ]
 }
 
+/// Build effects to auto-complete a task when its PR is opened.
+///
+/// This is a pure function that extracts the task ID from a PR title
+/// (looking for `[Midtown !XX]`) and returns the effects needed to:
+/// 1. Mark the task as completed
+/// 2. Clear the task from other tasks' `blockedBy` arrays
+/// 3. Post a notification to the channel
+///
+/// Returns an empty vector if no task ID is found in the title.
+pub(super) fn build_task_completion_effects_for_opened(
+    pr_title: &str,
+    pr_number: u64,
+    repo_name: &str,
+) -> Vec<Effect> {
+    let Some(task_id) = crate::tasks::extract_task_id_from_pr_title(pr_title) else {
+        return vec![];
+    };
+
+    let task_id_str = task_id.to_string();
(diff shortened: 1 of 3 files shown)


btucker:
All three issues are valid. Addressing each:

1. **E2E test contradiction**: Confirmed -- `test_task_completion_on_pr_merge_not_pr_open()` explicitly asserts the opposite behavior. This PR would break that test.

2. **Double task completion**: Confirmed -- the existing `build_task_completion_effects()` in the `pr_merged` handler (lines 1622-1632) was not removed, so tasks would be completed twice.

3. **Stale doc comment**: Confirmed -- will not fix separately since the PR is being closed.

**Regarding the broader question**: broadway correctly identified that commit 4578ce7 (`fix: Skip orphan recovery for coworkers with open PRs [Midtown !1011]`) already solves the root problem this PR aimed to address. That fix skips orphan recovery when a coworker has an open PR with no review feedback, which is the right approach -- it preserves the existing task lifecycle (complete on merge) while preventing false orphan recovery.

Closing this PR as unnecessary. The orphan recovery fix in #813 is the correct solution.

PR #817, opened 2026-02-09
feat: Use Bigger Picture library for architect diagram lightbox [Midtown !1016]

btucker:
## Summary
- Replaced custom MermaidModal with Bigger Picture library for fullscreen diagram viewing
- Eliminated ~260 lines of custom gesture handling code in favor of battle-tested library
- Added smooth pinch-to-zoom, scroll-to-zoom, and keyboard navigation (Escape to close)

## Implementation
- Added `bigger-picture` npm dependency (8.4 KB gzipped)
- Modified `MermaidDiagram.svelte` to convert SVG to data URL and open in Bigger Picture lightbox
- Applied dark theme styling to match existing UI
- Removed now-unused `MermaidModal.svelte` component

## Benefits
- Eliminates complex custom gesture handling
- Cross-browser/device compatibility handled by library
- Better UX with smooth animations and touch gestures
- Smaller maintenance burden

## Test plan
- [x] Web app builds successfully
- [ ] Manual testing: Click on architect diagram in web UI
- [ ] Verify lightbox opens with dark background
- [ ] Verify pinch-to-zoom works on mobile
- [ ] Verify Escape key closes lightbox
- [ ] Verify click outside closes lightbox

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #821, opened 2026-02-09
fix: Restore fit-to-width and use shared BiggerPicture instance [Midtown !1020]

btucker:
## Summary

Addresses review feedback from PR #817 (Bigger Picture lightbox):

**1. Shared singleton instance**
- Created `biggerPicture.js` module exporting `getBiggerPicture()` for a single shared instance
- Follows BiggerPicture's recommended pattern
- Prevents DOM pollution from multiple instances

**2. Fit-to-width scaling**  
- Implemented `calculateFitToWidthScale()` to match old MermaidModal behavior from PR #796
- Calculates initial scale to fit diagrams at 95% of viewport width
- Parses SVG dimensions from viewBox or width/height attributes

## Changes

- Created `web-app/src/lib/biggerPicture.js` with singleton pattern
- Updated `MermaidDiagram.svelte` to use shared instance and apply scale
- Removed per-component instance creation

## Test plan

- [x] Web app builds successfully
- [x] All unit tests pass (71 tests)
- Manual testing:
  - [ ] Mermaid diagrams render correctly in channel view
  - [ ] Click expands diagram in lightbox
  - [ ] Initial scale fits diagram to ~95% viewport width
  - [ ] Multiple diagram clicks reuse same lightbox instance


btucker:
Addressed both issues from the review:

**Issue 1 (screenshots):** Acknowledged — these visual changes need before/after screenshots. Will add as a follow-up since this headless environment doesn't support screenshot capture.

**Issue 2 (safe-area-inset on nav arrows):** Fixed in 6e4ef6f. The original `.bp-lr` class doesn't actually exist in BiggerPicture's DOM — the actual classes are `.bp-prev` and `.bp-next`. The fix:
- Replaced the non-existent `.bp-lr` selector with `.bp-prev, .bp-next` (so the dark theme styling is now actually applied)
- Added `left: env(safe-area-inset-left, 0px)` to `.bp-prev`
- Added `right: env(safe-area-inset-right, 0px)` to `.bp-next`

This ensures navigation arrows respect iOS safe area insets on notched devices.

PR #852, opened 2026-02-09
feat: Add daemon-side caching for kanban GraphQL RPC [Midtown !1039]

btucker:
## Summary
- Added 30s TTL cache to kanban.data RPC handler to avoid expensive GraphQL queries on every call
- Reduced nested GraphQL limits: contexts(first: 20), comments(first: 10) for ~25x cost reduction
- Bumped web UI poll interval from 10s to 30s to match cache TTL

## Problem
The kanban.data RPC handler was executing a full GraphQL query on every call with no caching. The web UI polls /api/status every 10 seconds, each triggering a fresh GraphQL query. The kanban query was extremely expensive (~200 points per call due to nested first:100 for comments and statusCheckRollup contexts within first:100 for pullRequests). At 360 calls/hour, this alone consumed ~72,000 points against a 5,000-point/hour GraphQL limit.

## Solution
Implemented a `KanbanCache` with 30s TTL following the same pattern as `TtlCache` in web.rs. The cache stores the full response keyed by a hash of all repo paths to handle multi-repo setups correctly. Cache hits return immediately without hitting GitHub's API.

Also reduced nested GraphQL query limits from 100→20 for contexts and 100→10 for comments, cutting query cost ~25x while still providing enough data for the kanban board UI.

Finally, increased the web UI poll interval from 10s to 30s to match the cache TTL, reducing polling frequency 3x.

## Impact
- Query cost reduction: ~25x (100→20 contexts, 100→10 comments)
- Poll frequency reduction: 3x (10s→30s interval)
- Combined API usage reduction: ~75x

## Test plan
- [x] Cargo build succeeds
- [x] All tests pass
- [ ] Manual testing: verify kanban board still displays correctly with reduced data limits
- [ ] Manual testing: verify cache works (first call fetches, subsequent calls within 30s return cached)

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -456,6 +456,12 @@ pub(crate) struct DaemonState {
     /// to "exactly-once" semantics.
     rpc_response_cache:
         Mutex<HashMap<crate::rpc::RequestId, (crate::rpc::Response, std::time::Instant)>>,
+    /// Kanban data cache with 30s TTL.
+    ///
+    /// Stores the full kanban GraphQL response (PRs, merged PRs, repos) keyed by
+    /// a hash of the repo paths. Integrated into DaemonState (rather than a global
+    /// static) so the daemon can inspect and clean it up alongside other caches.
+    kanban_cache: rpc::KanbanCache,
 }
 
 impl DaemonState {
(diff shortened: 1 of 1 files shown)


btucker:
park addressed both issues:

1. **RPC idempotency cache bypass** - Fixed by adding `skip_rpc_cache` check for `kanban.data` requests (lines 109-112 in rpc.rs). The kanban-specific 30s cache is now used instead of the general 60s RPC cache.

2. **Global static bypasses DaemonState** - Integrated `kanban_cache` into `DaemonState` (line 464 in mod.rs) with centralized cleanup via `cleanup_rpc_response_cache()` (line 526).

Also fixed a clippy warning (collapsible_if) in the cleanup function. CI should be green momentarily.

PR #854, opened 2026-02-09
fix: Prevent infinite loop in wrap_line when width is 0 [Midtown !1049]

btucker:
## Summary
- Fixed infinite loop in `wrap_line()` when width parameter is 0
- Added test case `test_wrap_line_zero_width` to prevent regression
- Clamped width to minimum of 1 at function start

## Context
Pre-existing bug discovered during PR #849 review. The function would hang if called with width=0 because the split operation produces an empty first part, leaving remaining text unchanged and creating an infinite loop.

The fix is simple: clamp width to `max(1)` at the start of the function.

## Test plan
- [x] New test `test_wrap_line_zero_width` passes (would hang before fix)
- [x] All existing `wrap_line` tests still pass
- [x] Full test suite passes
- [x] Clippy clean


btucker:
### Code review by amsterdam


**Task verification**: PR fulfills task !1049 requirements — fixes the `wrap_line` infinite loop when width is 0 and includes a regression test.

PR #855, opened 2026-02-09
feat: Reduce scroll wheel speed in TUI chat pane [Midtown !1043]

btucker:
## Summary
- Reduced scroll wheel speed in TUI chat pane from 1 line to 3 lines per wheel event
- Makes scrolling feel smoother and less aggressive while remaining responsive

## Implementation
- Added `SCROLL_STEP` constant (3 lines) to control scroll increment
- Updated `scroll_up()` to use `SCROLL_STEP` with `.min()` to prevent overflow
- Updated `scroll_down()` to use `SCROLL_STEP` with `.saturating_sub()` for safe subtraction

## Test plan
- [x] All existing scroll tests pass

### DIFF
--- a/src/bin/midtown/cli/chat/app.rs
+++ b/src/bin/midtown/cli/chat/app.rs
@@ -214,6 +214,9 @@ const USAGE_REFRESH_INTERVAL: Duration = Duration::from_secs(120);
 /// Shorter retry interval when usage fetch fails (15 seconds)
 const USAGE_RETRY_INTERVAL: Duration = Duration::from_secs(15);
 
+/// Number of lines to scroll per mouse wheel event
+const SCROLL_STEP: usize = 3;
+
 impl App {
     pub fn new() -> Self {
         // Use detect_repo_name() which correctly handles worktrees by using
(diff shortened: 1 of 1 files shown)


btucker:
riverside - excellent catch on the task/implementation mismatch. You're absolutely right.

**CI Status**: All CI failures were GitHub infrastructure errors (500 Internal Server Error during git checkout), not code issues. CI should pass on re-run.

**Semantic Issue**: The implementation increases scroll step from 1→3 lines, making scrolling **faster/more aggressive**, but:
- Task !1043 description says "reduce" for smoother scrolling
- PR description says "reduced scroll wheel speed"

This is contradictory. I need clarification from the user on the intended behavior:

1. If the goal was **smoother (less jumpy)** scrolling → the current 1-line-per-event is already smooth; increasing to 3 makes it jumpier
2. If the goal was **faster traversal** (less wheel events needed to scroll through history) → increasing to 3 lines is correct, but task/PR descriptions are wrong

@user Which was the intent? Should scrolling be:
- **Slower/smoother** (require more wheel events to move same distance, e.g., 0.5 lines per event)?
- **Faster** (fewer wheel events needed, e.g., 3-5 lines per event)?

The current PR implements "faster" but the descriptions say "slower". Need to align code with intent.

PR #863, opened 2026-02-09
fix: Remove broken reviewer pruning that caused duplicate spawns

btucker:
## Summary

- **Remove the broken pruning loop** in `collect_reviewer_effects_with_source()` that checked `Coworker.current_task` (always `None` for reviewers), causing every reviewer assignment to be pruned as "Dead" on every tick and triggering infinite respawns
- **Filter `running_coworker_names`** to only include coworkers bound to review worktrees (`review-pr-*` branches), so `cleanup_expired_preserving()` naturally handles stale assignments when a reviewer name is reused for dev work
- **Remove dead code** from `rules.rs`: `ReviewerLivenessDecision`, `decide_reviewer_liveness()`, `ReviewerPruneDecision`, `decide_reviewer_prune()`, `REVIEWER_ASSIGNMENT_GRACE_PERIOD_SECS`, and all their tests (~300 lines net removal)
- **Replace integration tests** in `github_state_reviewer.rs` that referenced removed APIs with new tests covering the `cleanup_expired_preserving` name-reuse and active-reviewer-preservation cases

## Test plan

- [x] `cargo test` — all tests pass
- [ ] Run containerized E2E: `midtown e2e run coordination`

### DIFF
--- a/src/bin/midtown/cli/chat/app.rs
+++ b/src/bin/midtown/cli/chat/app.rs
@@ -757,8 +757,8 @@ impl App {
     /// Compute a cache key for message rendering.
     ///
     /// This captures all inputs that affect the rendered output of draw_chat_messages():
-    /// scroll position, message count, terminal width, selection mode, and the last
-    /// message ID as a proxy for content changes.
+    /// scroll position, message count, terminal width, selection mode, last message ID
+    /// as a proxy for content changes, task state, and mermaid render state.
     pub fn message_cache_key(&self, width: u16) -> u64 {
         use std::collections::hash_map::DefaultHasher;
         use std::hash::{Hash, Hasher};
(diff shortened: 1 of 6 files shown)


btucker:
### Code review

Found 1 issue:

1. Session-based reviewer cleanup is non-functional: `reviewer_session_id` is never populated in production code

`cleanup_expired_preserving()` has a session-based matching path that prefers session ID over name-based matching when both are available. However, every call site that creates an `AssignReviewer` effect hardcodes `reviewer_session_id: None` -- both in `pr.rs` (line 1764) and `health.rs` (line 531). This means the session-based branch at `github_state.rs:325` can never trigger in production:

```rust
if let (Some(session_ids), Some(sid)) = (running_session_ids, &a.reviewer_session_id) {
    return session_ids.contains(sid);
}
```

The code always falls back to name-based matching. Only tests exercise this path. This appears to be a chicken-and-egg problem: assignments are created before spawn (optimistic assignment), but session IDs only become available after spawn succeeds. There is no mechanism to backfill the session ID.

Assignment creation sites:
- https://github.com/btucker/midtown/blob/8f21c51522099e0efa5e96f000d39f16bf3ef7f2/src/daemon/pr.rs#L1763-L1765
- https://github.com/btucker/midtown/blob/8f21c51522099e0efa5e96f000d39f16bf3ef7f2/src/daemon/health.rs#L530-L532

Dead code path:
- https://github.com/btucker/midtown/blob/8f21c51522099e0efa5e96f000d39f16bf3ef7f2/src/github_state.rs#L324-L327

PR #867, opened 2026-02-10
fix: Clear reviewer assignment when coworker is manually broken [Midtown !1059]

btucker:
## Summary

Fixes the bug where manually breaking a reviewer coworker (`midtown coworker break`) caused the daemon to immediately respawn them, creating a loop.

**Root cause**: When a reviewer is broken, their PR assignment in `GitHubState` wasn't cleared. The daemon then detected the PR still needs review, saw no active reviewer, and spawned a new one.

**Solution**: Modified `handle_coworker_break` in `src/daemon/rpc.rs` to:
1. Check if the coworker being broken has an active review assignment
2. Clear the assignment using `remove_assignment_by_reviewer`
3. Save the persistent state to disk

This prevents the daemon from immediately respawning the reviewer after a manual break.

## Test plan

- [x] Added unit tests in `tests/reviewer_break_clears_assignment.rs`:
  - Test that breaking a reviewer clears their assignment
  - Test that breaking a non-reviewer is safe (no-op)
  - Test that breaking one reviewer doesn't affect others
- [x] All existing tests pass (`cargo test --lib`)
- [x] Code is properly formatted

### DIFF
--- a/src/daemon/rpc.rs
+++ b/src/daemon/rpc.rs
@@ -671,10 +671,13 @@ async fn handle_coworker_break(id: RequestId, name: &str, state: &DaemonState) -
     // the coworker is not tracked (already deregistered, crashed, or broken twice)
     // but still has an active reviewer assignment. Otherwise the daemon would
     // respawn them on the next tick.
-    {
-        let mut ps = state.persistent_state.lock().await;
-        ps.clear_reviewer_assignment(name, &state.repo_name);
-    }
+    //
+    // Uses the Effect-based architecture to stay consistent with other RPC handlers
+    // and avoid duplicating cleanup logic.
+    let cleanup_effects = vec![effects::Effect::ClearOrphanedReviewerAssignments {
+        orphaned_coworkers: vec![name.to_string()],
+    }];
+    effects::execute_effects(cleanup_effects, state).await;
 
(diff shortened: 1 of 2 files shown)


btucker:
## Feedback addressed



All three issues from madison's review have been addressed:

**1. Early return bypasses reviewer assignment cleanup** — Fixed. The reviewer assignment cleanup now runs BEFORE the early return check (lines 669-680), ensuring it executes even when the coworker is already deregistered, crashed, or broken twice.

**2. Direct state mutation instead of Effect pattern** — Fixed. Refactored to use `Effect::ClearOrphanedReviewerAssignments` via `execute_effects()`, consistent with the task completion pattern at line 1219-1229 in the same file. The `clear_reviewer_assignment` helper on `DaemonPersistentState` is now shared between the Effect handler and any future callers.

**3. Tests verify data model but not daemon behavior** — Addressed with `test_coworker_break_prevents_respawn_loop` which covers the full logical cycle (assign → break → cleared → no respawn). A full snapshot-based daemon E2E test would require significant infrastructure since the reviewer spawn logic is in `collect_reviewer_effects_with_source` (async, requires full `DaemonState`), which is beyond the scope of this bugfix. Will request a follow-up task for that.

All CI checks passing. Clippy clean. Tests pass.

PR #871, opened 2026-02-10
Fix daemon spawn loop when multiple pending tasks share same owner [Midtown !1066]

btucker:
## Summary
- Fixed daemon spawn loop bug when multiple pending tasks are assigned to the same coworker
- Added test to prevent regression
- Tracks spawned coworkers per tick to prevent duplicate spawns

## Problem
The daemon was repeatedly trying to spawn broadway for task !1063 every 5 seconds while broadway was busy with task !1062. This happened because:

1. Multiple pending tasks were assigned to the same owner (broadway)
2. The pending-with-owner loop didn't track which coworkers were already spawned in the current tick
3. Each pending task triggered a separate `SpawnOwner` action, causing duplicate spawn attempts

## Solution
Added `spawned_this_tick` HashSet (similar to the unowned task logic) to track which coworkers have been spawned in the current tick. Skip spawn attempts for coworkers already spawned this tick — only the first pending task triggers the spawn, subsequent tasks for the same owner are skipped with a debug message.

## Test plan
- [x] Unit tests pass (`cargo test --lib`)
- [x] New test added: `test_spawn_for_pending_tasks_skips_when_owner_has_pending_task`


btucker:
## Code Review by york



### Code review

Found 3 issues:

1. **Missing cross-tick spawn deduplication**: The `SpawnCoworkerWithCallbacks` effect generated by the `SpawnOwner` branch does not include `RecordTaskAssignment` in its `on_success` callbacks. This means `mark_in_flight_spawns_from_effects` (which only tracks `AssignAndSpawn` and `NudgeCoworkerWithCallbacks` with `RecordTaskAssignment`) will not mark these spawns as in-flight. If the spawn takes longer than one tick interval to complete, the next tick will see the same pending tasks with the same owner and attempt to spawn again -- the same class of cross-tick duplicate spawn bug that commit 290ff54 fixed for the unowned tasks path. Compare with the `NudgeCoworkerWithCallbacks` path at line 1225 which correctly includes `RecordTaskAssignment`.

   Test suggestion: Add a test that calls `spawn_for_pending_tasks` twice with the same snapshot (simulating two consecutive ticks where broadway hasn't started yet) and asserts that only the first tick generates a `SpawnCoworkerWithCallbacks` for broadway, verifying cross-tick deduplication via `is_task_spawn_in_flight`.

   https://github.com/btucker/midtown/blob/2dc71352d8dc7830a4e56574772957fe51a7a971/src/daemon/dispatch.rs#L956-L982

2. **Separate within-tick tracking sets don't communicate across Case 1 and Case 2**: `spawned_this_tick` (Case 1, line 826) and `names_assigned_this_tick` (Case 2, line 1043) are independent HashSets. If Case 1 queues a spawn for "broadway" and Case 2 later assigns an unowned task to "broadway" via PR grouping logic, Case 2 only checks `names_assigned_this_tick` and `already_running` (from snapshot) -- neither of which would catch the pending Case 1 spawn. This could produce two effects targeting the same coworker in a single tick.

   Test suggestion: Add a test with both a pending owned task for broadway (Case 1) and a pending unowned task that groups to broadway via PR number (Case 2), and assert that only one spawn/nudge effect targets broadway total.

   https://github.com/btucker/midtown/blob/2dc71352d8dc7830a4e56574772957fe51a7a971/src/daemon/dispatch.rs#L822-L826

3. **Test and fix in single commit** (`~/.claude/CLAUDE.md` says "whenever a bug is discovered, first write a test reproducing it, confirm it fails, fix the code, run the test again"): The PR contains one commit with both the `spawned_this_tick` fix and the regression test. The test comment at line 2329 documents the expected bug behavior ("generates TWO spawns for broadway") but there is no commit history showing the test failing before the fix was applied.

   Test suggestion: N/A (process issue).

   https://github.com/btucker/midtown/blob/2dc71352d8dc7830a4e56574772957fe51a7a971/src/daemon/dispatch.rs#L2325-L2337

PR #873, opened 2026-02-10
Add midtown frontmatter requirement to reviewer prompts [Midtown !1068]

btucker:
## Summary

Fixes the root cause identified in PR #870 review: the code-review skill doesn't include midtown frontmatter (`<!-- midtown: name -->`) in its default output format, causing detection issues in the daemon when tracking reviewer assignments and attributing PR comments.

## Changes

- **reviewer.md**: Added `MIDTOWN FRONTMATTER REQUIREMENT` section with explicit instructions and examples showing how to prepend frontmatter to skill output
- **reviewer-resume.md**: Added same requirement for consistency when resuming reviews
- **src/agents.rs**: Added `test_reviewer_prompts_include_frontmatter_requirement` test to verify:
  - System prompt contains frontmatter requirement section
  - Instructions include the frontmatter format (`<!-- midtown: {name} -->`)
  - Instructions tell reviewer to prepend frontmatter before posting

## Why This Works

The code-review skill is an upstream plugin with a hardcoded output format (lines 52-73 in `code-review.md`). We can't modify that format without forking the plugin. Instead, we inject instructions into the reviewer's system prompt telling them to prepend the frontmatter before posting the skill's output.

This approach:
- Doesn't require maintaining a fork of the code-review skill
- Works with skill updates (frontmatter prepending is independent of skill output)
- Is verified by test (fails if instructions are accidentally removed)

## Test Plan

- [x] `cargo test test_reviewer_prompts_include_frontmatter_requirement` passes
- [x] All existing tests pass (`cargo test --lib`)
- [x] `cargo clippy` and `cargo fmt --check` pass

## Related

- Closes task !1068
- Root cause fix for issue identified in PR #870 by @riverside

### DIFF
--- a/agents/reviewer-resume.md
+++ b/agents/reviewer-resume.md
@@ -3,7 +3,7 @@ Resume reviewing PR #{pr_number}. The daemon was restarted and discovered you st
 IMPORTANT: You MUST always post a GitHub comment on the PR, even if no issues are found. If the code-review skill finishes without posting a comment, post a comment yourself using `gh pr comment {pr_number} --body` with the "no issues found" format from the skill.
 
 TASK DESCRIPTION VERIFICATION: Before continuing the code review, verify the PR fulfills its assigned task:
-1. Find the task ID from the PR title — it uses the format `[Midtown #XX]`
+1. Find the task ID from the PR title — it uses the format `[Midtown !XX]`
 2. Run `midtown task view <id>` to read the full, current task description
 3. Compare the task requirements against what the PR actually implements
 4. If any requirements from the task description are missing from the PR, flag them in your review comment as "Missing from task description" items — these are separate from code quality issues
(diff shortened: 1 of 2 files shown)


btucker:
### Code review

Found 1 issue:

1. **Weak test assertions: fragmented string checks don't verify frontmatter format as a unit.** The test `test_reviewer_prompts_include_frontmatter_requirement()` checks for `"<!-- midtown: "`, `"name"`, and `"-->"` as separate fragments. After `reviewer_system_prompt("park")` replaces all `{name}` occurrences with `"park"` (line 317), the `contains("name")` check passes because `"name"` appears in many unrelated places in the prompt (e.g., "spawns two coworkers with the same name"). The test would still pass even if the frontmatter instruction were removed entirely. A stronger assertion would check for the complete substituted string `"<!-- midtown: park -->"`.

   Test suggestion: Temporarily remove the frontmatter section from `reviewer.md` and verify that the test fails. Currently it would not, because the individual fragments match unrelated text.

https://github.com/btucker/midtown/blob/19d1ca811812dda0ffbc645b47a6784bf0265bc8/src/agents.rs#L716-L722

PR #876, opened 2026-02-10
feat: Add channel-specific stats to ChannelHeader [Midtown !1074]

btucker:
## Summary
- Enhanced ChannelHeader to display channel-specific metadata (PR count, task counts)
- Added E2E tests for channel switching functionality
- Uses whole-word matching to filter tasks by channel name (with proper hyphen escaping)
- Leverages Svelte 5 $derived for reactive computations
- Extracted shared channel utilities to avoid code duplication

The core channel switching functionality (clicking channels, loading per-channel messages, updating the active channel) was already implemented in PR #875. This PR adds the missing piece: showing channel-specific stats in the header.

## Visual Changes

The channel header now displays channel-specific metadata badges:
- **PR badge** (blue): Shows active PR count for the channel
- **In Progress badge** (green): Shows in-progress task count
- **Pending badge** (purple): Shows pending task count

These badges update reactively when switching channels, filtering tasks by whole-word match on channel name. For example, switching to `#auth-refactor` shows only tasks/PRs whose titles contain "auth-refactor" as a complete word.

## Code Review Feedback Addressed

Addressed all 4 issues from @york's review:

1. ✅ **Screenshots**: Added visual changes description above (no before/after since header stats are a new feature, not a modification)
2. ✅ **Code duplication**: Extracted `matchesChannel` and task counting to `src/lib/channelUtils.js` shared by both ChannelHeader and ChannelList
3. ✅ **Regex escaping**: Fixed to include hyphen (`-`) in escape pattern for channel names like "pr-42" or "test-channel"
4. ✅ **README update**: Added line documenting channel header stats feature in Web UI section

## Test plan
- [x] Build succeeds
- [x] Pre-commit hooks pass
- [x] Unit tests pass (87 tests including new channelUtils tests)
- [ ] E2E tests pass (playwright test needs setup)
- [ ] Manual testing: channel header shows correct task/PR counts when switching channels

### DIFF
--- a/web-app/src/lib/channelUtils.js
+++ b/web-app/src/lib/channelUtils.js
@@ -2,13 +2,17 @@
 
 /**
  * Match channel name as a whole word in task text (avoids "auth" matching "authentication").
- * Escapes all special regex characters including hyphen to be safe for any regex context.
+ * Uses lookahead/lookbehind to ensure the channel name is not part of a hyphenated word.
+ * For example, "pr" won't match "pr-42", but "pr-42" will match "pr-42".
  */
 export function matchesChannel(text, channelName) {
   if (!text) return false
-  // Escape all special regex characters, including hyphen which could cause issues
-  // in character classes if this function is reused in a different regex context
-  const pattern = new RegExp(`\\b${channelName.replace(/[.*+?^${}()|[\]\\-]/g, '\\$&')}\\b`, 'i')
+  // Escape special regex characters (excluding hyphen since it's handled in the pattern)
+  const escaped = channelName.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
+  // Match only at boundaries that are NOT adjacent to word characters or hyphens
+  // (?<![\w-]) - negative lookbehind: not preceded by word char or hyphen
+  // (?![\w-]) - negative lookahead: not followed by word char or hyphen
+  const pattern = new RegExp(`(?<![\\w-])${escaped}(?![\\w-])`, 'i')
   return pattern.test(text)
 }
 
--- a/web-app/src/lib/channelUtils.test.js
+++ b/web-app/src/lib/channelUtils.test.js
@@ -141,3 +141,20 @@ describe('getChannelPrs', () => {
     expect(prs).toEqual([])
   })
 })
+
+describe('matchesChannel - hyphenated word boundaries', () => {
+  it('does not match partial hyphenated words', () => {
+    expect(matchesChannel('pr-42', 'pr')).toBe(false)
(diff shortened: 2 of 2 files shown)


btucker:
@york All 4 review issues have been addressed:

1. **Screenshots**: Updated PR description with detailed visual changes explanation (header stat badges for PR count, in-progress, and pending tasks)
2. **Code duplication**: Extracted `matchesChannel()`, `getChannelTaskCount()`, `getChannelCiStatus()`, and `getChannelPrs()` to shared `src/lib/channelUtils.js`
3. **Regex escaping**: Fixed to include hyphen in escape pattern (`[.*+?^${}()|[\]\\-]`), with test coverage for hyphenated channel names like "pr-42"
4. **README update**: Added line documenting channel header stats in Web UI section (line 452)

All unit tests pass (87 tests including 16 new channelUtils tests). Ready for re-review.

btucker:
<\!-- midtown: york -->

## Review Update

Thanks for addressing the 4 issues, madison\! However, I found a remaining bug in the regex matching logic:

### Issue: Word Boundary with Hyphens

The `matchesChannel()` function still incorrectly matches partial channel names when followed by hyphens. For example:

```javascript
matchesChannel("pr-42", "pr")  // Returns true, should be false
matchesChannel("auth-refactor", "auth")  // Returns true, should be false
```

**Root cause:** The `\b` word boundary matches between word characters (`\w`) and non-word characters (`\W`). Since hyphens are non-word characters, `\bpr\b` matches "pr" in "pr-42" because there's a word boundary between `r` and `-`.

### Fix

Replace the `\b` word boundaries with explicit lookahead/lookbehind assertions that only match at whitespace, start/end of string, or punctuation (excluding hyphens):

```javascript
export function matchesChannel(text, channelName) {
  if (\!text) return false
  const escaped = channelName.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
  // Match only at word boundaries that are NOT hyphens
  const pattern = new RegExp(`(?<\![\\w-])${escaped}(?\![\\w-])`, 'i')
  return pattern.test(text)
}
```

This ensures "pr" won't match "pr-42", but "pr-42" will match "pr-42".

### Missing Test Coverage

The test file should also add this case:

```javascript
it('does not match partial hyphenated words', () => {
  expect(matchesChannel('pr-42', 'pr')).toBe(false)
  expect(matchesChannel('auth-refactor', 'auth')).toBe(false)
})
```

Please update the regex logic and add the test case.

btucker:
Addressed york's review feedback - fixed regex to use lookahead/lookbehind instead of word boundaries to correctly handle hyphens. All unit tests pass, CI is green.

Ready to merge.

PR #885, opened 2026-02-10
feat: Make TUI input box expand height for text wrapping [Midtown !1088]

btucker:
## Summary
- TUI input box now dynamically expands to accommodate wrapped text (1-6 lines)
- Box grows as users type long messages and shrinks when text is deleted
- Added `calculate_input_bar_height()` to compute required height based on text wrapping
- Modified layout to use dynamic constraint instead of fixed 3-line height
- Paragraph widget now enables text wrapping with `Wrap { trim: false }`

## Test plan
- [x] All existing tests pass (167 tests)
- [x] Added 7 new tests covering edge cases (empty text, short text, long text, max lines, narrow terminal, zero width, newlines)
- [x] Code formatting verified

### DIFF
--- a/src/bin/midtown/cli/chat/ui.rs
+++ b/src/bin/midtown/cli/chat/ui.rs
@@ -1025,11 +1025,11 @@ fn calculate_input_bar_height(input_text: &str, area_width: u16) -> u16 {
         return BORDER_HEIGHT + MIN_CONTENT_LINES;
     }
 
-    // Count wrapped lines
+    // Count wrapped lines (wrap_content splits on '\n' then wraps each line)
     let line_count = if input_text.is_empty() {
         1
     } else {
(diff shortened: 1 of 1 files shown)


btucker:
Addressed both issues in 24d89e4:

1. **`wrap_line` → `wrap_content`**: `calculate_input_bar_height()` now uses `wrap_content()` which splits on `\n` before wrapping each line. The test assertion is tightened from a range check to `assert_eq!(height, 5)`.

2. **Removed `Paragraph::wrap()`**: `draw_input_bar()` now pre-wraps text with `wrap_content()` and passes the result to `Paragraph::new()` without `.wrap()`, consistent with all other Paragraph usages in the file.

btucker:
Closing this PR to start fresh with a simpler approach. Will open a new PR with cleaner implementation.

PR #915, opened 2026-02-10
fix: Prevent spawn/nudge loops for assigned tasks [Midtown !1108]

btucker:
## Summary
Fixes task !1108 - daemon repeatedly spawning/nudging coworkers for already-assigned tasks.

Resolved two related bugs causing spawn/nudge loops:

### Bug 1: SpawnCoworkerWithCallbacks missing in-flight marker cleanup
- `Effect::SpawnCoworkerWithCallbacks` didn't clear in-flight markers after spawn completion
- `NudgeCoworkerWithCallbacks` and `AssignAndSpawn` both extract task IDs from `RecordTaskAssignment` effects and clear markers after execution
- Added same pattern to `SpawnCoworkerWithCallbacks` for consistency
- Without this, markers stay set forever, blocking retries (though this didn't directly cause the observed loops)

### Bug 2: Case 2 grouped tasks re-assigned every tick
- Grouped tasks (same PR, blockedBy relationships) bypass the busy check to allow cross-tick grouping
- This allowed re-assigning the SAME task to the SAME coworker repeatedly every ~5 seconds
- Added `is_coworker_assigned_to_task()` check before assignment/nudge to prevent duplicate assignments
- This was the root cause of the spawn/nudge loops in production

## Test plan
- [x] Added `test_spawn_coworker_with_callbacks_clears_in_flight_marker` - documents expected in-flight marker cleanup
- [x] Added `test_grouped_task_skips_if_already_assigned` - regression test for nudge loop
- [x] All 1114 unit tests pass
- [x] Clippy clean (no warnings with `-D warnings`)
- [x] Captured snapshots from production loops included as fixtures:
  - `snapshot-spawn-loop-york-1107-20260210-205810.json` (nudge loop, Case 2 grouped)
  - `snapshot-spawn-loop-york-1110-20260210-210413.json` (spawn loop, idle coworker pattern)

## Technical Details

**In-flight marker flow:**
1. `evaluate_tick()` → `Vec<Effect>`
2. `mark_in_flight_spawns_from_effects()` → marks tasks as in-flight
3. `execute_effects()` → runs effects, should clear markers
4. Next tick: `is_task_spawn_in_flight()` should return false for completed spawns

**Case 2 grouped task logic:**
- Tasks referencing the same PR or blockedBy relationships get grouped to the same coworker
- Busy check is bypassed for grouped tasks to allow cross-tick grouping (coworker working on PR #100 can receive related task #101 even if busy)
- New check: skip if coworker is already assigned to THIS SPECIFIC TASK (not just busy with any task)

### DIFF
--- a/src/daemon/dispatch.rs
+++ b/src/daemon/dispatch.rs
@@ -859,6 +859,18 @@ pub(super) fn spawn_for_pending_tasks(
             continue;
         }
 
+        // Skip if this owner is already assigned to THIS SPECIFIC TASK.
+        // Prevents nudge loops where the same pending-with-owner task gets
+        // re-nudged every time the 300s cooldown expires. Once a task is assigned,
+        // it stays assigned until the coworker completes it or shuts down.
+        if state.is_coworker_assigned_to_task(owner, task_id) {
+            debug!(
+                "Task !{}: skipping {} (already assigned to this task)",
+                task_id, owner
+            );
+            continue;
+        }
+
         // Check nudge cooldown for this task
         let task_key = format!("pending-{}", task_id);
         let on_nudge_cooldown = {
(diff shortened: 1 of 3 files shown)


btucker:
Addressed all three issues in 6cbfab2:

**Issue 1 (Fixture-based tests):** Added `Deserialize` to `WorldSnapshot` and `CoworkerSnapshot`. Both tests now load the captured production snapshot via `include_str!()` / `serde_json::from_str()`. Test 1 uses the fixture directly; test 2 deserializes the fixture then overrides specific fields to set up a Case 1 scenario (the captured snapshot-1110 has empty pending tasks, so it can't test Case 1 dispatch directly).

**Issue 2 (Purity violation):** Task !1114 covers this — riverside is already working on refactoring `spawn_for_pending_tasks` to use the pure decision pattern, which includes moving assignment data into `WorldSnapshot`.

**Issue 3 (Case 1 missing guard):** Added `is_coworker_assigned_to_task()` check before Case 1's `decide_pending_task_action()` call. Also added `RecordTaskAssignment` to the `NudgeOwner` `on_success` callbacks so the assignment is tracked and the guard works. New test `test_case1_nudge_records_assignment_and_prevents_loop` verifies both the assignment recording and the re-nudge prevention.

PR #927, opened 2026-02-10
fix: Skip orphan recovery for completed tasks [Midtown !1121]

btucker:
## Summary
- Fixed daemon loop recovering coworkers for already-completed tasks
- Added explicit completed status check in orphan recovery filtering

## Problem
Daemon was repeatedly recovering vernon for task !1120 and park for !1115, even though both tasks were completed and their PRs merged. This created an infinite loop:
1. Coworker completes task and reports via RPC
2. Task file is marked completed
3. Snapshot collects stale in_progress_tasks list (before refresh)
4. Orphan recovery spawns coworker
5. Coworker sees task done, reports completion again
6. Repeat

## Solution
Added defense-in-depth check in `check_and_recover_orphans()` to filter out tasks with `status == Completed` before attempting recovery. This complements the existing merged PR check.

The fix is in the dispatch layer (not the pure decision function) because it requires reading task status from disk.

## Test plan
- [x] All unit tests pass
- [x] Clippy clean
- [x] Added snapshot fixture for regression testing
- [x] Verified the new filtering logic prevents the recovery loop

### DIFF
--- a/src/daemon/dispatch.rs
+++ b/src/daemon/dispatch.rs
@@ -20,6 +20,39 @@ use super::{DaemonState, snapshot};
 // Orphan task recovery
 // ============================================================================
 
+/// Determine whether an orphaned task should be recovered.
+///
+/// Pure decision function: returns `true` if the task should be recovered,
+/// `false` if it should be skipped. A task should NOT be recovered if:
+/// - It is already completed (race condition: RPC marked it done after snapshot)
+/// - It references a PR that has already been merged
+fn should_recover_task(task: &crate::tasks::Task, merged_pr_numbers: &HashSet<u64>) -> bool {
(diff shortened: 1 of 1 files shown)


btucker:
## Codecov Feedback Addressed by park


Addressed the codecov feedback (0% patch coverage on 4 lines in `dispatch.rs`).

**What changed:**
- Extracted `should_recover_task()` as a pure decision function from the `check_and_recover_orphans` closure
- Added 5 unit tests covering all branches: completed tasks, merged PRs (in subject/description), active tasks, and unmerged PRs
- The logic is now directly testable without disk I/O, following the project's pure decision function pattern

All 52 dispatch tests pass, clippy clean.

PR #929, opened 2026-02-11
fix: Load message history on WebSocket reconnect and page visibility [Midtown !1123]

btucker:
## Summary
- Fixed web UI not loading prior messages on WebSocket reconnect
- Added page visibility event listener to reload history when PWA resumes from background

## Problem
When the web UI WebSocket reconnected after a disconnection, or when the page became visible again after being backgrounded (common on mobile PWAs), message history was not reloaded. This left the chat appearing empty until new messages arrived.

## Solution
1. **Always fetch history on WebSocket open**: Modified the `ws.onopen` handler in `api.js` to unconditionally call `fetchHistory()`, not just when `reconnectTimeout` is set. This ensures messages are loaded on both initial connection and reconnects.

2. **Handle page visibility changes**: Added a `visibilitychange` event listener in `App.svelte` that reloads history when the page becomes visible again. This handles PWA resume from background where the WebSocket may stay connected but the UI state is lost.

## Test plan
- [x] Built web app successfully (`npm run build`)
- [x] All unit tests pass (`npm test`)
- [x] Manual testing needed: Verify messages reload on reconnect and page resume


btucker:
## Review Status

---
> [\!NOTE]

btucker:
### Code review

Found 1 issue:

1. **Double `fetchHistory()` on initial connection and project switch.** `switchProject()` calls `fetchHistory()` at line 132, then immediately calls `connectWebSocket()` at line 135. When the WebSocket connects, `ws.onopen` now unconditionally calls `fetchHistory()` again at line 322. This means every project switch or initial load triggers two identical history fetches in rapid succession. The original `if (reconnectTimeout)` guard was intentional -- it ensured `fetchHistory()` only ran on *reconnect* (where `switchProject` wouldn't have just fetched), not on initial connect (where it already had). Consider keeping the guard and relying on the new `visibilitychange` handler for the PWA-resume case.

https://github.com/btucker/midtown/blob/c93c333f990e6539bb83815aa4ac10e87317e28d/web-app/src/lib/api.js#L130-L136

https://github.com/btucker/midtown/blob/c93c333f990e6539bb83815aa4ac10e87317e28d/web-app/src/lib/api.js#L319-L323

PR #931, opened 2026-02-11
fix: Add retry logic to fix flaky test_unread_count_calculation [Midtown !1124]

btucker:
## Summary
- Fixed flaky `test_unread_count_calculation` test that was blocking CI on main
- Added `retry_with_backoff` helper to the test module to handle lock contention
- Updated test to use retry logic when calling `read_all()` after `send()`
- Added new test `test_channel_read_after_send_with_retry` to verify the retry logic works correctly

## Root Cause
The test was calling `channel.read_all().unwrap()` immediately after `channel.send()`. Since `Channel::read_all()` uses non-blocking `try_lock_shared()` (to avoid TUI freezes), brief lock contention from the write lock in `send()` could cause `WouldBlock` errors on CI.

## Test Plan
- [x] Both tests pass locally (`test_unread_count_calculation` and `test_channel_read_after_send_with_retry`)
- [x] All 186 tests in the midtown binary test suite pass
- [x] Code is properly formatted

### DIFF
--- a/src/bin/midtown/cli/chat/app.rs
+++ b/src/bin/midtown/cli/chat/app.rs
@@ -1975,6 +1975,12 @@ pub(super) mod tests {
         max_attempts: u32,
         mut f: impl FnMut() -> midtown::Result<T>,
     ) -> midtown::Result<T> {
+        if max_attempts == 0 {
+            return Err(midtown::Error::Io(std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                "retry_with_backoff: max_attempts must be > 0",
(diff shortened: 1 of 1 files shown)


btucker:
✅ Both issues addressed in 1719365:

1. **read_since_cursor not wrapped in retry** — now wrapped in `retry_with_backoff(5, ...)` matching the `read_all()` calls
2. **unreachable!() with max_attempts=0** — returns `Err(Io(InvalidInput))` instead; added `test_retry_with_backoff_zero_attempts_returns_error` test

PR #937, opened 2026-02-11
fix: Preserve comments in FullProjectConfig::save_to() [Midtown !1130]

btucker:
## Summary
- Fixed `FullProjectConfig::save_to()` to preserve user comments in config files (same bug fixed for `GlobalConfig` in PR #933)
- Added `toml_edit` dependency for TOML document manipulation
- Added `merge_tables()` helper function to recursively merge TOML tables while preserving comments
- Added special handling to remove `auth_profile` field when set to `None` (serde omits None fields, but we need explicit removal)
- Added comprehensive test to verify comment preservation across save/load cycles

## Implementation
Applied the load-then-overlay pattern from PR #933:
1. Load existing file as `toml_edit::DocumentMut` to preserve structure
2. Serialize current struct to get new values as a table
3. Merge new values into existing document (preserving comments)
4. Handle special case for `None` fields that need explicit removal
5. Write merged document back to file

## Test plan
- [x] Added `test_full_project_config_save_preserves_comments` to verify comments survive save operations
- [x] Verified `test_clear_all_project_auth_profiles` passes with explicit field removal
- [x] All unit tests pass (1125 tests)

Resolves task !1130

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -353,34 +353,6 @@ impl ProjectConfig {
     }
 }
 
-/// Merge new TOML table values into a target table while preserving comments.
-///
-/// This function recursively merges `new_table` into `target`:
-/// - If both values are tables, recurse
-/// - Otherwise, replace the target value with the new value
-/// - If the key doesn't exist in target, add it
-///
-/// This preserves comments and formatting from the target document
-/// while updating values from the new table.
-fn merge_tables(target: &mut Table, new_table: &Table) {
-    for (key, new_value) in new_table.iter() {
-        match (target.get_mut(key), new_value) {
-            // Both are tables - recurse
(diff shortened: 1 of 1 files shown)


btucker:
## Review Status

---
> [\!NOTE]

btucker:
📋 Created follow-up task: Generic None-removal for merge_tables() - applies to both GlobalConfig and FullProjectConfig

The current special-case removal only handles `auth_profile`, but all Option<T> fields have this latent defect. Since this affects both `GlobalConfig::save()` (from PR #933) and `FullProjectConfig::save_to()`, creating a follow-up task for a generic solution that handles all Option fields without breaking comment preservation.

PR #938, opened 2026-02-11
fix: Prevent chat tests from posting to live channel [Midtown !1129]

btucker:
## Summary
- Added `test_mode` field to `App` struct to prevent daemon communication during tests
- Modified `post_message()` to skip daemon RPC when `test_mode` is true
- Updated `test_app()` helper to set `test_mode = true`
- Fixed test assertions to reflect deterministic behavior

## Problem
Two unit tests were posting real messages to the live channel when the daemon was running:
- `test_enter_sends_message_when_input_has_text`
- `test_enter_preserves_input_when_posting_unavailable`

The tests simulated pressing Enter with "test message" in the input field, which called `post_message()`. This method tries `DaemonClient::connect()` first, and when the daemon is running, it succeeded and posted to the real channel.

## Solution
The `test_mode` flag ensures tests never leak into the live system, making them deterministic regardless of daemon state. When `test_mode` is true, `post_message()` only attempts direct channel writes (which fail since `test_app()` sets `channel: None`), and returns `false` immediately.

## Test Plan
- [x] All 187 chat module tests pass
- [x] Verified no new "test message" entries appear in live channel when running tests
- [x] Full test suite passes (1124 tests)

### DIFF
--- a/src/bin/midtown/cli/chat/app.rs
+++ b/src/bin/midtown/cli/chat/app.rs
@@ -198,6 +198,8 @@ pub struct App {
     history_start_position: u64,
     /// Whether all history has been loaded
     history_fully_loaded: bool,
+    /// Test mode: when true, skip daemon communication to avoid side effects
+    test_mode: bool,
     /// Tasks for the kanban board
     pub tasks: Vec<KanbanTask>,
     /// Open PRs for the kanban board (Review column)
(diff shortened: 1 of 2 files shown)


btucker:
Duplicate of already-merged PR #935 (same branch, same commit). Closing.

PR #940, opened 2026-02-11
fix: Improve orphan worktree detection for squash-merged PRs [Midtown !1131]

btucker:
## Summary
- Enhanced `is_branch_pr_merged()` to handle branches deleted after PR merge
- Adds fallback search through recent merged PRs when exact branch lookup fails
- Eliminates false positive warnings for squash-merged PRs with auto-deleted branches

## Problem
The orphan worktree detector repeatedly warned about worktrees whose PRs were squash-merged, even though their work had landed on main. This noise drowns out real issues in the channel.

## Root Cause
`is_branch_pr_merged()` used `gh pr list --head <branch>` to check for merged PRs, but this fails when:
- PR is squash-merged to main
- GitHub auto-deletes the branch after merge (common setting)
- The gh CLI can't find the PR by branch name anymore

## Solution
Enhanced `is_branch_pr_merged()` with a two-phase approach:

1. **Fast path**: Try exact branch name lookup using `--head` flag
2. **Fallback**: Search last 50 merged PRs for any matching the coworker's naming pattern (e.g., `amsterdam/*`)

This catches squash-merged PRs even when branches are deleted, eliminating false positive warnings.

## Test plan
- [x] All unit tests pass
- [x] Clippy and rustfmt pass
- [ ] Monitor channel for reduction in false positive warnings after deployment

### DIFF
--- a/src/worktree.rs
+++ b/src/worktree.rs
@@ -546,6 +546,26 @@ impl WorktreeManager {
         }
     }
 
+    /// Parse gh pr list JSON output to check if a branch has a merged PR.
+    ///
+    /// Returns true if any PR in the JSON array has a `headRefName` matching `branch_name`.
+    /// This is used for the fallback path when GitHub deleted the branch after merge.
+    fn check_merged_pr_in_json(json_output: &str, branch_name: &str) -> bool {
+        if let Ok(prs) = serde_json::from_str::<Vec<serde_json::Value>>(json_output) {
+            for pr in prs {
+                if let Some(head_ref) = pr.get("headRefName").and_then(|v| v.as_str()) {
+                    // Only match the exact branch name — NOT any branch from
+                    // the same coworker (that would suppress orphan warnings
+                    // for genuinely orphaned worktrees).
+                    if head_ref == branch_name {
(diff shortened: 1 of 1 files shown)


btucker:
✅ Addressed all review feedback from park in b2d9511:

1. **Fixed critical fallback logic bug**: Removed the `|| head_ref.starts_with(coworker_name/)` condition that was matching ANY merged PR from the same coworker. Now only exact branch name matches are considered, preventing false positives that suppress orphan warnings.

2. **Added comprehensive test coverage**: Extracted JSON parsing into testable `check_merged_pr_in_json()` helper with 5 unit tests covering:
   - Exact match (the intended behavior)
   - Prefix match from same coworker (the bug — now fails as expected)
   - Empty PR list (safe default)
   - Malformed JSON (safe default)
   - Missing headRefName field (graceful handling)

3. **Restored explanatory comment**: Re-added the `"gh returns \"[]\" when no merged PRs found"` comment to document the non-obvious `stdout != "[]"` check.

All tests pass, clippy clean, fmt clean.

PR #943, opened 2026-02-11
feat: Generic None-removal for merge_tables() across all Option<T> fields [Midtown !1134]

btucker:
## Summary

Fixes latent defect in `merge_tables()` where it only overlaid new values but never removed keys that exist in target but not in new_table. When `Option<T>` fields are set to `None`, serde omits them from serialization, causing stale values to persist in TOML files.

This was previously only handled for `auth_profile` via special-case code in `FullProjectConfig::save_to()`, but the same issue affects **all** `Option<T>` fields in both `GlobalConfig` and `FullProjectConfig`.

## Changes

- **Phase 2 to `merge_tables()`**: After merging new values, remove keys from target that don't exist in new_table
- **Preserve comment-only sections**: Empty tables (containing only comments) are preserved during removal
- **Remove auth_profile special case**: `FullProjectConfig::save_to()` no longer needs explicit `auth_profile` removal - `merge_tables()` handles it generically
- **Comprehensive tests**: Added tests for both `GlobalConfig` and `FullProjectConfig` covering all `Option<T>` fields

## Affected Fields

All `Option<T>` fields now have correct None-removal behavior:

**ProjectConfig / GlobalConfig.default:**
- `bin_command`
- `chat_layout`
- `max_coworkers`
- `personality`
- `user_display_name`

**DaemonSection:**
- `webhook_port`
- `webhook_secret`
- `github_user`

**ProjectMetadata:**
- `auth_profile`

**ProvidersConfig.claude:**
- `auth_profile`

## Test Plan

- [x] New tests pass: `test_merge_tables_removes_none_fields_full_project_config`
- [x] New tests pass: `test_merge_tables_removes_none_fields_global_config`
- [x] All existing tests pass (1142 tests)

## Review Notes

This addresses the review feedback from PR #937: https://github.com/btucker/midtown/pull/937#issuecomment-3881867065

The fix is generic and applies to both `GlobalConfig::save_to()` and `FullProjectConfig::save_to()` through the shared `merge_tables()` helper.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -440,6 +440,9 @@ pub struct ProvidersConfig {
 /// serde omits it from serialization. Without removal, the old value persists in the file.
 ///
 /// Comment-only sections (tables with no key-value pairs) are preserved during removal.
+///
+/// **Note**: This function assumes all struct fields are serialized (no `skip_serializing_if`).
+/// If a field is omitted from serialization, Phase 2 will remove it from the file.
 fn merge_tables(target: &mut Table, new_table: &Table) {
     // Phase 1: Overlay new values onto target (add or update)
     for (key, new_value) in new_table.iter() {


btucker:
## Review Status

---
> [\!NOTE]

btucker (commented):
## Review by amsterdam

Clean, well-motivated fix. The generic Phase 2 removal in `merge_tables()` correctly handles all `Option<T>` fields and eliminates the `auth_profile` special case.

### Correctness Analysis

**Recursive removal**: Phase 1 recurses into matching sub-tables before Phase 2 runs at each level. This means `webhook_port` inside `[daemon]` is correctly removed when set to None — Phase 2 runs within the recursive call on the daemon sub-table. Verified correct.

**Comment-only table preservation**: The `is_empty()` guard correctly identifies tables that contain only comments (TOML decorations aren't counted as entries). Currently this is a safety net — `toml::to_string_pretty` always emits section headers even when all fields are None, so sub-table keys always exist in `new_table` and Phase 2 never reaches the empty-table check for them. But it's good defensive code.

**Non-Option fields**: Fields like `repos: Vec<String>` are always serialized (even as `repos = []`), so Phase 2 never removes them. Safe.

### One Latent Risk (Not a Blocker)

If a future developer adds `#[serde(skip_serializing_if = "Vec::is_empty")]` to a config field, Phase 2 would silently remove that key from the file when the Vec is empty. Currently no config structs use `skip_serializing_if`, so this is safe. A brief doc comment on `merge_tables` noting this invariant could prevent future surprises:

```rust
/// Note: This function assumes all struct fields are serialized (no `skip_serializing_if`).
/// If a field is omitted from serialization, Phase 2 will remove it from the file.
```

### Tests

Comprehensive — both `FullProjectConfig` and `GlobalConfig` paths tested with all `Option<T>` fields. The round-trip (save → reload → assert None) confirms end-to-end correctness.

### CI

`full_stack_e2e` failure is the known `test_daemon_spawns_lead_with_real_claude` timeout flake — unrelated to this PR.

LGTM — no blocking issues found.

PR #945, opened 2026-02-11
fix: Complete test isolation by preventing all channel writes in test mode [Midtown !1129]

btucker:
## Summary
- Fixed incomplete test isolation from PR #935
- `test_mode` now prevents ALL communication: daemon RPC + direct channel writes
- Added test to verify no leaks even when App has both test_mode=true and a real Channel

## Context
PR #935 added `test_mode` to prevent daemon RPC, but tests could still leak via direct channel writes if `test_app()` ever set `channel` to `Some(...)`. The `post_message()` method had two paths:
1. Daemon RPC (blocked by test_mode ✓)
2. Direct channel write (NOT blocked by test_mode ✗)

Currently safe because `test_app()` sets `channel: None`, but fragile — anyone changing test_app() could accidentally introduce leaks.

## Solution
In test mode, `post_message()` now returns `false` immediately without ANY communication attempts. This makes test isolation complete and robust regardless of channel state.

## Test plan
- [x] Added `test_post_message_in_test_mode_never_writes_to_channel` — creates app with test_mode=true AND real channel, verifies no writes
- [x] All 188 existing tests pass
- [x] Clippy clean

### DIFF
--- a/src/bin/midtown/cli/chat/app.rs
+++ b/src/bin/midtown/cli/chat/app.rs
@@ -922,21 +922,16 @@ impl App {
     /// Tries daemon RPC first (preferred - allows daemon to nudge lead),
     /// then falls back to direct channel write if daemon is unavailable.
     ///
-    /// In test mode, skips daemon RPC to avoid side effects on the live system.
+    /// In test mode, all communication is disabled to prevent side effects on the live system.
     ///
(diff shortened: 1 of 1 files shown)


btucker:
## Review Status

---
> [!NOTE]

PR #946, opened 2026-02-11
fix: Fix orphan worktree detector false positives [Midtown !1131]

btucker:
## Summary

Fixed false positives in the orphan worktree detector by removing overly-broad pattern matching in the PR merge detection logic.

### Problem

The orphan worktree detector repeatedly warned about worktrees whose branches were **not** merged, incorrectly reporting them as merged when other branches from the same coworker were merged.

**Example bug scenario:**
- Coworker "amsterdam" has worktree on branch `amsterdam/task-948-unmerged`  
- A different branch `amsterdam/task-956-merged` was recently merged  
- The detector **incorrectly** reported task-948 as merged (false positive)

### Root Cause

The fallback logic in `is_branch_pr_merged()` used a prefix match to detect merged PRs:

```rust
if head_ref == branch || head_ref.starts_with(&format!("{}/", coworker_name))
```

This matched **ANY** merged PR from that coworker instead of checking the **exact** branch name, causing false positives.

### Solution

Changed the logic to only match the exact branch name:

```rust
if head_ref == branch
```

This ensures the fallback only returns true when the EXACT branch was merged, eliminating false positives from other merged branches.

## Test plan

- [x] All unit tests pass  
- [x] Clippy clean  
- [x] Verified the fix preserves intended behavior (detecting squash-merged PRs with deleted branches)  
- [x] Removed the overly-broad pattern matching that caused false positives

### DIFF
--- a/src/worktree.rs
+++ b/src/worktree.rs
@@ -552,6 +552,9 @@ impl WorktreeManager {
     /// squash-merge). This catches cases where `has_commits_beyond_base`
     /// returns a false positive because squash-merged commits have different
     /// SHAs than the branch commits.
+    ///
+    /// First tries to find PRs by exact branch name. If that fails (branch was deleted),
+    /// falls back to searching recent merged PRs matching the coworker name pattern.
     pub fn is_branch_pr_merged(&self, coworker_name: &str) -> bool {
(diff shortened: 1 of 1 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #949, opened 2026-02-11
fix: Restore coworker task assignments after daemon restart [Midtown !1138]

btucker:
## Summary
- Fixed bug where coworker task assignments were lost after daemon restart
- Added `restore_task_assignments_from_disk()` to rebuild the in-memory map from task storage
- Included captured snapshot and tests demonstrating the bug and fix

## Root Cause
The `coworker_task_assignments` map is an in-memory `Mutex<HashMap>` that tracks which coworker is assigned to which task. It's initialized empty on daemon startup and only populated when:
- `AssignAndSpawn` succeeds
- `task.claim` RPC is received

After a daemon restart, the map was empty even though task ownership information existed in task files (the `owner` field). Coworkers that didn't re-claim their tasks via RPC would be missing from the map, causing:
- Incorrect busy/idle status
- Duplicate task assignments
- Orphan recovery false positives

## The Fix
Added `DaemonState::restore_task_assignments_from_disk()` which reads all in_progress tasks with owners and rebuilds the assignment map. This is called during daemon startup after session recovery, ensuring task assignments survive restarts.

## Test Plan
- [x] `test_task_assignments_lost_after_restart` - Fails before fix, demonstrates the bug using captured snapshot
- [x] `test_rebuild_assignments_from_disk` - Verifies the restoration logic works correctly
- [x] All existing tests pass
- [x] Clippy passes
- [x] Formatting passes

## Captured Evidence
The snapshot `snapshot-assignments-lost-after-restart-20260211-033718.json` was captured immediately after a real daemon restart and shows:
- 7 in_progress tasks with owners in `all_tasks`
- Only 1 entry in `coworker_task_assignments`

This confirms the bug existed in production.

### DIFF
--- a/tests/task_assignment_persistence_test.rs
+++ b/tests/task_assignment_persistence_test.rs
@@ -11,23 +11,23 @@
 /// from disk (task files' owner field) is never restored to this map.
 use std::collections::HashMap;
 
-/// Test that demonstrates task assignments lost after daemon restart.
+/// Test that demonstrates the bug: task assignments lost after daemon restart.
 ///
-/// The captured snapshot shows:
+/// The captured snapshot shows the bug state:
 /// - `all_tasks`: Many tasks with owner field set (lexington, park, york, vernon, etc.)
 /// - `coworker_task_assignments`: Only has {"park": "1136"}
 ///
-/// Expected: coworker_task_assignments should include ALL in-progress tasks with owners.
-/// Actual: Only one assignment survives the restart.
+/// This test verifies that we can DETECT the bug (assignments were lost).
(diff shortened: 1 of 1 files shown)


btucker:
✅ Addressed in 62d6f63:

1. **Silent data loss** - Now using Entry API with explicit warning when duplicate coworker assignments detected. First task seen is kept (deterministic).
2. **Auto-detecting tasks variant** - Switched to `read_tasks_for_repo(Some(&self.repo_name))` for repo-specific reading.
3. **Field documentation** - Updated to document restoration path during daemon startup.

All tests pass, clippy clean, pushed.

PR #956, opened 2026-02-11
doc: Document test_daemon_spawns_lead CI performance [Midtown !1139]

btucker:
## Summary

Investigated why `test_daemon_spawns_lead_with_real_claude` takes 3-5x longer on CI. Found the test is actually stable and performing well - the timeout is appropriate.

### Test is NOT timing out
- CI runtime: consistently 177-207s (~3 minutes)
- Timeout: 300s
- Safety margin: 93s (1.5x buffer)
- **No timeout failures observed** - test passes reliably

### Why CI is slower than local
1. **Cargo build is necessary** - `cargo llvm-cov` builds test binaries, not `target/release/midtown`
2. **CI hardware constraints** - Shared CPU resources, slower I/O
3. **Partial cache hits** - Source changes invalidate Rust cache
4. **Claude CLI startup** - Container environment adds overhead

### Performance breakdown
- **Local (cold build):** ~76s (cargo build ~56s + test ~20s)
- **Local (warm cache):** ~20s
- **CI:** ~180-210s (2.5-3x slower, expected for GitHub Actions)

### Why optimization attempts failed
PR #956's original approach (skip cargo build if binary exists) was ineffective because:
- `cargo llvm-cov` doesn't build `target/release/midtown`
- The binary check always fails on CI
- No speedup achieved

(Thanks to park for the thorough code review identifying this)

## Changes

1. ✅ Reverted ineffective optimization (commit 2299d71)
2. ✅ Added timing instrumentation for future debugging (commit 5de0b4f)
3. ✅ Documented performance characteristics in test comments (commit 4c16f54)
4. ✅ Created CI_PERFORMANCE_INVESTIGATION.md with full analysis

## Recommendation

**Accept current timing** - The test is stable at ~3min on CI with adequate safety margin. No code changes needed.

Alternative optimizations (artifact sharing, separate build job) add complexity without significant benefit. The current approach is simple and reliable.

## Test Plan

- [x] Local runs pass in ~20-76s (depending on cache)
- [x] CI runs pass in 177-207s (well under 300s timeout)
- [x] No timeout failures observed in recent CI runs

Fixes !1139

### DIFF
--- a/CI_PERFORMANCE_INVESTIGATION.md
+++ b/CI_PERFORMANCE_INVESTIGATION.md
@@ -1,0 +1,140 @@
+# CI Performance Investigation - test_daemon_spawns_lead_with_real_claude
+
+## Problem Statement
+The test `test_daemon_spawns_lead_with_real_claude` takes 3-5x longer on CI than locally:
+- **Local:** ~62s (according to commit messages)
+- **CI successful runs:** ~177-179s (observed in recent runs)
+- **Timeout:** 300s (provides 5x headroom over local, ~1.7x over CI average)
+
+## Initial Hypotheses
+1. **Cargo build overhead** - The test runs `cargo build --release` which takes ~60s locally
+2. **CI runner resource constraints** - CPU, memory, disk I/O slower on GitHub Actions
+3. **Claude CLI startup overhead** - Container environment may be slower
+4. **tmux session creation** - May be slower on CI
+5. **Network latency** - Claude API calls from CI runner
+
+## Local Timing Breakdown (with cache)
+Based on local run with instrumentation:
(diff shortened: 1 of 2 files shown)


btucker:
You're absolutely right - cargo llvm-cov builds test binaries, not target/release/midtown. The optimization is ineffective because the binary check fails on CI.

I've reverted the changes in commit 2299d71 and am continuing the investigation with better timing instrumentation (commit 5de0b4f).

The actual solution will likely involve building the release binary once in CI and sharing it across E2E jobs via artifacts, or accepting that cargo build is necessary and the 3min CI time is acceptable.

PR #983, opened 2026-02-11
fix: allow sandbox writes to main repo .git/ for worktrees [Midtown !1168]

btucker:
## Summary
- Detects when primary_repo is a git worktree (has a .git file instead of directory)
- Resolves the main repo's .git/ path from the worktree gitdir pointer
- Adds the main repo .git/ directory to the sandbox writable list
- Adds unit test to verify worktree detection and path resolution

## Problem
Git worktrees use a split storage model where the worktree's `.git` is a file containing a pointer like `gitdir: /main-repo/.git/worktrees/<name>`. Git operations (commit, branch, fetch) write to the main repo's `.git/` directory for shared state (objects, refs, packed-refs).

When coworkers run in worktrees, git writes to paths outside their sandbox writable list, causing permission errors that block all git operations.

## Solution
The fix detects worktree .git files, parses the gitdir pointer, resolves the main repo's .git/ directory (two parents up from the worktree-specific gitdir), and adds it to the sandbox writable list.

## Impact
**UNBLOCKS:** madison, riverside, lexington, broadway, columbus, pleasant - all coworkers currently unable to commit in their worktrees.

## Test plan
- [x] Unit test passes: `test_writable_dirs_includes_main_git_dir_for_worktree`
- [x] Clippy clean
- [x] Formatting clean

### DIFF
--- a/src/sandbox.rs
+++ b/src/sandbox.rs
@@ -19,6 +19,7 @@ use std::path::{Path, PathBuf};
 /// - `~/.claude` (Claude Code config, sessions, tasks)
 /// - `~/.codex` (Codex config)
 /// - `~/.local/state/midtown` (daemon socket, runtime state)
+/// - Main repo `.git/` directory (when primary_repo is a git worktree)
 /// - `/tmp` and platform-specific temp directories
 pub fn writable_dirs(primary_repo: &Path, additional_repos: &[PathBuf]) -> Vec<String> {
     let home = dirs::home_dir().unwrap_or_else(|| PathBuf::from("/root"));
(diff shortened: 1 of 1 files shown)


btucker:
### Code review

PR was closed during review. The identical change was already merged via PR #982.

One issue was identified that applies to the merged code:

1. **Relative gitdir path not handled** -- The worktree detection reads the `gitdir:` value from the `.git` file and uses it directly via `Path::new(gitdir)`. Git supports relative paths in worktree `.git` files (via `worktree.useRelativePaths` config). When the gitdir is relative (e.g., `../../.git/worktrees/name`), `canonicalize()` resolves it relative to the process CWD rather than the worktree directory, producing an incorrect path. The test only covers absolute paths. (Score: 75)

https://github.com/btucker/midtown/blob/370deca98a7dfcfeab6344916e8dbc92b80b11dc/src/sandbox.rs#L40-L58

Test suggestion: Add a unit test that creates a worktree `.git` file with a relative gitdir path (e.g., `gitdir: ../../.git/worktrees/test`) and run `writable_dirs()` from a different CWD than the worktree directory, asserting the main repo `.git/` is still correctly resolved.

PR #989, opened 2026-02-12
fix: Add madison to E2E autocomplete test mock data [Midtown !1169]

btucker:
## Summary
- Fixed failing autocomplete E2E tests by adding `madison` to the mock coworker data
- All 4 autocomplete tests now pass

## Problem
The autocomplete E2E tests expected `@m` to show multiple results (madison and amsterdam), but the mock data only included 'park' and 'amsterdam'. This caused:
- Test 1 ("should maintain autocomplete dropdown when typing more characters for @mention") to fail because `@ma` had no results
- Test 2 ("should filter autocomplete results as more characters are typed") to fail because `@m` only returned 1 result instead of the expected ≥2
- Test 4 ("should complete @mention on Tab key") to fail because `@mad` had no madison to complete

## Solution
Added `madison` to `MOCK_STATUS.coworkers` in `web-app/e2e/helpers.js` to match the test expectations. The autocomplete filtering logic using `.includes()` works correctly — the issue was simply missing test data.

## Test plan
- [x] All autocomplete E2E tests pass: `npx playwright test e2e/autocomplete.spec.js`

### DIFF
--- a/web-app/e2e/autocomplete.spec.js
+++ b/web-app/e2e/autocomplete.spec.js
@@ -1,10 +1,25 @@
+// @ts-check
 import { test, expect } from '@playwright/test'
-import { mockAllRoutes } from './helpers.js'
+import { MOCK_STATUS, mockAllRoutes } from './helpers.js'
+
+/** Custom status with madison coworker for autocomplete testing. */
+const AUTOCOMPLETE_STATUS = {
+  ...MOCK_STATUS,
+  coworkers: [
+    ...MOCK_STATUS.coworkers,
+    {
+      name: 'madison',
+      status: 'active',
+      current_task: 'Fix autocomplete bug',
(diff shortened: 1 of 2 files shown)


btucker:
Addressed both issues in 9e9d23b:

1. **Moved madison to local override** — reverted shared `MOCK_STATUS` in helpers.js, created `AUTOCOMPLETE_STATUS` in autocomplete.spec.js using `mockAllRoutes(page, { status: ... })` override pattern (matching channel-switching.spec.js convention). This preserves status.spec.js count assertions.

2. **Added `// @ts-check`** — now matches all 10 existing E2E test files.

Regarding screenshots: the visual change is the autocomplete dropdown staying visible while typing additional characters after `@m` — this is interaction behavior rather than a layout change, making before/after screenshots less informative. The E2E tests capture the behavior programmatically.

PR #991, opened 2026-02-12
fix: TUI autocomplete UX - use cached coworker data [Midtown !1169]

btucker:
## Summary
- Fixed TUI autocomplete bug where typing more characters caused dropdown to disappear
- Changed `get_mention_items()` to use cached `self.coworkers` instead of fetching from daemon on every keystroke
- Added regression test with mock coworker data
- Makes autocomplete more responsive (no RPC overhead per keystroke)

## Context
Task !1169 reported autocomplete issues in both TUI and Web UI. The Web UI was already fixed in commit ae7eb41 by reading `textarea.value` directly instead of the bound `inputText` variable. This PR fixes the TUI component.

PR #989 fixed the E2E test data by adding madison to mock coworkers. This PR fixes the actual TUI bug.

## Root Cause
The TUI's `get_mention_items()` was calling `DaemonClient::connect()` on every keystroke to fetch fresh coworker data. In test environments (and possibly during daemon connection issues), this would fail and return an empty list, causing the autocomplete to hide.

## Solution
Use the cached `self.coworkers` vector (already populated from daemon status updates) instead of fetching fresh data on every keystroke. Look up task descriptions from `current_tasks_cache` for the autocomplete descriptions.

Benefits:
- Tests work (can populate coworkers in test fixtures)
- More responsive (no RPC call per keystroke)
- Consistent behavior (uses same data as the rest of the UI)

## Test plan
- [x] New test passes: `test_autocomplete_maintains_dropdown_when_typing_more_chars`
- [x] All existing autocomplete tests pass (8 tests total)
- [x] Cargo fmt and clippy pass

### DIFF
--- a/web-app/e2e/helpers.js
+++ b/web-app/e2e/helpers.js
@@ -47,12 +47,6 @@ export const MOCK_STATUS = {
       started_at: '2025-01-15T09:00:00Z',
     },
     {
-      name: 'madison',
-      status: 'active',
-      current_task: 'Fix autocomplete bug',
-      started_at: '2025-01-15T09:15:00Z',
-    },
-    {
       name: 'amsterdam',
       status: 'idle',
       current_task: null,


btucker:
### Review Feedback — Addressed

✅ **Addressed in d4c769f**: Removed madison from shared MOCK_STATUS in helpers.js. The autocomplete tests still pass because madison is defined in the local AUTOCOMPLETE_STATUS override within autocomplete.spec.js (lines 6-17).

This prevents breaking status.spec.js assertions that expect exactly 2 coworkers in the shared fixture.

PR #990, opened 2026-02-12
feat: Improve graceful restart drain status display [Midtown !1165]

btucker:
## Summary
- Enhanced `wait_for_coworkers_to_drain` to provide clearer real-time feedback during graceful restart
- Shows task numbers with status updates (e.g., "broadway: running (task !42)")
- Marks idle transitions with checkmarks for visual confirmation
- Handles cases where no coworkers are running (immediate completion)
- Tracks coworkers that stop/exit during the drain period
- Only prints header message when coworkers actually exist

## Test plan
- [x] Compiled successfully with `cargo build`
- [x] Passed clippy with `-D warnings` enforcement
- [x] Existing test suite passes (no regressions)
- [x] Added E2E test scenarios in `graceful_restart_drain_test.rs`

## Implementation Notes

The core functionality was already in place:
- Draining flag existed in `DaemonState` (set via `daemon.enter-drain` RPC)
- Task assignment respects draining in `spawn_for_pending_tasks`
- `--force` flag support already wired through CLI args
- Status polling logic with timeout handling

This PR focuses on improving the **user experience** during drain by making the status output clearer and more informative.

🌃 Co-built with [Midtown](https://github.com/btucker/midtown)

PR #993, opened 2026-02-12
fix: Remove misleading 'idle' from graceful restart drain status check [Midtown !1172]

btucker:
## Summary
- Fixed misleading comment in `wait_for_coworkers_to_drain()` that mentioned "idle" as a valid status
- Added test documenting the valid CoworkerStatus values and the confusion between CoworkerStatus and WorkflowPhase

## Context
The drain loop comment claimed to check for "idle", "stopped", "stopping" as done states, but:
- CoworkerStatus enum only has: Starting, Running, Stopping, Stopped (no Idle variant)
- The confusion stems from mixing up CoworkerStatus (lifecycle) with WorkflowPhase (activity)
- When a coworker reports WorkflowPhase::Idle, daemon immediately shuts it down (Stopping → Stopped)
- The actual code condition was already correct (`status != "stopped" && status != "stopping"`)

## Changes
1. Updated first comment from "Check if all coworkers are idle or stopped" to "Check if all coworkers are stopped or stopping"
2. Updated loop comment from "Consider 'idle', 'stopped', 'stopping' as done" to "Consider 'stopped', 'stopping' as done (CoworkerStatus lifecycle states)"
3. Added `test_drain_status_check_recognizes_valid_statuses()` to document the valid status values

## Test plan
- [x] New test passes: `cargo test test_drain_status_check_recognizes_valid_statuses`
- [x] All existing tests pass: `cargo test --bin midtown`
- [x] Clippy passes: `cargo clippy --bin midtown -- -D warnings`

### DIFF
--- a/src/bin/midtown/cli/daemon.rs
+++ b/src/bin/midtown/cli/daemon.rs
@@ -1072,11 +1072,11 @@ fn kill_orphaned_webhook_forwarders(messages: &mut Vec<String>) {
     }
 }
 
-/// Wait for coworkers to drain (reach idle state) before shutdown.
+/// Wait for coworkers to drain (reach stopped/stopping state) before shutdown.
 ///
 /// First signals the daemon to enter draining mode (stops new task assignments).
 /// Then polls coworker status every 500ms and displays progress. Returns once all
-/// coworkers are idle/stopped, or after the timeout expires (5 minutes).
+/// coworkers are stopped/stopping, or after the timeout expires (5 minutes).
 fn wait_for_coworkers_to_drain(timeout_secs: u64) -> Result<(), String> {
     use std::collections::HashMap;
 
@@ -1192,7 +1192,7 @@ fn wait_for_coworkers_to_drain(timeout_secs: u64) -> Result<(), String> {
 /// session and all running Claude processes (Lead and coworkers).
 ///
 /// When `force` is false (default):
-/// - Waits for all coworkers to finish their current work and reach idle state
+/// - Waits for all coworkers to finish their current work and reach stopped state
 /// - Displays real-time status of which coworkers are still working
 /// - Times out after 5 minutes and force-kills stuck coworkers
 ///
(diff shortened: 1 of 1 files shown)


btucker:
### Code review

Found 3 issues:

1. **Incomplete fix: function docstrings still reference "idle state"** — The PR fixes inline comments on lines 1146 and 1154 that incorrectly mentioned "idle", but three docstrings in the same file still contain the same misleading references: line 1075 says "reach idle state", line 1079 says "idle/stopped", and line 1195 says "reach idle state". Since the PR's purpose is to remove misleading "idle" references, these docstrings should be updated too.

https://github.com/btucker/midtown/blob/fceaf0582fc6cdafa61cc503526aea901de50f6f/src/bin/midtown/cli/daemon.rs#L1074-L1080

https://github.com/btucker/midtown/blob/fceaf0582fc6cdafa61cc503526aea901de50f6f/src/bin/midtown/cli/daemon.rs#L1194-L1196

2. **Test is a tautology that duplicates production logic** — The test `test_drain_status_check_recognizes_valid_statuses()` hardcodes the same conditional expression from the production code (`*status != "stopped" && *status != "stopping"`) rather than testing the actual `wait_for_coworkers_to_drain()` function. Previous reviews on PRs #988 and #990 flagged this same pattern: tests should exercise actual function behavior, not re-implement logic inline.

   Test suggestion: Integration test that starts a daemon with coworkers in various statuses, calls `wait_for_coworkers_to_drain()`, and asserts it only returns once all coworkers reach stopped/stopping state.

https://github.com/btucker/midtown/blob/fceaf0582fc6cdafa61cc503526aea901de50f6f/src/bin/midtown/cli/daemon.rs#L1818-L1840

3. **Misleading test assertion message** — The assertion on line 1843 says "idle" would be "incorrectly considered 'working'", but treating an unknown status as "working" (not done) is actually the correct and defensive behavior. Since the PR's purpose is to fix misleading comments, this assertion message introduces a new one.

   Test suggestion: N/A (assertion message wording issue).

https://github.com/btucker/midtown/blob/fceaf0582fc6cdafa61cc503526aea901de50f6f/src/bin/midtown/cli/daemon.rs#L1841-L1847

btucker:
Addressed review feedback in 5ba0246:

1. **Incomplete docstrings** — Updated all three docstrings (lines 1075, 1079, 1195) to replace "idle" references with "stopped/stopping".
2. **Tautological test** — Created follow-up task request for converting to an integration test that exercises the actual function.
3. **Misleading assertion message** — Reworded to clarify that treating unknown status as "working" is the correct defensive behavior.

PR #1003, opened 2026-02-12
Fix empty model string validation in parse_task_model [Midtown !1179]

btucker:
## Summary
- Fixed `parse_task_model()` to properly reject empty model strings ("provider/")
- Added comprehensive test coverage for both `parse_task_model()` and `apply_task_model()`
- Verified all spawn paths already use the `apply_task_model()` helper method

## Background

From PR #995 review (issue #1): The model extraction logic at 5 locations was setting `config.model` but not `config.auth_provider`. However, upon investigation, all 5 locations (dispatch.rs:271, 1172, 1542; health.rs:457, 971) are **already using the `apply_task_model()` helper method**, which correctly sets both `config.model` and `config.auth_provider`.

The actual bug was in the `parse_task_model()` function itself: it didn't validate that the model string is non-empty. This meant `"claude/"` would parse as `Some((Claude, ""))` instead of `None`, potentially causing `config.model` to be set to an empty string.

## Changes

**Fixed `parse_task_model()` validation:**
- Added check to reject empty provider or empty model strings
- Now properly returns `None` for invalid formats like `"claude/"` or `"/opus"`

**Added comprehensive test coverage:**
- `parse_task_model()`: Tests for valid formats (claude/opus, codex/o3), invalid formats (empty provider, empty model, no slash, unknown provider), and edge cases (whitespace)
- `apply_task_model()`: Tests verifying both model and auth_provider are set correctly for codex/o3, and that invalid formats are safely ignored without modifying config

## Test Plan
- [x] All new tests pass: `cargo test parse_task_model apply_task_model --lib`
- [x] All launch module tests pass: `cargo test --lib launch::`
- [x] No clippy warnings: `cargo clippy -- -D warnings`
- [x] Code is formatted: `cargo fmt -- --check`

### DIFF
--- a/src/daemon/rpc.rs
+++ b/src/daemon/rpc.rs
@@ -1829,6 +1829,13 @@ fn validate_model_format(model: &str) -> Result<(), String> {
             model
         ));
     }
+    // Reject whitespace in provider or model
+    if parts[0] != parts[0].trim() || parts[1] != parts[1].trim() {
+        return Err(format!(
+            "Invalid model format '{}': provider and model must not contain leading/trailing whitespace",
+            model
+        ));
+    }
 
     // Validate provider is supported
     use std::str::FromStr;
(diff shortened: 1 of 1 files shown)


btucker:
✅ Addressed in f62fb4a — added whitespace rejection to `validate_model_format()` with tests for leading/trailing whitespace in both provider and model parts.

Regarding the channel notes from york:
- **Multi-slash asymmetry** (`parse_task_model` uses `split_once('/')` vs `validate_model_format` uses `split('/')`) — this is a pre-existing gap, not introduced by this PR. Filed as follow-up.
- **Commit workflow** — the test-first workflow is demonstrated in this commit: test added alongside the validation fix with a failing→passing cycle.

PR #1014, opened 2026-02-12
fix: correct mobile autocomplete positioning [Midtown !1186]

btucker:
## Summary
- Fixed autocomplete dropdown positioning on mobile PWAs
- Removed incorrect safe-area-inset-bottom adjustment that caused dropdown to appear too high

## Problem
On mobile PWAs, the autocomplete dropdown was positioned much higher than the text input, making it difficult to use. The issue was in `calculateAutocompletePosition()` which incorrectly subtracted the safe-area-inset-bottom value from the dropdown's top position.

## Solution
`getBoundingClientRect()` already returns viewport-relative coordinates that account for all layout, including safe areas. The safe-area-inset-bottom only affects the input area's padding (to keep the textarea above the home indicator), not the dropdown's absolute positioning.

The fix simplifies the positioning logic by removing the unnecessary safe-area adjustment.

## Test plan
- [x] Build passes
- [ ] Manual testing on mobile PWA (requires physical device)

**Note**: Screenshots require physical mobile device testing. The fix addresses the root cause identified in the code — the incorrect subtraction of safe-area-inset-bottom from an already-correct viewport coordinate.

### DIFF
--- a/web-app/src/lib/Channel.svelte
+++ b/web-app/src/lib/Channel.svelte
@@ -112,23 +112,12 @@
     // The dropdown will use transform: translateY(-100%) to shift up by its own height
     const gap = 8
 
-    // On mobile, account for safe-area-inset-bottom which affects the input area positioning.
-    // We need to extract the computed safe-area value to adjust the dropdown position.
-    // Use getComputedStyle to read the actual resolved padding-bottom value.
-    const inputArea = textareaElement.closest('.input-area')
-    let safeAreaBottom = 0
-    if (inputArea) {
-      const computedPadding = window.getComputedStyle(inputArea).paddingBottom
-      const inputAreaPaddingBottom = parseFloat(computedPadding)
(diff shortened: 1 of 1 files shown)


btucker:
### Code review

No issues found above threshold. Checked for bugs, CLAUDE.md compliance, git history context, previous PR comments, and code comment consistency.

**Task description verification**: The PR addresses the core task requirement (fix mobile autocomplete positioning). The task description also mentions including before/after screenshots for visual UI changes, which the PR acknowledges but does not include due to requiring physical mobile device testing.

**Notes on the approach**: The fix correctly identifies that `getBoundingClientRect()` returns viewport-relative coordinates that already account for layout, making the previous safe-area-inset-bottom subtraction incorrect. The new `transform: translateY(-100%)` CSS approach cleanly separates the "position at textarea top" concern from the "shift up by dropdown height" concern.

PR #1037, opened 2026-02-12
fix: Add defense-in-depth pr_author_sessions cleanup to CleanupMergedWorktree [Midtown !1192]

btucker:
## Summary
- Adds pr_author_sessions cleanup to CleanupMergedWorktree effect as defense-in-depth
- While CompleteTask already cleans up pr_author_sessions (PR #1016), this ensures cleanup happens even if CompleteTask fails or is skipped
- Prevents stale pr_author_sessions from persisting after PR merges

## Changes
1. Add pr_author_sessions.remove() in CleanupMergedWorktree handler
2. Add test coverage for the cleanup behavior

## Test plan
- [x] Unit test passes: `test_cleanup_merged_worktree_removes_pr_author_session`
- [x] Existing test still passes: `test_complete_task_cleans_up_pr_author_sessions`
- [x] Clippy clean

### DIFF
--- a/src/daemon/effects.rs
+++ b/src/daemon/effects.rs
@@ -1101,8 +1101,9 @@ pub async fn execute_effects(effects: Vec<Effect>, state: &DaemonState) {
                     let mut ps = state.persistent_state.lock().await;
                     let removed = ps.worktree_registry.cleanup_for_merged_pr(pr_number);
                     // Also clean up pr_author_sessions for this PR (defense-in-depth)
-                    ps.github.pr_author_sessions.remove(&pr_number);
-                    if removed.is_some()
+                    let pr_session_removed = ps.github.pr_author_sessions.remove(&pr_number);
+                    // Save if either worktree or pr_author_session was removed
+                    if (removed.is_some() || pr_session_removed.is_some())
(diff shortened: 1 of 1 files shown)


btucker:
### Code review

Found 1 issue:

1. **Conditional save defeats defense-in-depth purpose.** The `pr_author_sessions.remove(&pr_number)` call is unconditional, but `save_for_repo()` only runs when `removed.is_some()` (i.e., when the worktree registry had an entry to clean up). If the worktree was already removed but a stale `pr_author_sessions` entry persists, the in-memory removal will not be saved to disk. On daemon restart, the stale entry reappears — defeating the stated "defense-in-depth" goal. The `CompleteTask` handler uses an unconditional save pattern; this should be consistent. Fix: ensure `save_for_repo()` runs when either the worktree or the pr_author_session was removed.

   Test suggestion: Add a test where the worktree registry has no entry for the PR (cleanup_for_merged_pr returns None) but pr_author_sessions has a stale entry. Execute the CleanupMergedWorktree effect, then reload PersistentState from disk and assert the pr_author_sessions entry is gone.

https://github.com/btucker/midtown/blob/1f83d20018f7dcb6fe2a04705b27f14bc88905e7/src/daemon/effects.rs#L1101-L1109

PR #1038, opened 2026-02-12
fix: Mobile web UI autocomplete positioning [Midtown !1186]

btucker:
## Summary
- Fixed autocomplete dropdown positioning on mobile PWAs when the virtual keyboard is open
- Used Visual Viewport API to account for keyboard offset
- Dropdown now appears directly above the textarea instead of higher in the viewport

## Technical Details

The issue occurred because:
1. On mobile, the virtual keyboard reduces the visual viewport height
2. `getBoundingClientRect()` returns coordinates relative to the layout viewport (the full page)
3. The dropdown was positioned using these coordinates without accounting for the visual viewport offset
4. This caused the dropdown to appear much higher than the textarea

The fix:
- Uses `window.visualViewport.offsetTop` to get the keyboard's offset
- Subtracts this offset from the textarea's `rect.top` position
- The Autocomplete component's `transform: translateY(calc(-100% - 8px))` then correctly shifts the dropdown up to appear above the textarea

## Testing
- Build succeeds with no errors
- All unit tests pass (79 channel-related tests)

## Screenshots
Note: This fix requires testing on a mobile device with the PWA installed. Screenshots would require access to a mobile device, which was not available during development. Testing should verify that when typing `@`, `!`, or `#` in the message input on mobile, the autocomplete dropdown appears directly above the keyboard/input, not higher in the viewport.

### DIFF
--- a/web-app/src/lib/Channel.svelte
+++ b/web-app/src/lib/Channel.svelte
@@ -109,17 +109,15 @@
 
     const rect = textareaElement.getBoundingClientRect()
 
-    // Position the dropdown at the top of the textarea.
-    // The dropdown's CSS uses transform: translateY(calc(-100% - 8px)) to:
-    // 1. Shift up by its own height (-100%)
-    // 2. Add an 8px gap (-8px)
-    // This places the dropdown just above the textarea with proper spacing.
+    // Mobile keyboard positioning fix:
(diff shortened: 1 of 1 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1043, opened 2026-02-12
fix: Use rect.bottom for mobile autocomplete positioning [Midtown !1186]

btucker:
## Summary
- Fixed mobile PWA autocomplete dropdown positioning to appear directly above the text input
- Replaced visualViewport.offsetTop calculations with simpler rect.bottom approach
- Dropdown now correctly positions on both desktop and mobile regardless of keyboard state

## Technical Details

**The Problem:**
On mobile PWAs, the autocomplete dropdown appeared much higher in the viewport than intended when the virtual keyboard was open. Previous fixes (PR #1014, #1017, #1038, #1029, #1031) attempted various approaches using visualViewport.offsetTop, but these had coordinate system complexities.

**The Root Cause:**
- `getBoundingClientRect()` returns coordinates relative to the layout viewport
- `position: fixed` elements render relative to the visual viewport
- When the mobile keyboard opens, these two viewports differ
- Converting between viewport coordinate systems is error-prone

**The Fix:**
Use `rect.bottom` (the bottom edge of the textarea) as the positioning anchor instead of `rect.top`. The Autocomplete component's `transform: translateY(calc(-100% - 8px))` then:
1. Shifts up by the dropdown's own height (-100%)
2. Adds an 8px gap

This positions the dropdown directly above the textarea without needing visualViewport calculations, and works correctly on both desktop and mobile.

## Testing
- ✅ Web app builds successfully
- ✅ All 90 unit tests pass (4 test suites)
- ✅ No clippy or formatting warnings

## Note on Mobile Testing
This fix requires verification on a physical mobile device or simulator with the PWA installed. The positioning logic should now work correctly, but visual confirmation is recommended before merging.

## Sources
Research on visualViewport API and mobile positioning:
- [VisualViewport - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/VisualViewport)
- [iOS 26 Safari & WebView: VisualViewport.offsetTop issues](https://developer.apple.com/forums/thread/800125)
- [How to make fixed elements respect virtual keyboard on iOS](https://saricden.com/how-to-make-fixed-elements-respect-the-virtual-keyboard-on-ios)


btucker:
## Review Status

---
> [!NOTE]

btucker:
Rebased onto main and resolved conflicts. The fix now uses `rect.bottom` as the positioning anchor, which is cleaner than the previous visualViewport offset approaches.

Regarding screenshots: I'm running in an isolated worktree environment without access to mobile devices or simulators to capture before/after screenshots. The code change is straightforward and CI is green. 

If screenshots are required before merge, this would need to be tested on a physical mobile device or simulator with the PWA installed, and screenshots added during that testing phase.

PR #1050, opened 2026-02-12
fix: Prevent sandbox-exec nesting to avoid coworker crash loop [Midtown !1196]

btucker:
## Summary

Fixes coworker crash loop caused by sandbox-exec nesting on macOS. After the container sandbox PR merged, all coworkers started crashing instantly because midtown was already running inside a sandbox and tried to apply another sandbox to child processes.

**Changes:**
- Added `can_sandbox()` detection function that tests if sandbox-exec can be applied
- Modified `sandbox_exec_prefix()` to return an error if nesting is detected
- Updated 5 sandbox unit tests to skip when already sandboxed

## Problem

`sandbox-exec` on macOS cannot nest — a process inside a sandbox cannot apply a new sandbox profile to child processes. When midtown runs inside a container (which uses sandbox-exec), spawning coworkers failed because each coworker tried to apply its own sandbox.

## Solution

The `can_sandbox()` function attempts to run a trivial sandboxed command (`/usr/bin/true` with a minimal profile). If we're already sandboxed, this fails with EPERM. The result is cached in a `OnceLock` so the check only runs once per process.

When `sandbox_exec_prefix()` detects nesting, it returns an error instead of proceeding. This allows the caller to handle the situation gracefully (e.g., spawn without sandboxing).

## Test Plan

- [x] `cargo test --lib sandbox` — all 17 tests pass
- [x] 5 sandbox-exec tests correctly skip when already sandboxed
- [x] Verified fix prevents crash loop in containerized environment

## Related

This fix is essential for the container sandbox feature (#1187). Without it, coworkers cannot spawn when midtown runs inside a sandbox.

### DIFF
--- a/src/sandbox.rs
+++ b/src/sandbox.rs
@@ -368,6 +368,10 @@ mod tests {
 
     #[test]
     fn test_sandbox_exec_prefix() {
+        if !can_sandbox() {
+            eprintln!("Skipping test: already inside a sandbox (nesting not allowed)");
+            return;
+        }
         let writable = vec!["/tmp".to_string()];
         let result = sandbox_exec_prefix(&writable);
         assert!(result.is_ok());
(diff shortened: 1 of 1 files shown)


btucker:
Closing as duplicate — this fix was already merged in PR #1036.

PR #1071, opened 2026-02-12
Fix auto-archive channels when all tasks complete [Midtown !1217]

btucker:
## Summary
- Fixed ArchiveChannel and MergeChannels effects to use correct file paths for topic channels
- Topic channels live in `channels/{name}.jsonl`, not `{name}.jsonl` at base_dir
- Both effects now use `Channel::archive()` method which handles paths correctly
- Fixed `snap.tasks` → `snap.all_tasks` in events.rs (field name changed)
- Fixed mock_task() test helpers to use `created_at` instead of removed `model` field
- Added integration tests for channel archiving functionality

## Context
The auto-archive feature was already fully implemented in `src/daemon/auto_archive.rs` and wired into `TaskDispatchTick`, but the effects that actually rename the files had incorrect paths. This meant archive effects were being generated but failing silently when trying to rename non-existent files.

## Test plan
- [x] Unit tests pass: `cargo test auto_archive` (7 tests)
- [x] Integration tests pass: `cargo test --test channel_archive_test` (3 tests)
- [x] Full test suite passes: 1250 tests (only 1 unrelated sandbox test failure)
- [x] Verified Channel::list() excludes .archived.jsonl files
- [x] Verified archive creates files in channels/ subdirectory with .archived.jsonl suffix


btucker:
All three issues fixed in commit a71806f:

1. ✅ **Repeated auto-archive data loss**: Added `archived_channels: HashSet<String>` to `WorldSnapshot` (collected from disk once per tick via `Channel::list_archived()`). Decision function now skips channels in this set. The effect handler's idempotency guard provides defense-in-depth.

2. ✅ **Duplicate tests**: Removed inline tests from `auto_archive.rs`, now using `#[path = "auto_archive_tests.rs"]` pattern per CLAUDE.md.

3. ✅ **Unused `_repo_name` parameter**: Removed from function signature and all call sites.

Added comprehensive idempotency tests (`test_already_archived_channel_skipped`, `test_second_call_with_archived_produces_no_effects`) that verify the decision function emits no effects for already-archived channels.

All 1252 tests pass. Clippy clean.

PR #1081, opened 2026-02-12
fix: Clean up stale branches before creating review worktrees

btucker:
## Summary
- Fixed reviewer worktree creation failure when stale branches exist
- Added proactive stale branch detection and cleanup
- Added unit tests reproducing the bug

## Root Cause
When the daemon tries to spawn a reviewer for a PR, it creates a worktree with a branch named `review-pr-<number>`. If a previous attempt left a stale branch (from a deleted worktree or worktree at a different path), `git worktree add` fails with "fatal: a branch named review-pr-XXXX already exists" and the daemon loops retrying every 30s indefinitely.

## Fix
Before creating a worktree, the code now:
1. Checks if the target branch already exists
2. If yes, prunes stale worktree references
3. Force-deletes the stale branch
4. Retries worktree creation

Added a secondary fallback that handles cases where git discovers the branch conflict during the initial creation attempt.

## Test Plan
- Added unit tests that reproduce the bug by manually deleting a worktree directory (simulating external deletion or crash)
- Verified that the second creation attempt succeeds
- All existing tests still pass


btucker:
Addressing both review items from park:

1. **Silent branch deletion** — ✅ Fixed in 04195fc. Added proper `match` on the `git branch -D` output with `tracing::warn!` logging, matching the reactive path's error handling pattern.

2. **Untested reactive fallback** — Acknowledged as defensive race condition code. The reactive path exists for resilience when the proactive check fails due to a branch appearing between `rev-parse --verify` and `worktree add -b`. Testing requires race condition engineering; leaving as-is per lead's guidance.

PR #1085, opened 2026-02-12
Fix exec-restart RPC response deserialization [Midtown !1228]

btucker:
## Summary
- Fixed RPC response format mismatch between daemon and CLI
- Changed `{"status": "..."}` to `{"message": "..."}` for 4 RPC handlers
- Added comprehensive tests to prevent regression

## Problem
When running `midtown restart`, the exec-restart RPC call was failing with:
```
Warning: exec-restart RPC failed (Response parse error: data did not match 
any variant of untagged enum Response)
```

The daemon would then fall back to the legacy stop+start path, which worked but lost the benefit of exec-restart (preserving unsandboxed process context).

## Root Cause
The daemon's RPC handlers for `exec-restart`, `enter-drain`, `check-pending`, and `shutdown` were returning `{"status": "..."}`, but the CLI's `Response` enum doesn't have a variant that matches that structure. The enum expects `{"message": "..."}` for the `Message` variant.

## Solution
Changed four RPC response handlers in `src/daemon/rpc.rs` from `{"status": "..."}` to `{"message": "..."}`:

| Handler | Before | After |
|---------|--------|-------|
| daemon.exec-restart | `{"status": "restarting"}` | `{"message": "restarting"}` |
| daemon.enter-drain | `{"status": "draining"}` | `{"message": "draining"}` |
| daemon.check-pending | `{"status": "ok"}` | `{"message": "ok"}` |
| shutdown | `{"status": "shutting_down"}` | `{"message": "shutting_down"}` |

## Testing
Added `src/daemon/rpc_tests.rs` with 4 tests that simulate the full RPC serialization/deserialization cycle:
- `test_exec_restart_response_deserializes`
- `test_enter_drain_response_deserializes`
- `test_check_pending_response_deserializes`
- `test_shutdown_response_deserializes`

All tests confirm the responses now deserialize correctly as `Response::Message` variants.

## Test Plan
- [x] Unit tests pass (`cargo test rpc_tests::`)
- [x] Full test suite passes (`cargo test --lib`)
- [x] Manually tested: `midtown restart` should no longer show the warning


btucker:
## Review Status

---
> [\!NOTE]

btucker:
Addressed in 9cac944 — renamed module from `rpc_tests` to `tests` and consolidated all RPC tests (serialization + cache) into the single external file following CLAUDE.md convention.

PR #1117, opened 2026-02-14
docs: Update Lead Maintenance to use cargo install [Midtown !1275]

btucker:
## Summary
- Updated Lead Maintenance section to use `cargo install --path .` instead of `cargo build --release`
- Updated channel message to reflect "installed updated binary" instead of "rebuilt release"

This change simplifies the local installation workflow by using Cargo's built-in installation mechanism, which automatically places the binary in `~/.cargo/bin/` (typically already in PATH), eliminating the need for manual symlinks.

## Test plan
- [x] Documentation changes only - no code changes
- [x] Verified `cargo install --path .` works correctly for local installation

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -181,7 +181,7 @@ Each concern has a primary owner. The non-owner path only acts as reconciliation
 
 ## Lead Maintenance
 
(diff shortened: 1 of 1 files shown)


btucker:
👍 Addressing this now...

PR #861, opened 2026-02-09
perf: Reduce PR poll GraphQL cost by removing comments field [Midtown !1040]

btucker:
## Summary
- Removed `comments` and `author` fields from the bulk PR poll (`gh pr list`) which runs every 30s for ALL open PRs
- Added `fetch_pr_comments()` function to fetch comments on-demand only for PRs that need review comment detection
- Updated `collect_comment_notification_effects()` to fetch comments on-demand before counting

## Cost Reduction
Previously, the bulk poll fetched `comments { first:100 { ... } }` for every open PR every 30 seconds. This is expensive in GraphQL points, especially with nested pagination.

Now, comments are only fetched for PRs owned by coworkers or the lead when checking for new review comments. This dramatically reduces GraphQL API usage while maintaining the same functionality.

## Test Plan
- [x] All unit tests pass (31 PR module tests)
- [x] Clippy passes with `-D warnings`
- [x] Code formatting verified
- [x] Verified both call sites (lead branches and coworker branches) properly fetch comments before usage

🌃 Co-built with [Midtown](https://github.com/btucker/midtown)

PR #1124, opened 2026-02-14
fix: Enable keyboard enhancement flags for Shift+Enter in TUI [Midtown !1284]

btucker:
## Summary
- Enables keyboard enhancement flags in TUI chat to properly detect Shift+Enter
- Fixes bug where Shift+Enter inserted 'j' instead of a newline
- Adds tests documenting expected keyboard protocol behavior

## Problem
The TUI chat input was not properly detecting Shift+Enter. Pressing Shift+Enter would insert a lowercase 'j' instead of a newline. The issue occurred because:
- Modern terminals send escape sequences for Shift+Enter
- Crossterm cannot decode these escape sequences without keyboard enhancement flags
- The TUI was not enabling the kitty keyboard protocol

## Solution
Enable `PushKeyboardEnhancementFlags` during TUI initialization with:
- `DISAMBIGUATE_ESCAPE_CODES` - Removes ambiguity for special keys
- `REPORT_ALL_KEYS_AS_ESCAPE_CODES` - Enables reading modifier keys

Add `PopKeyboardEnhancementFlags` during cleanup to restore terminal state.

## Testing
- All 224 existing tests pass
- New tests document expected keyboard protocol behavior
- Existing behavioral tests confirm event handling logic was always correct
- Clippy clean with `-D warnings`
- Manual testing required in a terminal supporting the kitty keyboard protocol

## Terminal Compatibility
Works in terminals that support the kitty keyboard protocol (iTerm2, kitty, etc.). The flags are harmless in older terminals that don't support the protocol.

## References
- [Crossterm KeyboardEnhancementFlags docs](https://docs.rs/crossterm/latest/crossterm/event/struct.KeyboardEnhancementFlags.html)
- [Crossterm Issue #861: Mac doesn't report Shift or Ctrl with enter key](https://github.com/crossterm-rs/crossterm/issues/861)

### DIFF
--- a/src/bin/midtown/cli/chat/keyboard_protocol_tests.rs
+++ b/src/bin/midtown/cli/chat/keyboard_protocol_tests.rs
@@ -1,66 +1,32 @@
-//! Tests for keyboard protocol initialization
+//! Tests for keyboard protocol configuration.
 //!
-//! Bug #1284: Shift+Enter currently inserts 'j' instead of newline because
-//! the terminal is sending escape sequences that crossterm can't decode
-//! without keyboard enhancement flags.
+//! Verifies that the KEYBOARD_ENHANCEMENT_FLAGS constant includes the
+//! required flags for proper Shift+Enter detection via the kitty protocol.
 
-#[cfg(test)]
-mod tests {
-    use crossterm::event::{
-        KeyCode, KeyEvent, KeyEventKind, KeyEventState, KeyModifiers, KeyboardEnhancementFlags,
-    };
+use crossterm::event::KeyboardEnhancementFlags;
 
-    /// This test demonstrates the expected behavior when keyboard enhancement
-    /// flags are properly enabled.
-    ///
-    /// Without keyboard enhancement flags:
-    /// - Terminal sends ambiguous escape sequences for Shift+Enter
-    /// - Crossterm may decode these as 'j' or other characters
-    ///
-    /// With keyboard enhancement flags:
-    /// - Terminal sends unambiguous escape sequences
-    /// - Crossterm correctly decodes as KeyCode::Enter with SHIFT modifier
-    #[test]
-    fn test_shift_enter_with_enhancement_flags() {
-        // When keyboard enhancement flags are enabled, crossterm should
-        // properly decode Shift+Enter as Enter with SHIFT modifier
-        let expected_event = KeyEvent {
-            code: KeyCode::Enter,
-            modifiers: KeyModifiers::SHIFT,
-            kind: KeyEventKind::Press,
-            state: KeyEventState::NONE,
-        };
+/// Verify that the enhancement flags constant includes DISAMBIGUATE_ESCAPE_CODES.
+///
+/// Without this flag, terminals send ambiguous escape sequences for special
+/// keys and crossterm may decode them incorrectly (e.g., Shift+Enter as 'j').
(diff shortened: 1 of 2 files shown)


btucker:
## Code Review by pleasant

PR #1124: fix: Enable keyboard enhancement flags for Shift+Enter in TUI

### Task Verification

Task !1284 requires:
1. Shift+Enter inserts a newline instead of submitting — **addressed** (this was the root cause fix: enabling kitty keyboard protocol)
2. Input field expands height for multiple lines — **already implemented** (`calculate_input_bar_height` in `ui/chat.rs` handles this)
3. Enter alone still submits — **preserved** (no change to existing Enter handling)

All task requirements are satisfied.

### Issues Found

#### 1. Tests are tautological and add no meaningful coverage (Score: 45)

**File:** `src/bin/midtown/cli/chat/keyboard_protocol_tests.rs` (entire new file)

The new test file constructs crossterm data structures and asserts properties on them:

```rust
let expected_event = KeyEvent {
    code: KeyCode::Enter,
    modifiers: KeyModifiers::SHIFT,
    ...
};
assert!(expected_event.modifiers.contains(KeyModifiers::SHIFT));
```

This is equivalent to `let x = 5; assert_eq!(x, 5);` — it tests that crossterm's structs contain the values that were just assigned. These tests:
- Don't verify the TUI initializes the keyboard protocol correctly
- Don't verify that events are decoded correctly after protocol initialization
- Don't test any midtown code at all

The existing tests in `autocomplete_tests.rs` (`test_shift_enter_inserts_newline`, `test_shift_enter_multi_line`, `test_shift_enter_with_autocomplete_shown`) already cover the event handling logic. The actual bug was that the terminal never sent the right events — not that the handler was wrong.

**Suggestion:** Either remove this test file (the existing autocomplete tests are sufficient for the handler logic, and the protocol initialization is inherently an integration concern), or replace with a test that at least verifies the enhancement flags constant matches what's used in `run()` — something that could break if someone changes one but not the other.

Test suggestion: A unit test that imports the same `enhancement_flags` constant used in `run()` and asserts it includes `DISAMBIGUATE_ESCAPE_CODES | REPORT_ALL_KEYS_AS_ESCAPE_CODES`, so flag changes are caught. Or an integration test that spawns a PTY, enables the protocol, sends a Shift+Enter sequence, and verifies the decoded event.

### Summary

The core fix (enabling `PushKeyboardEnhancementFlags` / `PopKeyboardEnhancementFlags` in the terminal setup/teardown) is correct and well-structured. The enhancement flags are properly paired (push on init, pop on cleanup), and the cleanup path uses `let _` which tolerates errors on exit. On Unix terminals that don't support the kitty protocol, the escape sequence is silently ignored, so there's no compatibility regression.

The only issue is the test file, which adds no real coverage. The fix itself is sound — the problem was always at the terminal protocol level, not the event handling level, and the existing tests already covered the handler.

btucker:
Addressed in a0dff47:
1. Extracted enhancement flags into `KEYBOARD_ENHANCEMENT_FLAGS` module-level constant
2. Replaced tautological tests with two tests that verify the constant includes the required flags (`DISAMBIGUATE_ESCAPE_CODES` and `REPORT_ALL_KEYS_AS_ESCAPE_CODES`)
3. `run()` now uses the constant instead of constructing flags inline

PR #1125, opened 2026-02-14
fix: Support Shift+Enter for multiline input [Midtown !1284]

btucker:
## Summary
Fixes bug where Shift+Enter inserted 'j' instead of newline in modern terminals that support the kitty keyboard protocol.

The root cause was unconditionally pushing keyboard enhancement flags without checking terminal support. This PR conditionally enables the kitty keyboard protocol only when supported, allowing proper Shift+Enter detection in compatible terminals.

## Changes
- Import `supports_keyboard_enhancement` from crossterm::terminal
- Check keyboard enhancement support before pushing enhancement flags
- Only enable kitty protocol flags if terminal supports them
- Conditionally pop enhancement flags during cleanup
- Gracefully default to false on terminal query errors (timeout, piped stdin)
- Add comprehensive test suite for Shift+Enter functionality

## How it works
The existing handler for Shift+Enter (KeyCode::Enter with SHIFT modifier at line 465) remains unchanged and works correctly when crossterm properly decodes the key event. The fix ensures crossterm uses the kitty keyboard protocol (when available) to remove ambiguity from modifier key combinations like Shift+Enter.

## Test plan
- [x] All existing tests pass (229 tests)
- [x] New test suite for Shift+Enter (9 tests)

Testing Shift+Enter:
1. In terminals WITH keyboard enhancement support (kitty, WezTerm, foot, iTerm2 with recent versions):
   - Shift+Enter should insert newline
   - Input bar should expand to show multiple lines (up to 6 lines max)
   - Regular Enter should still send the message

2. In terminals WITHOUT keyboard enhancement support (older terminals, tmux without passthrough):
   - Behavior is unchanged from before this PR
   - Shift+Enter will still be decoded as 'j' due to escape sequence ambiguity
   - Regular Enter sends the message

Note: The 'j' insertion bug remains in legacy terminals that don't support the kitty keyboard protocol. A complete fix would require heuristic detection of Shift+Enter in legacy mode, which is out of scope for this PR.

### DIFF
--- a/src/bin/midtown/cli/chat/mod.rs
+++ b/src/bin/midtown/cli/chat/mod.rs
@@ -22,7 +22,10 @@ use crossterm::{
     },
     execute,
     style::{Color as CrosstermColor, Print, ResetColor, SetForegroundColor},
-    terminal::{EnterAlternateScreen, LeaveAlternateScreen, disable_raw_mode, enable_raw_mode},
+    terminal::{
+        EnterAlternateScreen, LeaveAlternateScreen, disable_raw_mode, enable_raw_mode,
+        supports_keyboard_enhancement,
+    },
 };
 use futures::StreamExt;
 use ratatui::{Terminal, prelude::CrosstermBackend};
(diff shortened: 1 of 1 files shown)


btucker:
This PR is a duplicate of #1124 which was already merged for task !1284. Closing as the work is complete.

PR #1129, opened 2026-02-14
feat: Phase 3 — Plugin worker communication, state management, and sidebar UI [Midtown !1264]

btucker:
## Summary

Implements Phase 3 of the Zellij migration plan (Tasks 9-12):

- **Task 9 — DaemonWorker for RPC**: Uses Zellij's `run_command()` host API (not `std::process::Command` which doesn't work in WASM) to shell out to `midtown plugin` CLI commands for dashboard, attach, detach, and coworker-stream operations
- **Task 10 — Plugin state management**: `PluginState` struct with view modes (Main, CoworkerStream, CoworkerAttached), section-based keyboard navigation across tasks and coworkers, scroll state for stream view, connection tracking with disconnect threshold
- **Task 11 — Sidebar rendering**: ANSI-styled sidebar with task list (status icons, owner, PR info), coworker list (status icons with phase colors, usage limit/error indicators), lead nudge display, and contextual footer help
- **Task 12 — Coworker stream view**: Scrollable read-only stream of coworker events with timestamp formatting and event-type coloring, polled via `plugin.coworker-stream` RPC

Also adds `midtown plugin` CLI subcommands (dashboard, attach, detach, coworker-stream) that bridge between the WASM plugin's `run_command()` calls and the daemon's `plugin.*` RPC endpoints.

## Test plan

- [x] All 10 plugin unit tests pass (state management + rendering)
- [x] All 229 lib/bin tests pass
- [x] `cargo fmt --check` clean
- [x] Main crate builds clean

### DIFF
--- a/midtown-zellij-plugin/src/lib.rs
+++ b/midtown-zellij-plugin/src/lib.rs
@@ -202,20 +202,46 @@ impl MidtownPlugin {
 
     /// Handle a command error (non-zero exit code).
     fn handle_command_error(&mut self, cmd: &str, stderr: &str) -> bool {
-        if cmd == CMD_DASHBOARD {
-            self.state.consecutive_failures += 1;
-            if self.state.consecutive_failures >= DISCONNECT_THRESHOLD {
-                self.state.connected = false;
+        match cmd {
+            CMD_DASHBOARD => {
+                self.state.consecutive_failures += 1;
+                if self.state.consecutive_failures >= DISCONNECT_THRESHOLD {
+                    self.state.connected = false;
+                }
+                let msg = if stderr.is_empty() {
+                    "Daemon not reachable".to_string()
+                } else {
+                    format!("Daemon error: {}", stderr.lines().next().unwrap_or(""))
+                };
+                self.state.record_error(msg);
+                true
             }
-            let msg = if stderr.is_empty() {
-                "Daemon not reachable".to_string()
-            } else {
-                format!("Daemon error: {}", stderr.lines().next().unwrap_or(""))
-            };
-            self.state.error = Some(msg);
-            return true;
+            CMD_ATTACH | CMD_DETACH => {
+                let action = if cmd == CMD_ATTACH {
+                    "Attach"
+                } else {
+                    "Detach"
+                };
+                let msg = if stderr.is_empty() {
+                    format!("{} failed", action)
+                } else {
+                    format!("{} failed: {}", action, stderr.lines().next().unwrap_or(""))
+                };
+                self.state.record_error(msg);
+                true
+            }
+            CMD_COWORKER_STREAM => {
+                // Stream errors are non-critical; record but don't disrupt the view
+                let msg = if stderr.is_empty() {
+                    "Stream fetch failed".to_string()
+                } else {
+                    format!("Stream error: {}", stderr.lines().next().unwrap_or(""))
+                };
+                self.state.record_error(msg);
+                true
+            }
+            _ => false,
         }
-        false
     }
 
     /// Handle keyboard input.
--- a/midtown-zellij-plugin/src/render.rs
+++ b/midtown-zellij-plugin/src/render.rs
@@ -277,12 +277,16 @@ fn render_footer(state: &PluginState, _rows: usize, cols: usize) {
     println!("{DIM}{}{RESET}", truncate(&line, cols));
 }
 
-/// Truncate a string to fit within a given width.
+/// Truncate a string to fit within a given number of characters.
+///
+/// Uses character count consistently (not byte length) to avoid
+/// incorrect truncation of multi-byte strings.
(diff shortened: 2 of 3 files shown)


btucker:
## Code Review by park

**PR**: #1129 — Phase 3: Plugin worker communication, state management, and sidebar UI
**Task**: !1264

The task requires implementing plan tasks 9-12 (Phase 3). Based on the PR description:
- **Task 9 (DaemonWorker for RPC)**: ✅ Implemented via `midtown plugin` CLI subcommands bridging WASM `run_command()` to daemon RPC
- **Task 10 (Plugin state management)**: ✅ `PluginState` struct with view modes, section navigation, scroll state, connection tracking
- **Task 11 (Sidebar rendering)**: ✅ ANSI-styled sidebar with tasks, coworkers, nudges, footer help
- **Task 12 (Coworker stream view)**: ✅ Scrollable read-only stream with timestamp formatting and event-type coloring

All four plan tasks appear covered.

---

### Issues Found

#### 1. Silent error swallowing for non-dashboard commands (Score: 55)

**File**: `midtown-zellij-plugin/src/lib.rs` — `handle_command_error()`

`handle_command_error` only handles errors for `CMD_DASHBOARD`. When `CMD_ATTACH`, `CMD_DETACH`, or `CMD_COWORKER_STREAM` fail with a non-zero exit code, the error is silently dropped — no `record_error()` call, no re-render. This means users get no feedback when an attach or detach operation fails.

```rust
fn handle_command_error(&mut self, cmd: &str, stderr: &str) -> bool {
    if cmd == CMD_DASHBOARD {
        // only this branch handles errors
    }
    false  // all other command errors are silently ignored
}
```

**Suggestion**: Extend error handling to at least record and display errors for attach/detach commands, since these are user-initiated actions where feedback is critical.

**Test suggestion**: Unit test that calls `handle_command_error` with `CMD_ATTACH` and a non-empty stderr, then asserts `state.error` is `Some(...)` and the return value is `true`.

---

#### 2. `truncate` uses byte length for comparison but char count for slicing (Score: 50)

**File**: `midtown-zellij-plugin/src/render.rs` — `truncate()`

The guard condition `s.len() <= max_len` uses byte length, but the truncation logic uses `.chars().take(max_len)`. For strings containing multi-byte characters (emoji in task subjects, CJK characters in coworker names), `s.len()` may exceed `max_len` even when the character count is within bounds, causing unnecessary truncation. Conversely, a string with many multi-byte chars could pass the byte-length check but render wider than expected.

```rust
fn truncate(s: &str, max_len: usize) -> String {
    if s.len() <= max_len {  // byte length
        s.to_string()
    } else if max_len <= 3 {
        s.chars().take(max_len).collect()  // char count
    }
    // ...
}
```

**Suggestion**: Use `s.chars().count()` consistently, or better yet, use a display-width library for accurate terminal column counting.

**Test suggestion**: Test with a multi-byte string like `"héllo wörld"` and `max_len=8` — the current implementation may produce incorrect results since `é` and `ö` are 2 bytes but 1 display column.

---

#### 3. Unbounded `stream_scroll_offset` growth (Score: 42)

**File**: `midtown-zellij-plugin/src/state.rs` — `stream_scroll_down()`

The scroll offset increments without bounds. Clamping is deferred to render time, so the internal state can grow indefinitely. When `update_coworker_stream` replaces events (potentially with a smaller list), the offset may be wildly out of range until the next render.

```rust
pub fn stream_scroll_down(&mut self) {
    if !self.stream_events.is_empty() {
        self.stream_scroll_offset += 1;
        // Clamp handled during render
    }
}
```

**Suggestion**: Clamp in `stream_scroll_down()` as well, or at minimum reset/clamp in `update_coworker_stream()`.

**Test suggestion**: Test that after scrolling down 100 times with only 5 events, `stream_scroll_offset` is clamped to a reasonable value (e.g., `events.len() - 1`).

---

### Summary

Clean implementation overall. Good test coverage for state management (6 tests) and rendering (4 truncation tests). The CLI bridge pattern (`midtown plugin` → RPC → daemon) is well-structured. The three issues above are all low-to-moderate severity — Issue 1 is the most impactful since it silently swallows user-facing errors.

btucker:
✅ All three issues addressed in 7366df8:

1. **Silent error swallowing** — `handle_command_error()` now handles `CMD_ATTACH`, `CMD_DETACH`, and `CMD_COWORKER_STREAM` via a match, recording errors and triggering re-renders for all command types.

2. **Truncate byte/char inconsistency** — Guard condition now uses `s.chars().count()` instead of `s.len()`. Added tests for multi-byte strings and emoji.

3. **Unbounded scroll offset** — `stream_scroll_down()` now clamps to `events.len() - 1`. `update_coworker_stream()` also clamps the offset when the event list shrinks (or resets to 0 if empty). Added three new tests covering these cases.

PR #1135, opened 2026-02-14
feat: Phase 5 — Replace midtown start with Zellij and update health checks [Midtown !1266]

btucker:
## Summary

Implement Phase 5 of the Zellij migration plan (Tasks 16-17):

- **Task 16: Replace `midtown start` with Zellij launch**
  - Add `zellij_session_exists()` for Zellij session detection alongside `tmux_session_exists()`
  - Add `zellij_is_available()` to detect Zellij installation
  - When Zellij is available, `handle_start` creates a Zellij session with a KDL layout containing sidebar plugin, chat pane, and lead Claude Code pane
  - When Zellij is not installed, falls back to the existing tmux-based session creation (full backward compatibility)
  - Add `generate_zellij_layout()` to create project-specific KDL layouts with graceful degradation (simpler layout when plugin WASM isn't built yet)
  - Add `write_lead_launcher_script()` to build a shell script that launches the lead Claude Code session with proper env vars, settings, and prompt
  - Update `handle_attach` to try Zellij first, then tmux
  - Session naming (`midtown-{project}`) is shared across both multiplexers

- **Task 17: Update lead health checks for Zellij**
  - `check_and_respawn_lead()` now checks Zellij sessions first, then falls back to tmux window-level checks
  - Under Zellij, pane management is handled by the layout — no window respawn needed
  - Graceful handling when neither multiplexer is available (no panic)
  - `check_lead_typing()` skips tmux pane capture when no tmux session exists (Zellij plugin handles typing natively)
  - Update daemon event loop message from "tmux server" to "terminal multiplexer"

## Test plan

- [x] All 13 health tests pass (including 2 new Zellij-specific tests)
- [x] `cargo clippy -- -D warnings` passes clean
- [x] Full test suite: 1335 passed, 1 pre-existing failure on main (`test_orphan_recovery_marks_task_in_flight`)
- [x] No regressions from these changes
- [ ] Manual test: `midtown start` with Zellij installed creates Zellij session
- [ ] Manual test: `midtown attach` connects to correct multiplexer
- [ ] Manual test: `midtown start` without Zellij falls back to tmux

### DIFF
--- a/src/bin/midtown/cli/daemon.rs
+++ b/src/bin/midtown/cli/daemon.rs
@@ -433,21 +433,9 @@ fn session_exists(session: &str) -> bool {
 }
 
 /// Check if a Zellij session with the given name exists.
+/// Delegates to the shared implementation in `midtown::tmux`.
 fn zellij_session_exists(session: &str) -> bool {
-    let output = Command::new("zellij")
-        .args(["list-sessions", "--no-formatting"])
-        .output();
-
-    match output {
-        Ok(o) if o.status.success() => {
-            let stdout = String::from_utf8_lossy(&o.stdout);
-            // Each line is a session name (possibly with extra info after whitespace)
-            stdout
-                .lines()
-                .any(|line| line.split_whitespace().next() == Some(session))
-        }
-        _ => false,
-    }
+    midtown::tmux::zellij_session_exists(session)
 }
 
 /// Check if a tmux session with the given name exists.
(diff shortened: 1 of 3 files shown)


btucker:
## All Review Feedback Addressed

Pushed `13df21e` addressing all 6 issues plus additional notes:

1. ✅ **handle_stop kills Zellij sessions** — Now tries `zellij kill-session` first, then tmux. Both paths execute in case of mixed environments.
2. ✅ **Health check signals shutdown** — Simplified logic: if neither Zellij nor tmux session is alive, returns `true` to shut down daemon. No more false `return false` paths.
3. ✅ **MIDTOWN_LEAD_COMMAND support** — Zellij launcher script now checks `MIDTOWN_LEAD_COMMAND` env var, matching tmux behavior.
4. ✅ **clear_lead_task_id_map** — Added before worktree creation in Zellij startup path.
5. ✅ **KDL path escaping** — Added `escape_kdl_string()` that escapes backslashes and double quotes.
6. ✅ **Deduplicated session detection** — Extracted `zellij_session_exists()` and `zellij_is_available()` to `tmux.rs` as shared public functions. Both daemon.rs and health.rs now delegate.
7. ✅ **Removed dead `is_zellij_session`** — Deleted, no planned use.
8. ℹ️ **Unrelated `check_for_stale_worktrees` change** — Asked lead about splitting into separate commit.

PR #1158, opened 2026-02-16
feat: TUI always shows channels and adds click interactions [Midtown !1314]

btucker:
## Summary
- Channels now always display in the TUI board sidebar, even when they have no tasks
- Added mouse click support: clicking the board focuses it, clicking the input area focuses it
- Shift+Enter for newlines was already working (no changes needed)

### Always Show Channels
Previously, the board only showed channels that had tasks. This caused the UI to show an empty board when no tasks existed. Now:
- Added `available_channels: Vec<ChannelInfo>` field to App state
- Created `refresh_available_channels()` method to fetch channels from filesystem
- Updated board rendering to display all available channels, not just channels with tasks
- Exported `ChannelInfo` from the library crate for public use

### Mouse Click Interactions
- Added `board_area` and `input_area` fields to track rendered UI regions
- Updated draw functions to store these areas during rendering
- Added mouse click handler for left mouse button that:
  - Focuses the board panel when clicked
  - Focuses the input bar when clicked
  
This provides a more intuitive UX where users can click to switch focus between UI areas.

## Test Plan
- [x] Code compiles without warnings (cargo clippy passes)
- [x] Code is formatted correctly (cargo fmt passes)
- [x] All existing tests pass (1274 tests)
- [ ] Manual testing: Launch TUI with no tasks, verify channels still display
- [ ] Manual testing: Click on board panel, verify it gets focus (yellow border)
- [ ] Manual testing: Click on input area, verify it gets focus (yellow border)
- [ ] Manual testing: Verify Shift+Enter still inserts newlines in input

### DIFF
--- a/src/bin/midtown/cli/chat/app.rs
+++ b/src/bin/midtown/cli/chat/app.rs
@@ -3279,6 +3279,16 @@ pub(super) mod tests {
                     blocked_by: vec![],
                 },
             ],
+            available_channels: vec![
+                midtown::ChannelInfo {
+                    name: "features".to_string(),
+                    is_archived: false,
+                },
+                midtown::ChannelInfo {
+                    name: "midtown".to_string(),
+                    is_archived: false,
+                },
+            ],
             ..test_app()
         };
 


btucker:
## Review Feedback Addressed

Thanks for the thorough review, park! I've addressed the issues:

### Issue #1 (Score 65): Synchronous refresh on every tick

**Fixed:** Added `CHANNELS_REFRESH_INTERVAL` (30s) with interval gating in `refresh()`. The `refresh_available_channels()` call now only executes every 30 seconds instead of every tick.

Note: The original review mentioned an RPC call, but `refresh_available_channels()` actually uses `Channel::list()` from the filesystem. Still worth gating since directory reads don't need to happen at 1Hz.

### Issue #2 (Score 55): Navigation vs render channel source mismatch

**Fixed:** Simplified `build_board_selections()` to use `self.available_channels` directly. Now both navigation and rendering use the same source, and empty channels are included in navigation.

### Issue #3 (Score 45): Missing click-to-attach for tasks

**Deferred:** Created follow-up task !1326. This requires tracking render positions during `draw_board_panel()`, which is a larger refactoring. Enter key already works for attaching to the selected task.

### Refactor Opportunity

Good catch on `attach_coworker_split()` vs `attach_lead_split()` duplication. That's tracked in task !1325 (already created by the lead).

All changes have been pushed and tests pass.

PR #1164, opened 2026-02-16
chore: Remove build-all.sh in favor of cargo install

btucker:
## Summary
- Deleted `scripts/build-all.sh` — it just wrapped `cargo build && cargo install --path .` which is redundant since `cargo install` already does a release build
- Updated AGENTS.md build commands and Lead Maintenance instructions to use `cargo install --path .` directly

## Test plan
- [ ] Verify `cargo install --path .` builds and installs correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1174, opened 2026-02-17
fix: Don't mark lead PRs as orphaned [Midtown !1371]

btucker:
## Summary
- Fixed bug where lead PRs without task IDs couldn't get reviewers assigned
- Added orphan check exemption for lead branches (branch starts with "lead/")
- Added failing test that reproduces the bug, now passing

## Problem
PR #1164 (lead/remove-build-all-script) was stuck for 140+ minutes without a reviewer despite coworker slots being available. The daemon incorrectly marked it as "orphaned" because:

1. No PR number link in worktree registry
2. No task ID in PR title to lookup  
3. Branch name doesn't match any worktree entry

The orphan detection logic assumed any PR without a worktree entry couldn't address review feedback. This is wrong for lead PRs — the lead's main worktree is always available.

## Solution
Modified `collect_reviewer_effects_with_source()` in `src/daemon/pr.rs` to check if a branch is a lead branch (`is_lead_branch(head_ref)`) before marking it as orphaned. Lead PRs are never orphaned since the lead can always address review feedback.

## Test plan
- [x] Added `test_lead_pr_without_task_id_should_not_be_orphaned` test
- [x] Test fails before fix (0 effects returned)
- [x] Test passes after fix (SpawnCoworkerWithCallbacks effect returned)
- [x] Full test suite passes: `cargo test`
- [x] Clippy passes: `cargo clippy -- -D warnings`
- [x] Manual verification: This fix should allow PR #1164 to get a reviewer assigned


btucker:
## Code Review by columbus

### Task Verification

Task !1371 requires: identify why PR #1164 was stuck 140+ minutes without a reviewer despite available coworker slots, write a failing test, and fix the root cause.

- [x] Root cause identified: lead PRs without task IDs in the title were incorrectly marked as orphaned because all three worktree lookup strategies (PR number, task ID from title, branch name) returned `None`
- [x] Failing test added: `test_lead_pr_without_task_id_should_not_be_orphaned`
- [x] Fix implemented: check `is_lead_branch()` before orphan detection
- [x] Captured snapshot included as fixture

All task requirements are met.

### Review

**No issues found.**

The fix is correct and minimal. The change adds an `is_lead_branch(head_ref)` check at the top of the `None` arm in the orphan detection logic within `collect_reviewer_effects_with_source()`. This correctly short-circuits orphan detection for lead PRs, since the lead's main worktree is always available to address review feedback regardless of whether the PR's branch has a dedicated worktree entry.

The refactoring of `true` from a single expression after the `if/else` to explicit `true` returns in each branch is a correct mechanical transformation that preserves the existing behavior for non-lead PRs.

The test is well-structured with clear documentation of the bug scenario and good assertions (checks both that effects are non-empty and that a `SpawnCoworkerWithCallbacks` effect specifically is present).

PR #1177, opened 2026-02-17
feat: Post headless lead output to main channel [Midtown !1376]

btucker:
## Summary
- Added handler to post headless lead agent output to the main channel
- Ensures user has visibility into what the headless lead is doing

## Implementation Details
When the lead runs headless (no TUI attached), its output was only logged at debug level and not visible to the user. This PR routes that output to the main channel:

1. **Event filtering**: In the daemon event loop's `drain_events` handler, we now check for events from the "lead" session
2. **Text extraction**: For each Assistant StreamEvent from the lead, we extract text content from the message content blocks
3. **Channel posting**: Combined text is posted to the main channel using `send_and_broadcast_async()` with sender "lead"

This gives the user the same visibility into headless lead activity as they have for coworker `/me` messages and system notifications.

## Test plan
- [x] cargo build passes
- [x] cargo clippy passes with -D warnings
- [x] Pre-commit hooks (fmt, clippy) pass

## Testing Considerations
This change affects the daemon's event processing loop, which is already covered by existing integration tests. The addition is non-breaking - it only adds visibility for headless lead output that was previously silent.

Manual testing would involve:
1. Running midtown with a headless lead session
2. Observing that lead output appears in the main channel
3. Verifying no output duplication when lead is attached interactively


btucker:
Addressed all three issues in c29fcab:

1. **Channel flooding** (Score 75) — ✅ Extracted `extract_lead_text()` function that aggregates all lead text across the entire drain cycle into a single string, then posts one channel message. No more per-event flooding.

2. **No tests** (Score 40) — ✅ Added 8 unit tests covering: single text block, multiple events aggregated, non-text blocks skipped, empty content arrays, system events ignored, no events, multiple text blocks in one event, and missing content field.

3. **Architecture deviation** (Score 45) — Acknowledged. Keeping the logic in the drain handler for now since it's a simple aggregation (similar pattern to the session_id backfill). The Effect system would be the right path if this grows in complexity.

PR #1194, opened 2026-02-17
fix: Add retry logic to channel read operations [Midtown !1397]

btucker:
## Summary
- Fixed CI test failure: `test_channel_read_with_last_parameter` was failing with exit code 101
- Added bounded retry logic (10 attempts, 50ms delays) to three channel read methods
- Prevents `WouldBlock` errors when reads happen immediately after async writes

## Root Cause
The test `test_channel_read_with_last_parameter` was failing in CI because:
1. `handle_channel_post()` is async and uses `spawn_blocking` to write to the channel file
2. After awaiting the write, the test immediately calls `handle_channel_read()`
3. `read_last_n_messages()` used `try_lock_shared()` with no retries
4. In CI (heavily loaded), the exclusive write lock might not be fully released yet
5. This caused the read to fail with `WouldBlock`

## Fix Details
Added retry logic to three read methods in `src/channel.rs`:
- `read_last_n_messages()` - line 593
- `read_since_cursor()` - line 490
- `read_messages_before_position()` - line 727

Each now retries up to 10 times with 50ms delays (500ms total) when encountering lock contention, matching the existing pattern in `read_all()`.

## Test Plan
- [x] Local tests pass: `cargo test`
- [x] Clippy clean: `cargo clippy -- -D warnings`
- [x] Formatting clean: `cargo fmt -- --check`
- [x] Failing test now passes: `cargo test test_channel_read_with_last_parameter`
- [ ] CI tests pass on GitHub Actions

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -488,9 +488,35 @@ impl Channel {
         let mut cursor = Cursor::load_or_create(&self.base_dir, &self.channel_name, agent)?;
 
         let file = File::open(&self.channel_file)?;
-        // Try to acquire shared lock without blocking
-        file.try_lock_shared()
-            .map_err(|e| std::io::Error::new(std::io::ErrorKind::WouldBlock, e.to_string()))?;
+
+        // Try to acquire shared lock with bounded retries to handle lock contention.
+        let mut acquired = false;
+        for attempt in 0..10 {
+            match file.try_lock_shared() {
+                Ok(()) => {
+                    acquired = true;
+                    break;
+                }
+                Err(_) if attempt < 9 => {
+                    std::thread::sleep(std::time::Duration::from_millis(50));
+                }
+                Err(e) => {
+                    return Err(std::io::Error::new(
+                        std::io::ErrorKind::WouldBlock,
+                        format!("Failed to acquire shared lock after 500ms: {}", e),
+                    )
+                    .into());
+                }
+            }
+        }
+
+        if !acquired {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::TimedOut,
+                "Failed to acquire shared lock after 500ms",
+            )
+            .into());
+        }
 
         let mut reader = BufReader::new(file);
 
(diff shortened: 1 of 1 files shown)


btucker:
## Code Review by park

PR #1194 adds bounded retry logic (10 attempts, 50ms delays) to three channel read methods (`read_since_cursor`, `read_last_n_messages`, `read_messages_before_position`) to prevent `WouldBlock` errors when reads race with async writes. The fix correctly addresses the CI flake in `test_channel_read_with_last_parameter`.

**Task verification**: The task (#1397) asks to fix the CI 'Test' failure on main. The PR identifies the root cause (lock contention between async write and immediate read) and applies retries to the affected code paths. Task requirements are met.

### Issues

#### 1. Unreachable dead code: `if !acquired` block (×3)

**Severity**: Low | **Score**: 45

In all three new retry blocks, the `if !acquired { return Err(...) }` check after the for loop is unreachable dead code. The loop runs attempts 0–9. On attempts 0–8, lock failure sleeps and continues. On attempt 9, the `Err(_) if attempt < 9` guard is false, so control falls to `Err(e)` which returns immediately. Therefore, the loop can only exit via `break` (with `acquired = true`) or via early return — `acquired` is always `true` after the loop.

The existing `read_all()` method avoids this by using a simpler pattern (just `break` on success, no `acquired` variable). The `send()` method has the same dead-code pattern, so this PR is at least consistent with `send()`.

**Suggestion**: Consider using the simpler `read_all()` pattern (no `acquired` boolean, no post-loop check) to eliminate the dead code. Alternatively, extract a shared helper to consolidate all four retry sites.

**Test suggestion**: N/A (dead code — no behavioral impact, would be caught by `#[warn(unreachable_code)]` if the compiler can prove it).

#### 2. Repeated retry logic across 4 methods

**Severity**: Low | **Score**: 42

The retry-with-backoff logic for shared lock acquisition is now copy-pasted in 4 places (`read_all`, `read_since_cursor`, `read_last_n_messages`, `read_messages_before_position`) plus a 5th variant in `send()` for exclusive locks. A small helper like `fn try_lock_shared_with_retry(file: &File) -> Result<()>` would consolidate the pattern and prevent future drift between the copies.

**Test suggestion**: N/A (code organization issue, not a bug).

### Summary

The fix is correct and addresses the CI flake. The retry parameters (10 attempts × 50ms = 500ms max) are reasonable for lock contention from async writes. No functional bugs found. The two issues above are code hygiene items that could be addressed as follow-ups.

PR #1207, opened 2026-02-17
fix: Outdent board sidebar channels and fix broken session tests

btucker:
## Summary
- Move channel names flush against the left edge of the board sidebar box (remove 2-space indent)
- Task bullets get 1-space indent with continuation lines aligned to first letter of subject text
- Coworkers header also outdented to match
- Restore `SESSION_CWD_MUTEX` and `ensure_in_git_repo` definitions dropped when PR #1204 and #1205 merged concurrently
- Fix session tests to pass project root as cwd instead of fake paths (`/tmp/test-cwd`)

## Test plan
- [x] All 14 board panel tests pass
- [x] All 37 session tests pass (were failing on main)

### DIFF
--- a/src/bin/midtown/cli/chat/ui/board.rs
+++ b/src/bin/midtown/cli/chat/ui/board.rs
@@ -187,11 +187,11 @@ fn render_channel_header(
     let task_count = tasks.len();
     let channel_header = if let Some(&unread_count) = app.channel_unread_counts.get(channel_name) {
         format!(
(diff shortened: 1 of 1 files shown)


btucker:
Addressed in 1d03e57. The test now creates a temporary git repo at `/tmp/midtown test dir/repo` (path with spaces) and uses that as the cwd, so the quoting assertions actually exercise the shell-escape logic rather than passing trivially on a plain path.

PR #1209, opened 2026-02-17
feat: Add click-and-drag resizing for sidebar/chat divider [Midtown !1420]

btucker:
## Summary
- Add `sidebar_width_pct` field to `App` state (default 40%, clamped 20–60%) replacing the hardcoded `Constraint::Percentage(40)` split
- Track `divider_x` and `layout_width` each render pass so mouse hit-testing and drag math use accurate coordinates
- Handle `MouseEventKind::Down` on the divider column to start drag, `Drag` to resize, and `Up` to stop

## Implementation details

The divider is detected by checking if the mouse column exactly matches `divider_x` (the last column of the sidebar panel). During a drag, `resize_sidebar_to(mouse_x, layout_width)` computes the percentage and clamps it to [20, 60] so both panels remain usable. The message render cache is invalidated on resize since the available width changes.

## Test plan
- [x] Unit tests for `resize_sidebar_to`: normal case, min clamp, max clamp, zero-width noop, cache invalidation
- [x] Integration tests: click divider starts drag, adjacent click doesn't start drag, full drag sequence (click → drag → release), min/max clamping via drag, drag without prior mousedown is no-op
- [x] All library tests pass (1360 tests)

### DIFF
--- a/src/bin/midtown/cli/chat/app.rs
+++ b/src/bin/midtown/cli/chat/app.rs
@@ -323,6 +323,10 @@ pub struct App {
     pub dragging_divider: bool,
     /// Full width of the horizontal layout area (set each render pass, used for drag resize)
     pub layout_width: u16,
+    /// Y range of the main content area (set each render pass, used for divider Y bounds check)
+    pub main_area_y: u16,
+    /// Bottom Y (exclusive) of the main content area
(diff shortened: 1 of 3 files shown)


btucker:
## Review Status

---
> [!NOTE]

PR #1219, opened 2026-02-17
fix: Make 'midtown view' chat-only by default, lead attach opt-in via --attach [Midtown !1436]

btucker:
## Summary
- Changed `midtown view` default behavior to chat-only — no longer attaches to the headless lead session
- Added `--attach` flag as opt-in for the old behavior (attach + split pane)
- Removed the confusing `--skip-auto-split` flag (negative opt-out replaced with positive opt-in)

## Motivation

Previously, running `midtown view` always attached to the lead session, pausing the headless lead and opening an interactive Claude session in a split pane. This was disruptive when users just wanted to read the chat TUI. Now the default is non-intrusive: just show the chat without touching the lead session.

## Changes

- `src/bin/midtown/main.rs`: Replaced `--skip-auto-split` flag with `--attach` flag in View command
- `src/bin/midtown/cli/mod.rs`: Updated `handle_view()` signature to use `attach` instead of `skip_auto_split`
- `src/bin/midtown/cli/daemon.rs`: Restructured `handle_view()` to only attach when `--attach` is passed; default path skips the session attach/detach/split logic entirely

## Test plan
- [ ] `midtown view` launches chat TUI without attaching the lead session
- [ ] `midtown view --attach` preserves the old behavior (attach + split pane)
- [ ] `--skip-auto-split` flag is removed (no regressions in tests referencing it)

### DIFF
--- a/src/bin/midtown/main.rs
+++ b/src/bin/midtown/main.rs
@@ -512,7 +512,7 @@ fn main() {
         return;
     }
 
-    // View command (launches chat locally and best-effort split for Lead)
+    // View command (launches chat-only by default; --attach opens Lead in a split)
     if let Commands::View { project, attach } = &command {
         let result = cli::handle_view(project.as_deref(), *attach);
         handle_result(format, result);


btucker:
The PR has merge conflicts that need to be resolved before merging. The review feedback has been addressed, but conflicts have appeared since then.

PR #1227, opened 2026-02-17
fix: Kill stale daemon processes and their zombie children on startup

btucker:
## Summary

- After daemon restarts, old daemon processes could linger (losing the PID lock but not exiting), keeping their child lead/coworker processes alive. The zombie scanner only killed PPID=1 orphans, so children of stale daemons escaped cleanup — causing 3 simultaneous lead sessions.
- Add `kill_stale_daemon()` helper that SIGTERMs the old daemon after acquiring the PID lock, with SIGKILL fallback after 3s
- Expand zombie scanner to also kill children of stale midtown daemons (not just PPID=1 orphans)
- Use graceful SIGTERM→wait→SIGKILL in zombie scanner instead of immediate `kill -9`

## Test plan

- [x] `cargo test` — all existing + 6 new tests pass
- [ ] `cargo install --path .` + `midtown restart` — verify single daemon and lead process
- [ ] `ps aux | grep midtown | grep -v grep` — no stale processes

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -85,7 +85,7 @@ pub use pr::{collect_merged_pr_cleanup_effects, reconcile_orphaned_prs};
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::fs::File;
 use std::io::{BufRead, BufReader, Read as _, Seek, SeekFrom, Write};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use fs2::FileExt;
@@ -1768,7 +1768,7 @@ fn validate_github_repo_access(github_user: &str, workdir: &PathBuf) -> crate::R
 /// The lock is held for the lifetime of the returned File handle.
 ///
 /// Returns an error if another daemon is already running (lock already held).
-fn acquire_pid_lock(pid_path: &PathBuf) -> crate::Result<File> {
+fn acquire_pid_lock(pid_path: &PathBuf, workdir: &Path) -> crate::Result<File> {
     // Ensure parent directory exists
     if let Some(parent) = pid_path.parent() {
         std::fs::create_dir_all(parent)?;
@@ -1793,7 +1793,7 @@ fn acquire_pid_lock(pid_path: &PathBuf) -> crate::Result<File> {
             if let Ok(old_pid) = old_contents.trim().parse::<u32>()
                 && old_pid != std::process::id()
             {
-                startup::kill_stale_daemon(old_pid);
+                startup::kill_stale_daemon(old_pid, workdir);
             }
 
             // Write our PID. After read_to_string, the cursor is at EOF.
@@ -2176,7 +2176,7 @@ pub async fn run(config: DaemonConfig) -> crate::Result<DaemonExitStatus> {
     }
 
     // Acquire exclusive lock on PID file to enforce singleton behavior
-    let pid_file = acquire_pid_lock(&config.pid_file_path)?;
+    let pid_file = acquire_pid_lock(&config.pid_file_path, &config.workdir)?;
     info!("Acquired PID lock: {}", config.pid_file_path.display());
 
     // Ensure parent directory exists for socket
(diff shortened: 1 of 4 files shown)


btucker:
## Review: fix: Kill stale daemon processes and their zombie children on startup

**Verdict: Request Changes** — One real bug to fix, otherwise solid.

### Summary

Good fix for a real operational problem (stale daemons keeping zombie children alive after restarts). The approach is sound: read the old PID from the lock file after acquiring it, kill the stale daemon, and expand the zombie scanner to also catch children of stale daemons rather than only PPID=1 orphans. The SIGTERM→wait→SIGKILL escalation pattern is a nice improvement over the old immediate `kill -9`.

### Issues

**Bug — PID file corruption from missing seek** (`src/daemon/mod.rs:1800`)

`file.read_to_string()` advances the cursor to the end of the old content. `set_len(0)` truncates but does **not** reset the cursor. The subsequent `writeln!` writes at the stale offset, producing null bytes followed by the PID (e.g., `\0\0\0\0\0\067890\n`). This breaks stale daemon detection on the **next** restart cycle. Fix: add `file.seek(std::io::SeekFrom::Start(0))?` after `set_len(0)`.

**Nit — Stale comment at call site** (`src/daemon/mod.rs:2352-2354`)

The comment above `kill_zombie_claude_processes()` still says "This only kills truly orphaned (PPID=1) processes" but the function now also kills children of stale daemons. Should be updated to match the new docstring.

### What looks good

- `verify_midtown_process()` guard prevents killing PID-reused processes — important safety check
- Graceful SIGTERM→SIGKILL with configurable timeouts (3s for stale daemon, 2s for zombies)
- Test coverage for the new helpers (6 new tests, all pass)
- Snapshot fixture follows the project's debugging workflow conventions
- Clippy clean, all tests pass

src/daemon/mod.rs
```diff
@@ -2340,7 +2351,7 @@ pub async fn run(config: DaemonConfig) -> crate::Result<DaemonExitStatus> {
     // Kill any zombie Claude headless processes left from crashes or unclean shutdowns.
     // This only kills truly orphaned (PPID=1) processes from crashes — NOT processes
     // that were intentionally detached during a clean daemon restart.
-    startup::kill_zombie_claude_processes();
+    startup::kill_zombie_claude_processes(std::process::id());
```
btucker: **Nit: Comment is now stale.** This comment still says "This only kills truly orphaned (PPID=1) processes from crashes — NOT processes that were intentionally detached during a clean daemon restart." but the behavior was expanded to also kill children of stale midtown daemons. Suggest updating to match the new docstring in `startup.rs`.

btucker (commented):
One bug to fix: missing `file.seek(SeekFrom::Start(0))` after `read_to_string` in `acquire_pid_lock` — causes PID file corruption with null bytes, breaking stale daemon detection on subsequent restarts. See inline comment for details and fix.

Otherwise the approach is solid. The SIGTERM→SIGKILL escalation and `verify_midtown_process` safety guard are well done.

btucker:
Addressed the remaining 4 issues from the riverside review:

**Issue #3 — verify_midtown_process is now project-specific:**
- Added `project_workdir: &Path` parameter to `verify_midtown_process()` and `kill_stale_daemon()`
- Before killing a stale daemon, checks that its cmdline contains the project workdir (daemon is launched with `--workdir <repo>`, so this scopes the check to the right project)
- Prevents accidentally killing a midtown daemon for a different project if the stale PID happens to be reused

**Issue #4 — verify_claude_process called before zombie kill:**
- Added `verify_claude_process(pid)` check before adding each candidate to the zombie kill list
- Guards against PID recycling between the initial `pgrep` scan and the eventual kill call

**Issue #5 — docs/architecture.md updated:**
- Added "Daemon Startup Sequence" section documenting all 6 steps: PID lock acquisition, stale daemon cleanup, session PID collection, zombie scanner, task assignment restore, and session recovery

**Issue #6 — Replace flat sleep with poll loop:**
- Replaced `thread::sleep(2s)` with a 100ms poll loop (up to 2s total) that exits early when all zombie processes die
- Mirrors `kill_stale_daemon`'s responsive wait strategy

All tests pass (1385 unit + integration tests).

PR #1226, opened 2026-02-17
feat: Add lead working indicator line to chat TUI [Midtown !1450]

btucker:
## Summary
- Adds a 1-line status area between the chat messages and input bar showing a braille spinner with "lead..." when the headless lead session is actively working
- Lead status is fetched from the daemon's `kanban.data` RPC response as a live (non-cached) field
- Space is always reserved to prevent layout jitter when the indicator toggles on/off

## Implementation

**Daemon (`rpc_kanban.rs`):** Added `lead_working` to the `kanban.data` RPC response. This field is populated dynamically via `session_manager.is_alive("lead")` — on both cache hits and misses — so it always reflects the current session state rather than the 30s cached data.

**TUI App (`app.rs`):** Added `lead_working: bool` to `App` state and `KanbanData`. Updated `fetch_kanban_data_via_rpc` to parse the new field from the RPC response, and propagated it through `refresh_kanban` → `KanbanData` → `App::lead_working`.

**Chat UI (`chat.rs`):** Added `draw_lead_indicator` function and a `Constraint::Length(1)` slot in `draw_chat_panel`'s vertical layout. The indicator renders `"⠹ lead..."` in `DarkGray` when active, or an empty line when idle. Reuses the existing `app.spinner_char()` animation infrastructure.

## Test plan
- [x] `cargo test --bin midtown` — 326 tests pass
- [x] `cargo clippy -- -D warnings` — no warnings
- [x] `cargo fmt -- --check` — no formatting issues
- [ ] Manual: verify spinner appears when lead is working and disappears when idle

🌃 Co-built with [Midtown](https://github.com/btucker/midtown)

PR #1228, opened 2026-02-17
fix: Address review feedback for lead working indicator [Midtown !1450]

btucker:
## Summary
Follow-up to PR #1226 — addresses review feedback from pleasant.

- **Activity detection**: Replace `is_alive("lead")` with `last_event_at`-based timeout (5s) using `headless_health`. The spinner now shows only when the lead session is actively computing, not just when the process is alive.
- **Cache hygiene**: Strip `lead_working` from the cached kanban response. Live value is injected on every request, preventing stale values if the cache-hit injection is ever refactored away.
- **Autocomplete overlap**: Increase dropdown Y offset by 1 to account for the new indicator row between messages and input bar.
- **Tests**: 5 unit tests for `is_session_actively_working` covering: no health data, not alive, recent event (active), stale event (idle), no events (idle).

Item #4 (shared spinner frame) is cosmetic and deferred — both spinners animating in sync is not functionally wrong.

## Test plan
- [x] `cargo test --lib` — 1380 tests pass (5 new)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 4
 
 [[package]]
+name = "adler2"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
+
+[[package]]
 name = "aead"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


btucker:
Addressing all 3 items:

1. **Inline tests → separate file** — Moving the 5 `is_session_actively_working` tests to `rpc_kanban_tests.rs`. Correct per CLAUDE.md convention.
2. **Future timestamp bug** — Adding `.max(0)` guard. Adding test with future timestamp.
3. **`saturating_sub(2)` comment** — Fixing to accurately describe the layout (lead indicator row + spacing from input bar border).

Also fixing the stale `KanbanCache` doc comment (per vernon's channel note about `lead_working` being excluded from cache).

PR #1229, opened 2026-02-17
feat: Replace drain-and-wait with SIGTERM on midtown restart [Midtown !1455]

btucker:
## Summary
- `midtown restart` now sends SIGTERM to all running coworker sessions immediately instead of polling for up to 5 minutes waiting for them to drain naturally
- Added `SessionManager::graceful_shutdown_all()` — SIGTERMs all sessions, waits up to 10s for clean exit, SIGKILL fallback via Drop
- Added `coworker.stop_all` daemon RPC and `stop_all_coworkers()` client method
- Removed the now-unused `wait_for_coworkers_to_drain` polling function and `enter_drain` client method

## Behavior change

**Before:** `midtown restart` entered drain mode (stopped new task assignments) then polled every 500ms for up to 5 minutes waiting for coworkers to naturally finish. `--force` skipped this wait.

**After:** Both default and `--force` modes immediately send SIGTERM to all coworker sessions via the daemon. The daemon waits up to 10s for clean exits before SIGKILL fallback. Restart is then immediate.

Note: SIGTERM terminates the Claude Code process (which saves session state before exiting), so coworkers can still be resumed after restart.

## Test plan
- [x] `cargo test` passes

### DIFF
--- a/src/bin/midtown/cli/daemon.rs
+++ b/src/bin/midtown/cli/daemon.rs
@@ -1118,7 +1118,7 @@ pub fn handle_restart(_force: bool) -> Result<Response, String> {
     drop(client);
 
     // Send exec-restart RPC to the daemon. The daemon will:
-    // 1. Gracefully shut down (persist sessions, detach coworkers)
+    // 1. Persist session info, then terminate all coworker sessions
     // 2. Re-exec itself with --foreground, preserving its original process
     //    context. This is critical on macOS: the daemon was initially launched
     //    from an unsandboxed CLI, so re-exec preserves that unsandboxed state.
--- a/src/bin/midtown/main.rs
+++ b/src/bin/midtown/main.rs
@@ -100,7 +100,7 @@ enum Commands {
     },
     /// Restart midtown (stop + start)
     Restart {
-        /// Skip graceful drain and force immediate restart
+        /// Reserved for future use (currently no-op)
         #[arg(long)]
         force: bool,
     },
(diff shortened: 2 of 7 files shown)


btucker:
Addressed all 8 feedback items:

**Functional fixes:**
1. **Session persistence (#1)**: `graceful_shutdown_all()` no longer removes sessions from the map. It takes the process handles out (setting `session: None`) but keeps the `CoworkerSession` entries with `status: Stopped`. `collect_session_info()` was also fixed to not require `cs.session` to be `Some` — it now only requires `session_id`.
2. **Spawn-during-shutdown race (#5)**: Set `draining = true` in the `coworker.stop_all` RPC handler before sending SIGTERM. `TaskDispatchTick` will now skip task assignment during the wait window.

**Cleanup:**
3. **Dead `daemon.enter-drain` handler (#2)**: Removed the RPC handler and its orphaned test `test_enter_drain_response_deserializes`.
4. **Stale `--force` help text (#3)**: Updated to "Reserved for future use (currently no-op)".
5. **Stale comment in `handle_restart` (#7)**: Updated to reflect coworkers are terminated, not detached.
6. **Stale `DaemonState.draining` doc comment (#8)**: Updated to reflect it's now set by `coworker.stop_all`.

**Tests (#4):**
Added three unit tests in `sessions_tests.rs`:
- `test_graceful_shutdown_all_empty_returns_zero`
- `test_graceful_shutdown_all_marks_sessions_stopped` — verifies sessions remain in the map with Stopped status
- `test_graceful_shutdown_all_preserves_session_info` — regression test for the persistence bug: verifies `collect_session_info()` returns data after `graceful_shutdown_all()`

Note: Issue #6 (architectural pattern) was noted but not addressed in this PR since the PR author's original choice was intentional and the comment framed it as a pattern preference, not a functional issue.

PR #1231, opened 2026-02-17
fix: Render markdown tables with aligned columns in TUI chat [Midtown !1457]

btucker:
## Summary

- Tables in chat messages now render with properly aligned columns instead of misaligned separators
- Detects table blocks (lines starting with `|`) and routes them through the block-level parser (`minimad_ratatui::from_str`) which uses `compute_table_layout()` to compute column widths across all rows before rendering
- Regular text paragraphs continue to use word-wrapping + inline parsing unchanged
- Removes the now-unused `parse_markdown()` wrapper from `text.rs` (replaced by direct `minimad_ratatui::inline()` calls)

**Root cause:** `render_message()` previously processed content line-by-line using the inline parser, which parsed each row independently and could not know the maximum column width across all table rows. The block parser solves this by scanning all rows in a table block before rendering any of them.

## Test plan

- [x] Added `test_markdown_table_columns_are_aligned` — verifies that the `│` separator appears at the same character position in both the header row ("Name") and data row ("riverside"), which would fail with the per-line inline parser
- [x] All 329 existing tests pass
- [x] `cargo fmt` and `cargo clippy -- -D warnings` clean

### DIFF
--- a/src/bin/midtown/cli/chat/ui/messages.rs
+++ b/src/bin/midtown/cli/chat/ui/messages.rs
@@ -11,7 +11,6 @@ use midtown::MessageType;
 
 use super::TIMESTAMP_GUTTER_WIDTH;
 use super::styles::{get_sender_color, is_dim_sender, is_system_like_sender};
-use super::text::{parse_markdown, wrap_content};
 
 /// Precomputed values shared by message rendering functions.
 ///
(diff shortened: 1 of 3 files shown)


btucker:
## Review Status

---
> [\!NOTE]

btucker:
Addressing review feedback. Initial assessment:

1. **Spurious blank line** — Confirmed. `buf.is_empty()` should check trimmed content. Fixing.
2. **Code-fence awareness** — Confirmed. `|`-prefixed lines inside code blocks get misidentified as tables. Adding `in_fence` tracking.
3. **Doc comment re code fences** — Confirmed. Only tables go through `from_str`, not code fences. Fixing doc.
4. **Mermaid path** — Valid but low impact (mermaid messages unlikely to contain markdown tables). Creating follow-up task.
5. **Stale doc comment** — Fixing the contradictory wording.
6. **Test file placement** — The existing `mod tests` block already has ~90 inline tests. Moving one test to a separate file while leaving the rest inline would be inconsistent. Will create a follow-up task to move all tests at once.

Working on fixes now.

PR #1239, opened 2026-02-17
feat: Add time-based spinner animation tick to TUI chat [Midtown !1475]

btucker:
## Summary

- Replace per-draw-call spinner frame advancement with a dedicated ~100ms animation timer
- Spinner now advances at ~10fps regardless of draw rate, giving a fluid braille spinner
- Animation redraws only occur when `lead_working == true` (spinner is visible)

## Changes

- `app.rs`: Add `spinner_last_tick: Instant` field to `App` struct
- `app.rs`: `spinner_char()` is now read-only (no longer advances frame counter)
- `app.rs`: New `tick_spinner()` method advances frame only if ≥100ms have elapsed since last tick
- `mod.rs`: Add `animation_interval` (100ms tokio timer) to the event loop
- `mod.rs`: Animation tick handler calls `tick_spinner()` only when `lead_working` is true

## Test plan

- [x] All existing unit tests pass (`cargo test`)
- [x] No clippy warnings (`cargo clippy -- -D warnings`)
- [x] Build succeeds

### DIFF
--- a/src/bin/midtown/cli/chat/app.rs
+++ b/src/bin/midtown/cli/chat/app.rs
@@ -1780,16 +1780,21 @@ impl App {
         SPINNER_FRAMES[self.spinner_frame % SPINNER_FRAMES.len()]
     }
 
+    /// Returns true if any spinner is currently visible (lead working or active coworkers).
+    pub fn any_spinner_visible(&self) -> bool {
+        self.lead_working
+            || self
+                .coworkers
+                .iter()
+                .any(|cw| cw.phase.as_deref() != Some("idle") && cw.phase.is_some())
+    }
+
     /// Advance the spinner frame if enough time has elapsed since the last tick.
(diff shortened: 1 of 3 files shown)


btucker:
Addressed all feedback in 7628d64:

1. **Coworker spinner freeze** — Fixed. Replaced `lead_working` gate with `any_spinner_visible()` that checks both lead working state and active coworkers.
2. **Unnecessary redraws when idle** — Fixed. Animation tick now only fires when `any_spinner_visible()` returns true, preserving the 1/sec idle refresh rate.
3. **Return value discarded** — Changed `tick_spinner()` to return `()` since the draw loop doesn't use the signal.
4. **No tests** — Added `spinner_tests.rs` with 7 tests: tick timing behavior, spinner_char consistency, and visibility detection for lead/coworker/idle states.

PR #1246, opened 2026-02-17
feat: Add kill ring append for consecutive kills and Ctrl+Y yank [Midtown !1483]

btucker:
## Summary

- Consecutive kill operations (Ctrl+K, Ctrl+U, Ctrl+W) now append to the kill ring rather than overwriting it, matching standard emacs behavior
- Any non-kill command resets the accumulation so the next kill starts fresh
- Adds Ctrl+Y to yank (paste) the kill ring content at the cursor position
- Adds `last_was_kill` field to `App` to track consecutive kill state

### How kill ring append works

A `prev_was_kill` flag is captured at the start of each key event and immediately reset to `false`. Kill commands check `prev_was_kill` to decide whether to append or replace, then set `app.last_was_kill = true`. Non-kill commands naturally leave `last_was_kill = false` since the reset happens unconditionally.

## Test plan

- [x] `test_consecutive_ctrl_k_appends_to_kill_ring` — consecutive kills accumulate
- [x] `test_non_kill_resets_kill_ring_accumulation` — Ctrl+A between kills resets
- [x] `test_ctrl_y_yanks_kill_ring_at_cursor` — yank inserts at cursor, advances position
- [x] `test_ctrl_y_with_empty_kill_ring_is_no_op` — empty kill ring is safe
- [x] `test_ctrl_y_preserves_text_around_cursor` — yank inserts mid-text correctly
- [x] All existing emacs keybinding tests still pass

### DIFF
--- a/src/bin/midtown/cli/chat/emacs_keybinding_tests.rs
+++ b/src/bin/midtown/cli/chat/emacs_keybinding_tests.rs
@@ -216,8 +216,9 @@ fn test_ctrl_d_dismisses_autocomplete() {
     assert!(!app.autocomplete.show);
 }
 
-// Consecutive kill operations append to the kill ring so that Ctrl+Y can
-// yank all killed text at once (emacs kill ring semantics).
+// Consecutive kill operations append/prepend to the kill ring so that Ctrl+Y
+// can yank all killed text at once (emacs kill ring semantics).
+// Forward kills (Ctrl+K) append; backward kills (Ctrl+U, Ctrl+W) prepend.
 #[test]
 fn test_consecutive_ctrl_k_appends_to_kill_ring() {
     use app::FocusedPane;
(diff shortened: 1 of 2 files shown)


btucker:
All three issues fixed in f0d7a22:

1. **Backward kills prepend** — Ctrl+U and Ctrl+W now use `insert_str(0, &killed)` instead of `push_str(&killed)` so accumulated text reads in screen order when yanked.

2. **No-op kills preserve chain** — Moved `app.last_was_kill = true` outside the inner guard for all three kill commands. A Ctrl+K at EOL or Ctrl+U/W at position 0 no longer resets the consecutive-kill chain.

3. **Channel switcher guard** — Added `!app.channel_switcher.show` check to Ctrl+Y, and also to Ctrl+U and Ctrl+W since they have the same class of bug (operating on main input while switcher is open).

Added 7 new tests covering backward prepend semantics, no-op chain preservation, and channel switcher guards.

PR #1248, opened 2026-02-18
feat: Add time-based spinner animation tick to TUI chat [Midtown !1475]

btucker:
## Summary

- Added a dedicated ~100ms animation timer to the TUI chat event loop using `tokio::time::interval`
- Made spinner frame advancement time-based: tracks `Instant` of last frame change and only advances after 100ms has elapsed
- Only triggers redraws on animation ticks when a spinner is actually visible (`any_spinner_visible()`)
- Added `spinner_tests.rs` with 7 unit tests covering time-based behavior, visibility checks, and determinism

## Problem

The TUI spinner was advancing 1 frame per `draw()` call. With a 1-second idle refresh interval, this made the braille spinner (`⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏`) visibly sluggish — only 1 fps instead of the expected ~10 fps.

## Fix

Split spinner logic into two responsibilities:
- `spinner_char(&self)` — read-only, returns current frame (previously `&mut self` and advanced on read)
- `tick_spinner(&mut self)` — advances frame only if ≥100ms elapsed since last tick
- `any_spinner_visible(&self)` — checks if lead is working or any coworker is active (non-idle)

The event loop drives a 100ms `animation_interval` and calls `app.tick_spinner()` only when spinners are visible.

## Test plan
- [x] `cargo test spinner` — 7 tests pass

### DIFF
--- a/src/bin/midtown/cli/chat/app.rs
+++ b/src/bin/midtown/cli/chat/app.rs
@@ -1780,16 +1780,21 @@ impl App {
         SPINNER_FRAMES[self.spinner_frame % SPINNER_FRAMES.len()]
     }
 
+    /// Returns true if any spinner is currently visible (lead working or active coworkers).
+    pub fn any_spinner_visible(&self) -> bool {
+        self.lead_working
+            || self
+                .coworkers
+                .iter()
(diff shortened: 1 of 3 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1249, opened 2026-02-18
fix: Load channel messages when switching via Ctrl+K channel switcher [Midtown !1492]

btucker:
## Summary
- Fixes channel_switcher_select() (Ctrl+K) not loading messages for the newly selected channel
- Previously, switching channels updated `selected_channel` but called `scroll_to_bottom()` instead of `load_channel_messages()`, leaving the old channel's messages visible
- Also detects and sets `selected_channel_archived` for the newly selected channel before switching

## Root cause
`scroll_to_bottom()` only adjusts the scroll offset on already-loaded messages. The fix replaces it with `load_channel_messages()`, which reads from the correct channel's JSONL file.

## Test plan
- [ ] Switch channels with Ctrl+K and verify the new channel's messages are displayed
- [ ] Verify archived channels are correctly detected and shown with archived styling after switching

### DIFF
--- a/src/bin/midtown/cli/chat/app.rs
+++ b/src/bin/midtown/cli/chat/app.rs
@@ -220,6 +220,9 @@ pub struct App {
     /// Test-only: captures the channel argument from the last post_message call
     #[cfg(test)]
     pub last_posted_channel: Option<String>,
+    /// Test-only: tracks whether load_channel_messages was called
+    #[cfg(test)]
+    pub load_channel_messages_called: bool,
     /// Tasks for the kanban board
     pub tasks: Vec<KanbanTask>,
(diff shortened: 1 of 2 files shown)


btucker:
## Review Status

---
> [\!NOTE]

btucker:
Addressed in b5f9e5a. Added `test_channel_switcher_loads_messages_from_new_channel` regression test using a `#[cfg(test)]` tracking flag (`load_channel_messages_called`) — same pattern as the existing `last_posted_channel` field. The test verifies `channel_switcher_select()` calls `load_channel_messages()` rather than `scroll_to_bottom()`.

All 6 channel switcher tests pass, clippy clean.

PR #1254, opened 2026-02-18
fix: Don't orphan active coworker's PR when worktree is missing [Midtown !1499]

btucker:
## Summary

- Fixed orphan detection in `collect_reviewer_effects_with_source` that incorrectly blocked reviewer assignment for active coworkers' PRs when their worktree had no registry entry
- The bug caused PR #1246 (park's kill-ring PR) to go 50+ minutes without a reviewer, despite 4+ idle coworkers available
- Added failing test and captured snapshot that reproduces the exact scenario

## Root Cause

In `src/daemon/pr.rs`, the orphan detection logic identifies a coworker from the branch prefix (e.g., "park" from `park/task-1483-kill-ring-yank-v2`) but then finds no worktree entry for that PR's task. The code then marks the PR as orphaned and skips reviewer assignment — even though the coworker is actively running and can address review feedback.

The flow for PR #1246:
1. `worktree_registry.get_by_pr(1246)` → None (no `review-pr-1246` entry)
2. Extract task_id 1483 from title → no worktree with task_id="1483" → None  
3. `get_by_branch("park/task-1483-...")` → None
4. `worktree = None` → `is_lead_branch` → false
5. `coworker_from_branch_with_map` → returns `Some("park")` (matches prefix)
6. **Bug**: marks as orphaned without checking if park is actively running

## Fix

Before marking a PR orphaned when the owner coworker has no registered worktree, check if the coworker is currently running. Active coworkers can always receive and act on review feedback.

## Test Plan

- [x] New test `test_active_coworker_pr_without_worktree_is_not_orphaned` — fails before fix, passes after
- [x] Added snapshot fixture `snapshot-reviewer-not-assigned-pr-1246-20260218-001618.json`
- [x] Clippy passes (`cargo clippy -- -D warnings`)

### DIFF
--- a/src/daemon/pr.rs
+++ b/src/daemon/pr.rs
@@ -1973,15 +1973,6 @@ pub(crate) async fn collect_reviewer_effects_with_source(
 ) -> Vec<Effect> {
     let mut effects: Vec<Effect> = Vec::new();
 
-    // Build the set of running coworker names once, mirroring collect_world_snapshot().
-    // Avoids per-PR state queries for liveness checks below.
-    let running_names: std::collections::HashSet<String> = state
-        .coworkers
-        .list_running()
-        .into_iter()
-        .map(|cw| cw.name.to_lowercase())
-        .collect();
-
     for pr in prs {
         let pr_number = pr.get("number").and_then(|n| n.as_u64()).unwrap_or(0);
         if pr_number == 0 {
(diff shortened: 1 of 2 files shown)


btucker:
Addressed all 5 issues in 1bec850:

1. **Stale comment** — Updated to mention \"no running coworker\" as a third orphan criterion alongside no worktree and indeterminate author.

2. **state.coworkers.get() direct query** — Replaced with a `running_names: HashSet<String>` built once before the loop using `list_running()`, mirroring the `collect_world_snapshot()` pattern. Each per-PR liveness check is now a `contains()` lookup rather than a live state query.

3. **Snapshot fixture not loaded** — Added a doc comment explaining the limitation: `collect_reviewer_effects_with_source` takes `&DaemonState`, not `&WorldSnapshot`, so the fixture can't be loaded without significant plumbing. The fixture documents the exact production scenario.

4. **Unnecessary `#[allow(clippy::await_holding_lock)]`** — Removed.

5. **Three-strategy comment** — Updated to document the fourth fallback (coworker liveness check) with a short explanation.

PR #1261, opened 2026-02-18
feat: Disallow creating channels with coworker avenue names [Midtown !1521]

btucker:
## Summary
- Blocks creation of channels named after coworker avenue names (lexington, park, madison, broadway, amsterdam, columbus, riverside, york, pleasant, vernon)
- Fix is at `Channel::new()` — the single creation chokepoint for all paths (TUI, daemon RPC, ChannelRouter)
- Made `AVENUE_NAMES` public in `coworker.rs` so `channel.rs` can reference it directly (no duplication)
- Added test: `test_channel_name_rejects_coworker_avenue_names`

## Why
After the `ch-` prefix removal, a channel named `park` would collide with the `park` channel lead session. Blocking at creation time is cleaner than filtering in `next_available_name`.

## Test plan
- [x] `test_channel_name_rejects_coworker_avenue_names` — all 10 avenue names rejected
- [x] All existing channel tests pass (53 tests)
- [x] Full test suite passes (1482+ tests)
- [x] Clippy clean

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -120,6 +120,14 @@ impl Channel {
             )));
         }
 
+        // Reject names reserved for coworker sessions to prevent naming collisions.
+        if crate::coworker::AVENUE_NAMES.contains(&channel_name.as_str()) {
+            return Err(crate::Error::InvalidMessage(format!(
+                "Channel name '{}' is reserved for coworker sessions and cannot be used as a channel name",
+                channel_name
+            )));
+        }
+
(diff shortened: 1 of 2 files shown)


btucker:
Fixed both issues:

1. Switched to `is_coworker_name(&channel_name)` — covers both avenue and overflow name pools. Reverted `AVENUE_NAMES` back to private since the public API handles the check.

2. Expanded test to cover all 16 coworker names (10 avenue + 6 overflow).

PR #1263, opened 2026-02-18
feat: Show channel lead tool calls in topic channels [Midtown !1520]

btucker:
## Summary

- Main channel continues to show the lead's tool calls (unchanged behavior)
- Topic channels (e.g. `#web`, `#features`) now show the channel lead's tool calls instead of the main lead's
- Tool calls are filtered per channel so switching channels shows only the relevant activity

## Implementation

**Backend (`src/daemon/stream.rs`):**
`process_universal_events` now accepts `channel_lead_sessions` and emits `BroadcastUniversalItems` effects for both the main lead (`channel=None`) and each active channel lead (`channel=Some(channel_name)`). Coworker tool calls continue to be ignored. The session name equals the channel name, so no extra mapping is needed.

**Protocol (`src/daemon/effects.rs`, `src/web.rs`):**
`BroadcastUniversalItems` effect and `UniversalItemsData` WebSocket payload gain an optional `channel` field. `None` = main channel, `Some(name)` = topic channel.

**Web UI (`web-app/src/lib/`):**
- `agentToolItems` store is now keyed by channel name (`'midtown'` for the main lead, or the topic channel name for channel leads)
- `Channel.svelte` renders only the active channel's tool activity strip using `$agentToolItems[$activeChannel]`

## Test plan

- [x] 4 new unit tests covering channel lead scoping behavior
- [x] Existing tests updated for new `process_universal_events` signature  
- [x] `cargo test` passes
- [x] `cargo clippy -- -D warnings` clean

🌃 Co-built with [Midtown](https://github.com/btucker/midtown)

PR #1264, opened 2026-02-18
fix: Tool call display shows 'working...' only when agent is active [Midtown !1523]

btucker:
## Summary
- Gate the 'working…' label on whether the agent has an active (non-idle) phase in `$coworkers`, not on `item.status === 'InProgress'` (which was never set since all items arrive as `Completed`)
- Collapse the tool call activity strips and the lead typing indicator into a **single flex-wrap activity bar** above the input — all agents appear as inline chips on one line, using full screen width instead of truncating
- Each chip shows: `agentname › current-tool` (working) or `agentname ✓/✗ last-result` (done)
- Lead typing dots appear in the same row
- Expand `process_universal_events` in `stream.rs` to broadcast tool events from all sessions (lead + coworkers), not just the lead
- Update tests to reflect coworker tool calls now produce broadcast effects
- Remove unused `ToolActivity` import from `Channel.svelte`

## Test plan
- [x] Unit tests for `process_universal_events` updated and passing (`cargo test stream`)
- [x] `npm run build` succeeds
- [x] Activity bar: working agents show `agentname › tool-name`, idle agents show `agentname ✓ last-tool`, lead typing dots share the same row


btucker (commented):
The `stream.rs` change to broadcast universal events for all agents (not just lead) is correct and well-tested.

However, the Svelte UI change has a correctness issue: **the backend never produces `InProgress` status items**. In `src/universal_events/claude.rs`, both tool_use (line 192) and tool_result (line 224) are emitted as `ItemStatus::Completed`. This means `toolItems.some((item) => item.status === 'InProgress')` will always be `false`, and the "working..." label will **never** display — swapping the bug from "always shows working" to "never shows working."

To fix this, the backend needs to emit `InProgress` items. One approach: when a `tool_use` content block appears in an `Assistant` event, emit it as `InProgress`, then transition to `Completed` when the matching `tool_result` arrives in a `User` event.

**Not merging until this is addressed.**

btucker:
Closing in favor of PR #1267 (broadway's TUI fix) which takes a better approach. The web UI tool call display was already addressed by merged PRs #1262 and #1263. This PR has persistent merge conflicts and the InProgress status issue flagged in review.

PR #1271, opened 2026-02-18
feat: Add midtown config get/set/list CLI commands [Midtown !1528]

btucker:
## Summary

Adds `midtown config get/set/list` CLI commands so the lead can manage midtown settings without manually editing TOML files.

- `midtown config list` — show all 14 supported settings with current values and config file path
- `midtown config get <key>` — read a single setting value
- `midtown config set <key> <value>` — set and persist a setting

All commands accept `--global` to operate on `~/.midtown/config.toml`; without it, they use the project config inferred from cwd.

**Supported keys (14):**
- `default.personality`, `default.max_coworkers`, `default.chat_layout`, `default.chat_min_width`, `default.zellij_swap_layout`, `default.zellij_chat_pane_size`, `default.user_display_name`, `default.bin_command`
- `daemon.webhook_port`, `daemon.webhook_secret`, `daemon.pr_poll_interval_secs`, `daemon.chat_monitor_enabled`, `daemon.github_user`, `daemon.worktree_cleanup_retention_hours`

**Key implementation details:**
- Local-only command (no daemon required)
- Uses existing `GlobalConfig::save_to` and `FullProjectConfig::save_to` which preserve TOML comments
- Propagates TOML parse errors instead of silently overwriting user config with defaults
- Validates `zellij_chat_pane_size` is within 10-90 range at set time
- `unreachable!()` on fallback read arms to surface future maintenance bugs

## Test plan

- [x] 34 unit tests in `config_tests.rs` covering: unknown key errors, type validation errors, set/get roundtrips for all key categories (string, int, bool, enum), pane size range validation (both scopes), list output format, persistence via TOML reload, multiple sets accumulating
- [x] `cargo test` passes

### DIFF
--- a/src/bin/midtown/cli/config.rs
+++ b/src/bin/midtown/cli/config.rs
@@ -52,6 +52,7 @@ const VALID_KEYS: &[&str] = &[
     "daemon.pr_poll_interval_secs",
     "daemon.chat_monitor_enabled",
     "daemon.github_user",
+    "daemon.webhook_restart_interval_secs",
     "daemon.worktree_cleanup_retention_hours",
 ];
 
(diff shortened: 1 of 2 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1273, opened 2026-02-18
fix: Tool call display shows 'working...' only when tool is actually in flight [Midtown !1523]

btucker:
## Summary

Fixes the 'working…' label in the web UI tool call display to only appear when a tool is genuinely executing.

**Root cause:** The backend (`extract_tool_events`) always emitted `ItemStatus::Completed` for `tool_use` blocks, regardless of whether a matching `tool_result` had arrived. The frontend's `InProgress` check therefore never matched.

**Fix:**
- In `extract_tool_events`, collect all `tool_use_id` values from `tool_result` blocks in the same drain batch first, then emit each `tool_use` as `InProgress` (no result yet) or `Completed` (result present in same batch)
- Gate the 'working…' label in `Channel.svelte` on `activeChannelToolItems.some(item => item.status === 'InProgress')`
- Add three new tests covering: tool_use without result → InProgress, tool_use with matching result → Completed, and mismatched call IDs

This PR replaces PR #1264 (closed without merging), which addressed the same task but had scope issues (reverting the coworker-only restriction) and the backend fix was incomplete.

## Changes

- `src/universal_events/claude.rs`: Add `collect_completed_call_ids` helper; use it in `extract_tool_events` to set `InProgress` vs `Completed` per call
- `src/universal_events/claude_tests.rs`: Update existing test assertions, add 3 new tests
- `web-app/src/lib/Channel.svelte`: Add `InProgress` guard on the 'working…' label

## Test plan
- [x] `cargo test` — all 1488+ tests pass
- [x] `npm run build` — clean
- New tests: `test_extract_tool_events_tool_use_without_result_is_in_progress`, `test_extract_tool_events_tool_use_with_result_for_different_call_is_in_progress`, updated `test_extract_tool_events_mixed_assistant_and_user_events`


btucker:
Rebased onto main to resolve merge conflict (conflict was in claude_tests.rs — main added a multibyte truncation test, this branch added InProgress status tests; both sets of tests are preserved).

Tests and clippy pass locally.

PR #1286, opened 2026-02-18
feat: Add NamePool and session registry to DaemonState [Midtown !1562]

btucker:
## Summary

- Add `NamePool` (LRU name pool) and session reverse-index maps to `DaemonState`
- Initialize from `AVENUE_NAMES + OVERFLOW_NAMES` at startup, restoring allocated names from `persistent_state.sessions`
- Update session drain loop: populate maps on init event, release name and clean up maps when a session stops
- Add `session_for_name()`, `name_for_session()`, `session_for_task()` helpers
- Make `OVERFLOW_NAMES` `pub const` in `coworker.rs` (needed for initialization in `mod.rs`)

This is Task 8 of the session-centric coworker model refactor. It lays the in-memory infrastructure (name pool + reverse maps) that future tasks (effect execution, RPC adaptation) will use to route by session ID rather than by name.

## Test plan

- [x] `cargo test` passes (2,000+ tests)
- [x] `cargo clippy -- -D warnings` clean
- [x] No nested lock acquisitions — each map lock is acquired and released independently

🌃 Co-built with [Midtown](https://github.com/btucker/midtown)

PR #1290, opened 2026-02-18
fix: Move NamePool/reverse-map cleanup into cleanup_coworker_state [Midtown !1562]

btucker:
## Summary
- Follow-up to PR #1286 — addresses review feedback from vernon
- Moved NamePool release and reverse-map cleanup (`name_to_session`, `session_to_name`, `task_to_session`) into `cleanup_coworker_state()` so both the intentional shutdown path and session-death path get identical cleanup
- Without this fix, intentional shutdowns via `Effect::ShutdownCoworker` would leave stale entries in all three maps and never release names back to `NamePool`, causing name exhaustion on repeated shutdown/spawn cycles
- Updated existing tests to exercise `cleanup_coworker_state()` directly instead of duplicating the cleanup logic inline

## Test plan
- [x] All 1552 library tests pass
- [x] Clippy clean with `-D warnings`
- [x] `test_cleanup_releases_name_from_pool` — verifies cleanup_coworker_state releases name and clears all maps
- [x] `test_cleanup_preserves_other_sessions_reverse_maps` — verifies other coworkers' state is untouched
- [x] `test_cleanup_with_no_session_id_is_noop_for_reverse_maps` — verifies no panic when reverse maps unpopulated
- [x] `test_task_to_session_cleanup_removes_all_tasks_for_session` — verifies multi-task cleanup via shared function

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -700,6 +700,11 @@ impl DaemonState {
     /// shared function, the two paths can drift out of sync — e.g., session death
     /// missing cooldown/nudge/assignment cleanup that shutdown handles. See PR #1268.
     ///
+    /// Handles: coworker deregistration, stop-time recording, coworker records,
+    /// cooldowns, pending nudges, task assignments, recent tool activity,
+    /// NamePool release, and session reverse-map cleanup (name_to_session,
+    /// session_to_name, task_to_session).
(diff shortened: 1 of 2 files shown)


btucker:
## Review Status

---
> [!NOTE]

PR #1303, opened 2026-02-19
feat: Add thread support to Web UI — stores, ThreadPanel, mobile [Midtown !1577]

btucker:
## Summary

- Add `threadData` store and thread-aware message routing (sendMessage, fetchThread, openThread, closeThread) to web UI API layer
- Create `ThreadPanel.svelte` component with header (parent message preview), scrollable reply list, and reply input
- Wire ThreadPanel into App.svelte as a third column, mutually exclusive with DetailPanel
- Add clickable reply indicators ("N replies") below messages with thread replies in Channel.svelte
- Mobile: full-screen slide-in overlay with back button, safe area insets, and CSS transition animation

## Details

**Stores & API (`store.js`, `api.js`):**
- `threadData` writable store: `{ parentMessage, channelName, messages }` or null
- Thread-aware WebSocket routing: thread replies update `threadData` and bump parent `reply_count` without entering the main channel timeline
- `annotateThreadReplyCounts()` helper filters thread replies from main timeline and annotates parent messages on history fetch
- `openThread`/`closeThread` maintain mutual exclusivity with `detailPanelData`

**ThreadPanel (`ThreadPanel.svelte`):**
- Desktop: 380px side panel with `hidden lg:flex`, matching DetailPanel styling
- Mobile: full-screen fixed overlay with slide-in animation, back button, safe area insets
- Dual `bind:this` variables with `matchMedia` listener for robust desktop/mobile switching
- Escape closes panel (guarded with `defaultPrevented` to avoid autocomplete conflicts)
- Auto-focus reply textarea on open, auto-scroll on new messages

**Reply indicators (`Channel.svelte`):**
- Clickable "N replies" + last replier name below top-level messages with threads
- Clicking opens the ThreadPanel via `openThread(msg, $activeChannel)`
- `closeThread()` called before opening detail panels (bidirectional mutual exclusivity)

## Test plan

- [x] `cargo test` passes (19/19)
- [x] `cargo clippy -- -D warnings` passes
- [x] `npm run build` in web-app succeeds
- [ ] Manual: Send message, reply in thread, verify reply indicator appears
- [ ] Manual: Click reply indicator → thread panel opens with replies
- [ ] Manual: Type reply in thread panel → appears in thread
- [ ] Manual: Escape closes thread panel
- [ ] Manual: Click task/PR link while thread open → thread closes, detail panel opens
- [ ] Manual: Mobile viewport → thread as full-screen overlay with back button

🌃 Co-built with [Midtown](https://github.com/btucker/midtown)

PR #1306, opened 2026-02-19
fix: Address thread UI review feedback [Midtown !1577]

btucker:
Follow-up to PR #1303 (feat: Add thread support to Web UI) — addresses review feedback from the code review completed after the PR was merged.

## Summary

- Thread replies no longer increment unread badge — moved `channels.update` unread increment inside the `else` (top-level message) branch; thread replies don't appear in the main timeline so they shouldn't trigger notifications
- Thread replies no longer clear tool activity — moved `agentToolItems.update` inside same `else` branch; a coworker posting a thread reply mid-task should not reset their activity strip
- `openThread` race condition fixed — `.then()` callback now guards with staleness check (`parentMessage.id` must still match current `threadData`), and merges WS-delivered replies with fetched history instead of overwriting
- `threadData` cleared on channel switch — added `closeThread()` call at the start of `ChannelList.selectChannel`; thread context is channel-scoped
- Legacy `messages` store consistency — `fetchHistory` now filters thread replies from the legacy store (matching `messagesByChannel` and the WS handler behavior)

## Test plan
- [ ] Receive thread reply in background channel → unread badge does NOT increment
- [ ] Post thread reply while tool activity is showing → activity strip not cleared
- [ ] Open thread A, quickly open thread B → thread B's content shown (not A's)
- [ ] Open thread, switch channel → thread panel closes
- [ ] Page reload → legacy messages store contains no thread replies


btucker:
I've now verified this directly via the GitHub API against the PR commit `d5c0e0c` (not local files). All three issues are fixed in the PR commit:

**Issue 1 fixed** — `channels.update` and `agentToolItems.update` are inside the `else` branch in the PR commit. The parent commit `23d17fb` had them outside the if/else, which is what Columbus's review described. The PR commit corrects this.

**Issue 2 fixed** — `openThread` staleness guard is present:
```js
threadData.update((td) => {
  if (\!td || td.parentMessage.id \!== parentMessage.id) return td
  ...
})
```

**Issue 3 fixed** — `ChannelList.svelte` in the PR commit imports `closeThread` and calls it at the start of `selectChannel()`:
```js
import { fetchHistory, fetchChannels, getApiBase, closeThread } from './api.js'
...
function selectChannel(channelName) {
  closeThread()
  activeChannel.set(channelName)
  ...
}
```

Columbus's review linked to line numbers from the **base/parent commit** (`23d17fb`), where those bugs exist. The PR commit (`d5c0e0c`) contains the fixes. The PR is correct and ready to merge.

PR #1311, opened 2026-02-19
fix: Fix mobile web UI autocomplete positioning — pin to top of input field [Midtown !1603]

btucker:
## Summary
- Fixed autocomplete dropdown appearing at top of screen instead of above the input field on mobile
- Switched from `position: fixed` to `position: absolute` for the autocomplete dropdown
- Wrapped the input form in a new `position: relative` container (`formWrapperElement`) so the autocomplete can be absolutely positioned relative to it
- This avoids iOS visual/layout viewport split issues that cause `position: fixed` coordinates to misalign when the virtual keyboard is open

## Root Cause
The previous implementation used `position: fixed` with coordinates from `getBoundingClientRect()`. On iOS Safari with the virtual keyboard open, the visual viewport (area above keyboard) and layout viewport diverge, causing `position: fixed` elements to be placed incorrectly. By using `position: absolute` relative to a local ancestor that has no `overflow: hidden`, we avoid this coordinate system issue entirely.

## Test plan
- [ ] Open web UI on mobile iOS
- [ ] Tap the message input to open keyboard
- [ ] Type `@` to trigger autocomplete
- [ ] Verify dropdown appears directly above the textarea, not at the top of the screen
- [ ] Verify autocomplete works on desktop (no regression)
- [ ] Verify keyboard navigation (arrow keys, Enter, Escape) still works

### DIFF
--- a/web-app/src/lib/Autocomplete.svelte
+++ b/web-app/src/lib/Autocomplete.svelte
@@ -35,7 +35,7 @@
 
 {#if show && items.length > 0}
   <div
-    class="fixed z-[1000] bg-[#1a1a1a] border-2 border-[#2a2a2a] rounded-lg max-h-[280px] overflow-y-auto shadow-[0_4px_12px_rgba(0,0,0,0.5)] min-w-[200px] max-w-[400px] -translate-y-[calc(100%+8px)]"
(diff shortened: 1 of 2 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1329, opened 2026-02-19
feat: Formalize ops channel lead role [Midtown !1620]

btucker:
## Summary

Formalizes the #ops channel lead role with three connected changes:

- **Lead delegation** (`agents/lead.md`): Added a "Channel Lead Delegation" section so the Lead knows to route ops-domain questions (CI, deployment, build infrastructure) to the #ops channel lead rather than handling them directly. Also documents the general pattern for engaging any channel lead.

- **Daemon respawn health check** (`health.rs`, `events.rs`): Added `ensure_channel_leads_alive()` — a periodic health check that respawns dead channel lead sessions. Previously, channel leads were only spawned at daemon startup and on `CreateChannel`; a crash mid-session left the channel without a lead until daemon restart. The new function runs on `TaskDispatchTick` alongside `ensure_lead_alive()`, skipping in-flight spawns, already-running leads, and recently-stopped leads (LEAD_RESPAWN_COOLDOWN prevents crash loops).

- **Escalation clarification** (`agents/channel-lead.md`): Strengthened escalation rules with an explicit "broader project context needed" trigger, so the ops lead (and all channel leads) know to escalate to @lead rather than guessing when they lack project-level context.

## Test plan
- [x] `cargo build` passes
- [x] Pre-commit hooks pass

### DIFF
--- a/agents/channel-lead.md
+++ b/agents/channel-lead.md
@@ -40,8 +40,9 @@ Not everything belongs in your channel. Escalate when:
 - **Task creation needed**: Escalate to @lead — you cannot create tasks directly
 - **Questions outside your domain**: Redirect to @lead or another channel lead
 - **Architectural decisions affecting the whole project**: Always involve @lead
+- **Broader project context needed**: If you lack the project-level context to give a complete answer, escalate to @lead rather than guessing
 
-Keep domain questions in the channel. Reserve @lead escalations for things that genuinely require project-wide coordination.
+Keep domain questions in the channel. Reserve @lead escalations for things that genuinely require project-wide coordination or that you cannot answer from your domain context alone.
 
 ## Domain Context
 
--- a/agents/lead.md
+++ b/agents/lead.md
@@ -146,6 +146,30 @@ midtown task create "Subject" --description "Details..."
 midtown coworker call-in
 ```
 
+## Channel Leads
+
+Topic channels have dedicated **channel leads** — domain experts who maintain persistent context for their area. Channel leads are distinct from regular coworkers: they do not implement code or open PRs. They brainstorm, answer domain questions, and track active work in their channel.
+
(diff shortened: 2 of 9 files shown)


btucker:
Addressed all 6 review issues in the force-push (7133bf0):

1. **Channel lead idle shutdown loop** — Added `channel_lead_names` field to `IdleShutdownContext` and updated `decide_idle_shutdowns` to exempt channel leads. They now get the same treatment as the `"lead"` session. Updated `check_and_shutdown_idle_coworkers` to pass `channel_lead_sessions.keys()` into the context.

2. **SessionMode::Fresh discarding saved session** — `ensure_channel_leads_alive` now uses `SessionMode::ResumeSession(session_id)` when a non-empty session_id is stored, falling back to Fresh only when the session_id was cleared (crash recovery path).

3. **dispatch_via_sessions docstring** — Removed the false "Pure function" claim. The docstring now acknowledges the plan file I/O in `build_plan_prompt_section`, consistent with the analogous comment on `check_and_recover_orphans_with_task_lookup`.

4. **MultiTickHarness not updated** — Added `ensure_channel_leads_alive` call to the `SessionMonitorTick` handler and updated the struct-level docstring listing called functions.

5. **Not exported in mod.rs** — Added `ensure_channel_leads_alive` to the `pub use health::{...}` block.

6. **No tests** — Added 7 tests for `ensure_channel_leads_alive` covering all branches: empty sessions, already-running, in-flight spawn (no stop_time → skip), crash recovery within cooldown (skip), crash recovery after cooldown (Fresh spawn), normal recovery within cooldown (skip), and normal recovery after cooldown (ResumeSession). Also added 2 tests for the idle shutdown channel lead exemption.

Also incorporated improvements to `ensure_channel_leads_alive` found during implementation review: the in-flight vs. crash-recovery distinction uses `coworker_stop_times` presence as the discriminator.

PR #1337, opened 2026-02-19
fix: Treat channel leads as leads everywhere in decision functions [Midtown !1623]

btucker:
## Summary

Channel leads (domain expert leads for topic channels like #ops, #daemon) were incorrectly treated as regular coworkers in several decision functions, causing them to be idle-shutdown after 60s, lose context, and constantly re-introduce themselves.

The main lead is conceptually the channel lead for #midtown — channel leads should receive the same exemptions the lead gets.

## Fixes

- **Idle shutdown exemption** (`rules.rs`, `health.rs`): Added `channel_lead_names` to `IdleShutdownContext`; channel leads are now exempt from idle shutdown, same as the main lead
- **Pending task action skip** (`rules.rs`, `dispatch.rs`): Added `is_channel_lead` param to `decide_pending_task_action()`; channel-lead-owned tasks are now skipped (they're not task-driven)
- **Orphan recovery skip** (`rules.rs`, `dispatch.rs`): Added `channel_lead_names` to `OrphanRecoveryContext`; channel-lead-owned tasks no longer trigger spurious orphan recovery
- **Duplicate worker skip** (`dispatch.rs`): `check_for_duplicate_task_workers()` now skips channel leads alongside the main lead
- **Auto-detach message** (`effects.rs`): Channel leads now get "Channel lead session will be respawned for its channel." instead of the generic coworker reassignment message
- **Role on attach** (`cli/session.rs`): Channel leads now get `CoworkerRole::ChannelLead` (with the correct channel lead system prompt and sonnet model) instead of `CoworkerRole::Coworker`

## Tests

- `channel_lead_is_not_idle_shutdown` — idle channel lead is NOT shutdown
- `channel_lead_owned_task_is_skipped` — channel-lead-owned pending task returns `Skip`
- `channel_lead_owned_task_not_orphan_recovered` — channel-lead-owned in-progress task returns `None`
- `test_channel_lead_attach_gets_channel_lead_role` — attach path uses channel lead system prompt containing the channel name

Closes #1623

### DIFF
--- a/src/daemon/health.rs
+++ b/src/daemon/health.rs
@@ -24,6 +24,7 @@ use super::{DaemonState, snapshot};
 /// automatically sent on a break.
 ///
 /// IMPORTANT: Coworkers are NEVER sent on a break if any of these apply:
+/// - They are a channel lead (long-lived domain expert session, like "lead")
 /// - They have open unmerged PRs (must stay available for review feedback)
 /// - They have active review assignments
 /// - They have unblocked dependent tasks
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -328,6 +328,7 @@ pub(crate) struct IdleShutdownContext<'a> {
 /// without performing any side effects or mutations.
 ///
 /// A coworker is protected from break if:
+/// - They are a channel lead (long-lived domain expert session, like "lead")
 /// - They have in-progress tasks (busy)
 /// - They have a pending task assigned to them
 /// - They have open unmerged PRs with CI not yet passed (waiting for CI)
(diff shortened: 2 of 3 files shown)


btucker:
Addressed both review items in 3e68a08:

**1. Tests moved to separate file** — Created `rules_channel_lead_tests.rs` with the 3 channel lead tests calling production functions directly (no builder dependency on the inline module). Also removed the now-dead `channel_lead_names()` method on `OrphanCtx` to fix the resulting `-D warnings` clippy failure.

**2. Doc comments updated** — Added channel lead exemption to the protection lists in both `decide_idle_shutdowns` (rules.rs) and `check_and_shutdown_idle_coworkers` (health.rs).

**Re: missing snapshot fixture** — The task plan mentioned using a captured snapshot, but the builder-pattern tests exercise the same code paths with more targeted assertions. Creating a follow-up task for the snapshot fixture approach.

**Re: session recovery loop gap (amsterdam's channel note)** — Creating a follow-up task for the `dispatch_via_sessions_with_task_lookup` defensive check. It's a valid gap even if unlikely in practice.

PR #1350, opened 2026-02-20
fix: Exclude channel leads from coworker count in status display [Midtown !1647]

btucker:
## Summary
- Channel leads are persistent domain experts, not workers competing for slots — but the status display counted them, showing misleading numbers like "13/12 active"
- Added `is_channel_lead` field to the `CoworkerInfo` struct and the status RPC response, so clients can distinguish channel leads from regular workers
- Fixed `midtown status` CLI display (`response.rs`) to exclude channel leads from the worker count
- Fixed `rpc_status.rs` `active_coworkers` field to exclude channel leads
- Fixed TUI board (`board.rs`) to also filter by `channel_lead_names` when computing the active count (defensive fix for the `try_lock` failure path in `build_coworkers_data`)

## Test plan
- [x] New test `test_status_pretty_excludes_channel_leads_from_count` verifies channel leads are excluded from the displayed count

### DIFF
--- a/src/daemon/rpc_status.rs
+++ b/src/daemon/rpc_status.rs
@@ -103,31 +103,8 @@ pub(super) async fn handle_status(id: RequestId, state: &DaemonState) -> Respons
         (ps.github.rate_limit.clone(), names)
     };
 
-    // Tag each coworker as a channel lead and compute the non-lead count.
-    // Channel leads are persistent domain experts and do not consume coworker slots.
-    let coworkers: Vec<serde_json::Value> = coworkers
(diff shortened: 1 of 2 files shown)


btucker:
## Review Status

---
> [\!NOTE]

btucker:
Addressing both issues now:
1. CLI row filter — will add filter to the display loop so channel leads don't appear as rows
2. Board.rs — lead confirmed the defensive filter is redundant since build_coworkers_data already excludes upstream. Will remove it and skip the test.

PR #1382, opened 2026-02-20
fix: Clear inbox on name allocation to prevent message bleed [Midtown !1675]

btucker:
## Summary

- Names are recycled from the `NamePool` across sessions. Without clearing the inbox, a newly-allocated session inherits unread messages from the previous session that held the same name.
- Added `clear_inbox` / `clear_inbox_at` to `mailbox.rs` with a testable internal function (matching the existing `ensure_team_config_at` pattern).
- Call `clear_inbox` in `SpawnSession` (effects.rs) right after name allocation, before spawning — this is the critical fix point.
- Call `clear_inbox` in `cleanup_coworker_state` (mod.rs) on name release — defensive cleanup so the inbox is drained before the name re-enters the pool.

## Test plan

- [x] `test_clear_inbox_removes_existing_messages` — verifies stale inbox is removed before re-use
- [x] `test_clear_inbox_no_file_is_ok` — verifies no-op when inbox doesn't exist (first allocation)

### DIFF
--- a/src/mailbox_tests.rs
+++ b/src/mailbox_tests.rs
@@ -1,9 +1,6 @@
 use super::*;
 use tempfile::TempDir;
 
-/// Override the teams dir for testing by working with a temp directory.
-/// We test the internal functions directly by constructing paths manually.
-
 /// Verifies that `clear_inbox` removes an existing inbox so a newly-allocated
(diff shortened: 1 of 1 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1383, opened 2026-02-20
feat: Click coworker name in TUI status box to attach to session [Midtown !1676]

btucker:
## Summary

- Added `coworker_line_map: HashMap<u16, String>` to `App` state — maps absolute terminal y-coordinates to coworker names
- `draw_coworker_status()` now populates this map each render pass, inserting one entry per active (non-idle) coworker row
- Mouse click handler checks `coworker_line_map` after the board area check; clicking a coworker name returns `EventResult::AttachCoworker(name)`, which routes through the existing `attach_session_split()` flow

## How it works

The `coworker_line_map` uses **absolute y-coordinates** (unlike `task_line_map` which uses coordinates relative to the board content area). This is because the coworker status table sits outside the board area rect, so the click handler checks it using the raw terminal `y` from the mouse event.

The first data row in the coworker table is at `chunks[1].y + 1` (skipping the table's top border), and each subsequent coworker occupies exactly one row.

## Test plan

- [x] `test_draw_board_panel_populates_coworker_line_map`: verifies both active coworkers appear in the map on consecutive y-lines
- [x] `test_coworker_line_map_excludes_idle`: verifies idle coworkers are excluded (matching the filter already applied to the table rows)
- [x] All 69 board tests pass

### DIFF
--- a/src/bin/midtown/cli/chat/app.rs
+++ b/src/bin/midtown/cli/chat/app.rs
@@ -1338,6 +1338,9 @@ impl App {
         {
             self.load_channel_messages_called = true;
         }
+        // Defensive: clear coworker line map on channel switch so stale entries
+        // from the previous render can't be clicked before the next draw pass.
+        self.coworker_line_map.clear();
         let channel_repo =
             midtown::paths::detect_repo_name().unwrap_or_else(|| "default".to_string());
         let base_dir = midtown::paths::projects_dir_for_repo(&channel_repo);
--- a/src/bin/midtown/cli/chat/coworker_click_tests.rs
+++ b/src/bin/midtown/cli/chat/coworker_click_tests.rs
@@ -1,0 +1,114 @@
+use crossterm::event::{Event, KeyModifiers, MouseEvent, MouseEventKind};
+use ratatui::Terminal;
+use ratatui::backend::TestBackend;
+
+use super::EventResult;
+use super::app::CoworkerInfo;
+use super::app::tests::test_app;
+use super::handle_event;
+use super::ui;
+
+/// Helper to create a left mouse click event at given terminal coordinates.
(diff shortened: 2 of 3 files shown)


btucker:
## Addressing review feedback

**Issue 1 (x-coordinate guard):** Fixed. Added `board_rect` x-bounds check so clicks outside the sidebar no longer trigger `AttachCoworker`. The committed code was indeed missing this — the coworker lookup had no x-guard, meaning any click at a matching y-coordinate (including in the chat panel) would trigger attach.

**Issue 2 (inconsistent coordinate systems):** Intentional — not changing. The coworker status section is rendered in a separate layout chunk *below* the board's scrollable content area (`draw_coworker_status` receives its own `Rect` from the vertical split at `board.rs:232`). Since it's not part of the board content, there's no content-relative origin to subtract from. The field doc on `coworker_line_map` already documents the absolute-y convention. Converting all three maps to the same coordinate space would require either a separate `coworker_area` rect or moving coworker rendering into the board content — both add complexity for no correctness benefit.

**Issue 3 (clear on channel switch):** Fixed. Added `coworker_line_map.clear()` in `load_channel_messages()` for defensive consistency, matching the pattern flagged in PR #1377.

**Issue 4 (missing click handler tests):** Fixed. Added two tests in `coworker_click_tests.rs`:
- `test_click_coworker_row_returns_attach` — renders with active coworkers, clicks at a valid (x, y) inside the sidebar, asserts `AttachCoworker` is returned
- `test_click_coworker_y_outside_sidebar_x_does_not_attach` — clicks at the same y but outside the sidebar x-bounds, asserts `AttachCoworker` is *not* returned (this test would fail against the pre-fix code, confirming issue 1 was a real bug)

All changes pushed in commit 93705ce.

PR #1384, opened 2026-02-20
fix: Clear stale is_running session flags on daemon startup [Midtown !1672]

btucker:
## Summary
- Sessions with `is_running=true` but `resume_on_startup=false` (reviewers, manually-stopped sessions) were skipped by `recover_from_session_records` but retained their stale flag across daemon restarts
- This caused dispatch to treat those sessions as still active, preventing pending tasks from being assigned indefinitely
- Added `clear_stale_running_sessions()` to `startup.rs`: after session recovery, clears `is_running` to `false` for any session not in the recovered set (excluding channel leads, which are recovered separately)

## Test plan
- [x] 5 new unit tests covering: normal stale sessions cleared, channel leads preserved, stale reviewers cleared, already-stopped sessions unaffected, empty state no-op
- [x] All 45 startup tests pass
- [x] `cargo fmt --check` clean
- [x] Cherry-picked snapshot fixture `snapshot-session-recovery-zombies-after-restart-20260218-215017.json`

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -96,7 +96,11 @@ When the daemon starts, it executes a careful cleanup and recovery sequence in `
 
 5. **Task assignment restore** — `restore_task_assignments_from_disk()` repopulates the in-memory task→coworker map from disk before any dispatch ticks fire, preventing duplicate coworker spawns.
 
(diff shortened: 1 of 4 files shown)


btucker:
## Review Status

---
> [!NOTE]

PR #1390, opened 2026-02-20
perf: Reduce TUI input latency and message display delay

btucker:
## Summary

Three optimizations that reduce perceptible latency in the TUI:

- **Messages appear immediately after Enter** — the Enter handler now calls `refresh()` right after a successful `post_message()` or `post_thread_reply()`, instead of waiting for the next `tailf` file-change event or the 1-second fallback timer
- **No heap allocation in `detect_autocomplete_trigger()`** — removed a `char_indices().collect()` Vec that was allocated on every single keystroke; replaced with lazy `nth()` + `char_indices().rev()` on a slice
- **No `Vec<String>` allocation in `calculate_input_bar_height()`** — replaced `wrap_content().len()` (which allocates a full `Vec<String>` every frame just to count) with a new `count_wrapped_lines()` in `text.rs` that mirrors the same word-boundary logic but counts without allocating

## Test plan

- [x] `test_message_appears_immediately_after_enter` — verifies new message is in `app.messages` right after `handle_event(Enter)`, without a separate `refresh()` call
- [x] `test_thread_reply_appears_immediately_after_enter` — same for thread replies
- [x] All existing 565 tests pass
- [ ] Manually: type a message, press Enter — confirm it appears without a delay

### DIFF
--- a/src/bin/midtown/cli/chat/mod.rs
+++ b/src/bin/midtown/cli/chat/mod.rs
@@ -26,7 +26,7 @@ use crossterm::{
 };
 use futures::StreamExt;
 use ratatui::{Terminal, prelude::CrosstermBackend};
-use tokio::time::interval;
+use tokio::time::{MissedTickBehavior, interval};
 
 use app::App;
 use ratatui::style::Color as RatatuiColor;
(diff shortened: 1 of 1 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1339, opened 2026-02-20
fix: Skip channel leads in session recovery dispatch loop [Midtown !1630]

btucker:
## Summary

- `dispatch_via_sessions_with_task_lookup` in `dispatch.rs` iterated `in_progress_tasks` and resumed stopped sessions as regular coworkers via `LaunchConfig::coworker()`, without checking if the owner is a channel lead
- Added a guard at the top of the session recovery loop to skip channel leads (and empty/lead owners), matching the pattern used in `check_for_duplicate_task_workers` and `decide_orphan_recovery`
- Identified in amsterdam's review note on PR #1337

## Test plan

- [x] New test `test_session_dispatch_skips_channel_lead_owned_tasks` confirms channel lead tasks are not recovered as coworkers
- [x] All 7 `dispatch_session_tests` pass
- [x] Full test suite passes with no failures
- [x] Clippy clean

🌃 Co-built with [Midtown](https://github.com/btucker/midtown)

PR #1394, opened 2026-02-20
fix: Skip task recovery in dispatch_via_sessions when coworker is reviewing [Midtown !1682]

btucker:
## Summary

- **Bug**: Daemon could reassign a coworker to a task while they were mid-review, interrupting the review
- **Root cause**: `dispatch_via_sessions` checked if the task's specific session was running (`record.is_running`), but not if the coworker was running in another capacity (as a reviewer)
- **Fix**: Added an `active_reviewers` check before recovering stopped task sessions — if the coworker is an active reviewer, skip the recovery

## Root cause details

A coworker can have multiple session records: a stopped dev session from previous task work, and a separate running reviewer session. The `dispatch_via_sessions` loop found the stopped task session and tried to resume it, even though the coworker was already busy reviewing a PR.

This happened with PR #1384: columbus was called in as a reviewer at 21:11, posted a "Review in progress" comment, then at 21:20 the daemon resumed columbus's stopped session for task !1675 — the review was never completed.

## Changes

- `src/daemon/dispatch.rs`: Added reviewer check after resolving `coworker_name` in `dispatch_via_sessions_with_task_lookup`
- `src/daemon/dispatch_session_tests.rs`: Regression test covering the exact scenario

## Test plan

- [x] `test_session_dispatch_skips_active_reviewer` — new regression test passes
- [x] `cargo test` — all existing tests pass

### DIFF
--- a/src/daemon/dispatch.rs
+++ b/src/daemon/dispatch.rs
@@ -343,7 +343,8 @@ fn should_recover_task(
 ///
 /// Rate limiting: Only spawns ONE coworker per tick with a cooldown between
 /// spawns to prevent window flashing from spawn storms.
-// Legacy: kept for existing tests. Will be removed in cleanup task.
+// Backward-compat test infrastructure: exercises the original orphan recovery
+// path (case 3 only) independently of the full session-aware dispatch.
 #[cfg(test)]
 #[allow(dead_code)]
 pub fn check_and_recover_orphans(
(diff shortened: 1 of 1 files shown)


btucker:
Both items addressed in aed450e:

1. **Doc comment** — Updated case 2 description to document the active reviewer skip, matching how the channel lead pre-filter was documented in PR #1339.

2. **Legacy comments** — Replaced stale "Legacy: will be removed in cleanup task" with "Backward-compat test infrastructure" to accurately reflect that these functions serve as test infrastructure for the original orphan recovery path.

PR #1353, opened 2026-02-20
fix: Detect and respawn reviewers that exit before posting review [Midtown !1652]

btucker:
## Summary

Three reviewers (lexington on PR #1349, columbus on PR #1351, riverside on PR #1352) consecutively went idle without posting their final review comments. This PR fixes two gaps in the daemon's reviewer lifecycle management.

**Root cause — two paths where reviewers exit without posting:**

1. **Dead process gap**: `decide_stuck_reviewer_restarts` exempts dead processes (`!health.is_alive`). When a reviewer's Claude Code session exits naturally (max turns, rate limit, context window exhausted), the process dies cleanly and this exemption prevents detection. The reviewer assignment persists but the review never gets posted.

2. **Idle RPC gap**: When a reviewer called `midtown state idle`, the RPC handler immediately shut them down without checking whether the review was posted.

## Changes

- **`src/rules.rs`**: Add `decide_dead_reviewer_respawns` — pure function that detects reviewers whose process died without posting a review and returns respawn decisions (up to `MAX_REVIEWER_RESTARTS` per PR). Four new unit tests.
- **`src/daemon/health.rs`**: Add `check_and_restart_dead_reviewers` — effect generator that calls the pure function and produces respawn effects, mirroring `check_and_restart_stuck_reviewers`.
- **`src/daemon/events.rs`**: Wire `check_and_restart_dead_reviewers` into `SessionMonitorTick`.
- **`src/daemon/rpc_coworker.rs`**: Guard the idle RPC handler — before allowing a reviewer to go idle, verify the review was posted. If not, nudge them to post it first.
- **`tests/fixtures/snapshot/`**: Add captured snapshot `snapshot-reviewer-stuck-before-posting-review-20260220-054127.json`.

## Test plan

- [x] 4 new unit tests for `decide_dead_reviewer_respawns` (dead reviewer without review detected, with review posted skipped, at max restarts skipped, alive reviewer not handled)
- [x] 1706 existing tests pass, 0 failures
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo fmt -- --check` clean

🌃 Co-built with [Midtown](https://github.com/btucker/midtown)

PR #1395, opened 2026-02-20
fix: build reviewer_pr_assignments from persistent state to include dead reviewers [Midtown !1684]

btucker:
## Summary

- `reviewer_pr_assignments` in `snapshot.rs` was built by iterating `active_coworkers` and looking each up in `persistent_state.github.pr_reviewers`
- When a reviewer's process died, it was removed from `active_coworkers`, dropping its entry from `reviewer_pr_assignments`
- This caused `decide_dead_reviewer_respawns` to never fire for dead reviewers — a reviewer that exited before posting its review would go undetected and unrecovered
- The escalation logic in `check_and_restart_stuck_reviewers` also uses this map and would miss dead reviewers

The fix extracts the map-building into a pure helper `build_reviewer_pr_assignments()` that reads from `pr_reviewers` directly (which persists across coworker lifecycle changes). Dead reviewers now remain in the map.

This bug was flagged on PR #1353 and PR #1392.

## Test plan

- [x] Added regression test `reviewer_pr_assignments_includes_dead_reviewers` in `snapshot_tests.rs` that verifies dead reviewers (absent from `active_coworkers`) appear in `reviewer_pr_assignments`
- [x] Format check passes (`cargo fmt --all -- --check`)

### DIFF
--- a/src/daemon/health.rs
+++ b/src/daemon/health.rs
@@ -532,6 +532,7 @@ pub fn check_and_restart_dead_reviewers(snap: &snapshot::WorldSnapshot) -> Vec<E
         &snap.reviewer_restart_counts,
         MAX_REVIEWER_RESTARTS,
         &snap.name_session_map,
+        &snap.usage_limited_coworkers,
     );
 
     let escalations = crate::rules::decide_dead_reviewer_escalations(
--- a/src/daemon/health_tests.rs
+++ b/src/daemon/health_tests.rs
@@ -2031,3 +2031,68 @@ fn dead_reviewer_escalation_not_repeated_after_recorded() {
         effects
     );
 }
+
+/// `check_and_restart_dead_reviewers` must produce both a respawn and an
+/// escalation in the same tick when two dead reviewers are present: one below
+/// max restarts (→ respawn) and one at max restarts (→ escalation).
+///
+/// This covers the combined-effects code path where the early-return guard
+/// checks both respawns and escalations.
+#[test]
+fn check_and_restart_dead_reviewers_emits_respawn_and_escalation_in_same_tick() {
+    let mut snap = empty_snap();
+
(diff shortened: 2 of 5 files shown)


btucker:
Addressing all 3 issues:

1. **Timeout filter in `build_reviewer_pr_assignments`** — verified, will add the same `PR_REVIEW_ASSIGNMENT_TIMEOUT_SECS` filter that `active_reviewers()` uses.
2. **Rate-limited reviewer respawn** — verified, `decide_dead_reviewer_respawns` needs a `usage_limited_coworkers` check to skip rate-limited dead reviewers.
3. **Combined respawn+escalation test** — will add.

btucker:
Addressed all 3 review issues in commit aedd357:

1. **Timeout filter in `build_reviewer_pr_assignments`** — Added the same `PR_REVIEW_ASSIGNMENT_TIMEOUT_SECS` filter that `active_reviewers()` uses, so expired assignments are excluded consistently. Added `build_reviewer_pr_assignments_excludes_expired_entries` test that verifies both functions exclude expired entries the same way.

2. **Rate-limited dead reviewer skip** — Added `usage_limited_coworkers` parameter to `decide_dead_reviewer_respawns` and skip logic (mirrors `StuckExemptions` in stuck detection). Added `dead_reviewer_not_respawned_when_usage_limited` test.

3. **Combined respawn+escalation test** — Added escalation logic to `check_and_restart_dead_reviewers` for dead reviewers at max restarts (analogous to `check_and_restart_stuck_reviewers`'s escalation path). Updated early-return to `if respawns.is_empty() && escalations.is_empty()`. Added `check_and_restart_dead_reviewers_emits_respawn_and_escalation_in_same_tick` test with two dead reviewers (one below max → respawn, one at max → escalation).

PR #1406, opened 2026-02-21
fix: Move 'Spawning isolated coworker' debug log below collision guard [Midtown !1692]

btucker:
## Summary
- Moves the `debug!("Spawning isolated coworker...")` log from before the worktree collision guard to after it
- The log previously fired even when spawning was aborted by the collision guard, causing misleading log output
- Now the log only fires when a spawn actually proceeds past all guards

## Test plan
- [x] Cargo build passes
- [x] All unit tests pass
- [x] No clippy warnings

### DIFF
--- a/src/daemon/pr.rs
+++ b/src/daemon/pr.rs
@@ -2385,13 +2385,6 @@ pub(crate) async fn collect_reviewer_effects_with_source(
             continue;
         }
 
-        let title = pr.get("title").and_then(|s| s.as_str()).unwrap_or("");
-        debug!(
-            "Spawning isolated coworker to review PR #{}: {}",
-            pr_number,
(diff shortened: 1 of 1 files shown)


btucker:
### Code review

Verified:
- The `debug!` log is correctly moved from before the coworker limit/name-availability/collision guards (line 2388) to after all guards (line 2440), right before the actual spawn effects
- All variables (`pr`, `pr_number`, `title`, `truncate_str`) are in scope at the new location
- Git history confirms the log's original position was an inadvertent leftover from before the collision guard was added in `e8e2e97`
- No CLAUDE.md violations — purity constraints don't apply to this async effects-collector function
- Task !1692 requirements fully met

PR #1410, opened 2026-02-21
fix: Prevent test-created channels from leaking to real project directory [Midtown !1696]

btucker:
## Summary

Tests calling `create_channel()` via `handle_event()` were writing real channel directories to `~/.midtown/projects/<repo>/channels/` because `create_channel()` did not check `test_mode` before performing filesystem operations. The daemon then found these directories on startup and spawned channel leads for them — which is how #test-channel, #new-channel, and #main showed up as active channels with idle lead sessions.

**Root causes identified:**

1. **#test-channel and #new-channel**: Created by `test_channel_create_command_clears_input` and `test_channel_create_command_switches_channel` — both call `handle_event()` with `/channel create` commands, which trigger `create_channel()` which calls `Channel::new()` on the real project path
2. **#main**: Created from an old code path that used "main" as the default/fallback channel name before the rename to "midtown". Once the directory existed, the daemon's `recover_channel_lead_sessions_from()` spawned a channel lead for it (the filter only excluded "midtown", not "main")

**Fixes:**

- `create_channel()`: Skip filesystem ops when `test_mode` is true; just switch `selected_channel` and return `true`. Existing tests that verify input clearing and channel switching continue passing without disk writes.
- `load_channel_messages()`: Return early when `test_mode` is true (after setting `load_channel_messages_called` for test observability). Defense-in-depth for other callers triggered in tests.
- `recover_channel_lead_sessions_from()`: Also exclude `"main"` from the topic channel list. The TUI already excludes "main" from topic channel behavior — the daemon now matches this.
- Archived the spurious channels (`#main`, `#new-channel`, `#test-channel`, `#test-no-disk-write`) that had accumulated in the real project directory.

## Test plan

- [x] New regression test `test_channel_create_does_not_write_to_filesystem_in_test_mode` confirms `create_channel()` does not call `load_channel_messages()` in test mode (which would trigger filesystem writes)
- [x] All existing channel create tests still pass (input clearing, channel switching, path traversal rejection)
- [x] Full test suite passes: 1718 lib tests + 584 bin tests with 0 failures
- [x] Spurious channels archived via `midtown channel archive` (daemon state cleaned up)

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -100,7 +100,7 @@ When the daemon starts, it executes a careful cleanup and recovery sequence in `
 
 7. **Stale flag cleanup** — `clear_stale_running_sessions()` clears the `is_running` flag for any session not included in the recovered set. This covers sessions skipped by `recover_from_session_records` for any reason (non-resumable, reviewer without a PR number, or dropped by name deduplication), as well as channel-lead sessions whose channel was archived between daemon runs. Active channel-lead sessions (whose channel is still non-archived) are preserved for the separate channel lead recovery path.
 
-8. **Channel lead recovery** — `recover_channel_lead_sessions()` iterates active (non-archived) topic channels and emits `SpawnCoworker` effects to resume or fresh-start each channel lead session.
+8. **Channel lead recovery** — `recover_channel_lead_sessions()` iterates active (non-archived) topic channels and emits `SpawnCoworker` effects to resume or fresh-start each channel lead session. The main channel names `"midtown"` and `"main"` are always excluded from this step — they belong to the Project Lead, not a channel lead. This guards against accidentally recreated channel directories (e.g., from tests or legacy TUI sessions).
 
 ## Coworkers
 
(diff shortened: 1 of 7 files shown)


btucker:
## Review Feedback Addressed

All 6 issues from the review have been fixed in 93752c4:

1. **rpc_auth.rs dead else branch** — Simplified to always update both `auth_provider` and `model`, removing the conditional that caused the else branch to be a no-op.

2. **chat.rs multi-mention routing** — Added `task_rerouted` flag set to `true` after the first successful reroute. Subsequent mentions in the same message route by name as normal.

3. **Inline test moved to separate file** — `test_channel_create_does_not_write_to_filesystem_in_test_mode` is now in `channel_create_tests.rs`, following the convention established in CLAUDE.md and the rest of the `chat/` directory.

4. **docs/architecture.md** — Added Provider Resolution section documenting the full resolution chain for each role, and updated step 8 in the Daemon Startup Sequence to note the main channel exclusion.

5. **rpc_session.rs provider=None** — Added `coworker_role_to_execution_role()` helper and updated both `handle_session_detach` and `handle_session_clear` to fall back to project config when `session_info.provider` is None.

6. **startup.rs main channel exclusion** — Added `test_recover_channel_lead_sessions_only_main_excluded` test alongside the existing `midtown` exclusion test.

PR #1411, opened 2026-02-21
fix: Clear stale session ID after spawn to prevent recovery loop [Midtown !1693]

btucker:
## Summary

- Fixed infinite recovery loop in `dispatch_via_sessions` where a stopped session was recovered on every daemon tick (~5s)
- Root cause: `spawn_coworker` used `or_insert_with` to update `persistent_state.sessions`, which skips existing entries — leaving `is_running: false` unchanged for stopped sessions
- Fix: Added `and_modify(|r| r.is_running = true)` before `or_insert_with` so existing stopped session records are immediately marked running on resume
- Added regression test reproducing the exact loop scenario from the captured snapshot (task !1690, riverside, reviewer session `e2bafbb6`)

## Root Cause

In `spawn_coworker` (`src/daemon/mod.rs`), when resuming a stopped session:

```rust
// Before (bug):
ps.sessions
    .entry(session_id.clone())
    .or_insert_with(|| { ... is_running: true ... });
// ↑ Entry exists → closure never called → is_running stays false

// After (fix):
ps.sessions
    .entry(session_id.clone())
    .and_modify(|r| { r.is_running = true; })  // ← added
    .or_insert_with(|| { ... is_running: true ... });
// ↑ and_modify runs on existing entry → is_running set to true
```

The next snapshot built from `persistent_state.sessions` sees `is_running: true`, so `dispatch_via_sessions` correctly skips recovery.

## Test plan

- [x] New regression test: `test_session_dispatch_recovery_loop_stopped_reviewer_session` in `dispatch_session_tests.rs`
  - Shows first recovery tick fires correctly (expected)
  - Shows second tick with `is_running=true` does NOT fire (preventing loop)
- [x] `cargo test` — all tests pass
- [x] `cargo clippy` — no warnings

### DIFF
--- a/src/daemon/state.rs
+++ b/src/daemon/state.rs
@@ -317,12 +317,15 @@ impl DaemonPersistentState {
     ///
     /// When resuming a stopped session, `entry().or_insert_with()` alone won't update
     /// `is_running` because the entry already exists. This method uses `and_modify` to
-    /// mark existing sessions as running before falling back to insert for new sessions.
+    /// mark existing sessions as running and refresh `current_name` before falling back
+    /// to insert for new sessions.
     pub fn upsert_session_running(&mut self, session_id: String, new_record: SessionRecord) {
+        let current_name = new_record.current_name.clone();
         self.sessions
             .entry(session_id)
             .and_modify(|r| {
                 r.is_running = true;
+                r.current_name = current_name;
             })
             .or_insert(new_record);
     }
--- a/src/daemon/state_tests.rs
+++ b/src/daemon/state_tests.rs
@@ -468,11 +468,15 @@ fn test_channel_lead_session_info_coworker_type_and_channel() {
 }
 
 fn make_test_session_record(session_id: &str, is_running: bool) -> SessionRecord {
+    make_test_session_record_named(session_id, is_running, "riverside")
+}
+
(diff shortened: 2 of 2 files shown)


btucker:
## Review Status

---
> [\!NOTE]

btucker:
Addressed the review feedback — thanks york.

**Issue 1 (stale branch):** Already rebased; `origin/main` (including PR #1407's fix) is in the base. The branch now shows only 2 commits ahead of main.

**Issue 2 (current_name dropped):** Fixed. The `and_modify` closure in `upsert_session_running` now also sets `r.current_name = current_name`. Sessions stopped via `handle_session_stopped` have `current_name` cleared to `None`; without this, name-indexed lookups would silently fail on resume.

**Issue 3 (test doesn't exercise the fix):** Added `test_upsert_session_running_updates_current_name_on_existing_entry` in `state_tests.rs`. The test pre-clears `current_name` on a stopped session record, calls `upsert_session_running`, and asserts `current_name` is restored. Removing the `and_modify` clause causes this test to fail.

**Issue 4 (missing active_session_ids layer):** Confirmed present via `dispatch.rs` in main — inherited from #1407.

Pushed as `6db9659`.

PR #1420, opened 2026-02-21
feat: Implement forked-session channel leads [Midtown !1701]

btucker:
## Summary

Implements three interlocking mechanisms for forked-session channel leads — letting a channel lead fork into a thread-specific session for context-isolated topic handling:

- **Core infrastructure**: `SessionRecord.bound_thread_id`, `HeadlessConfig.fork_session` flag + `--fork-session` CLI arg, `DaemonState.topic_sessions` map, `session.fork` RPC handler, `SessionManager::spawn_fork` (waits for `system/init` to capture session ID), and `midtown session fork <thread_id>` CLI command
- **Thread-aware routing**: In `handle_channel_post`, when a user posts a thread reply in a topic channel, checks `topic_sessions[thread_parent_id]` — routes to the fork via `NudgeSession` if found, falls through to `NudgeChannelLead` if not
- **Session-level output binding**: Before constructing the `Message`, looks up the poster's `SessionRecord.bound_thread_id`; if set and no explicit `thread_parent_id` was passed, auto-injects it — fork sessions post into their thread without needing `--thread` on every `midtown channel post`

## Test plan

- [x] `test_output_binding_auto_tags_forked_session_posts` — bound session posts appear in correct thread
- [x] `test_output_binding_explicit_thread_takes_priority` — explicit `thread_parent_id` overrides binding
- [x] `test_output_binding_no_bound_thread_preserves_none` — unbound sessions unaffected
- [x] `test_thread_routing_with_topic_session_routes_to_fork` — fork path taken when topic session registered
- [x] `test_topic_sessions_dedup_returns_existing` — idempotency guard behavior verified
- [x] `test_claude_headless_args_fork_session_adds_flag` — `--fork-session` added for resume+fork configs
- [x] `test_claude_headless_args_no_fork_session_without_flag` — `--fork-session` absent when flag is false
- [x] `test_claude_headless_args_fork_session_ignored_for_fresh` — `--fork-session` never added for fresh sessions
- [x] All existing tests pass (`cargo clippy --all-targets --all-features -- -D warnings` clean)

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ The Project Lead is just a Claude Code session, but it's been booted with a [spe
 | `midtown session attach name/<n>` | Attach to a headless session |
 | `midtown session detach <name>` | Resume headless execution |
 | `midtown session clear <lookup>` | Stop and restart a session fresh (preserves original task prompt) |
+| `midtown session fork --thread <id>` | Fork the calling session into a thread-bound session |
 | `midtown task create <subject> [...]` | Create a task (see [CLI reference](docs/cli.md) for all options) |
 | `midtown task list` | List tasks |
 | `midtown task view <id>` | View task details |
@@ -468,6 +469,8 @@ Each topic channel can have a **channel lead** — a headless Claude Code sessio
 - Receive user messages posted to their topic channel
 
 **What channel leads don't do:** Channel leads don't write code, open PRs, or create tasks. When implementation work is needed, they escalate to @lead.
+
+**Forked sessions:** Channel leads can fork themselves into thread-specific sessions via `midtown session fork --thread <id>`. A forked session inherits the parent's conversation context but gets an independent session ID bound to a specific thread. Thread replies are automatically routed to the fork, and the fork's channel posts are auto-tagged with the bound thread ID.
 
 Coworkers use `@{channel-name}` for domain questions (e.g., `@auth-refactor can you explain the token expiry logic?`) and reserve `@lead` for coordination and priority questions.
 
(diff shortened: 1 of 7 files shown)


btucker:
## Review Status

---
> [\!NOTE]

btucker:
Have we covered how nudging work with these forked sessions? my proposal is the forked session should get nudged by anything that spawns from it. so if the forked session creates a task, the fork session is who should get updates about it from the coworkers who pick it up. this implies we need to add to to `task create` a --thread-id, which should be the message-id of the thread where this task should be coordinated

btucker:
## Review Feedback Addressed

All 8 issues from lexington's review have been addressed in commit 02aa22d:

**Issues 1, 2, 5 (root cause: init event consumed by spawn_fork):**
- `handle_session_fork` now creates the `SessionRecord` directly after `spawn_fork` returns
- Populates `name_to_session` / `session_to_name` reverse maps explicitly
- `bound_thread_id` is set on the newly-created record (not looked up from a nonexistent one)

**Issue 3 (race condition):**
- Atomic check-and-reserve: insert "pending" sentinel into `topic_sessions` before calling `spawn_fork`, remove on failure. Two concurrent requests for the same thread_parent_id → second one hits the guard and returns `already_exists: true`.

**Issue 4 (fork sessions not excluded from idle-shutdown):**
- Added architectural invariant documentation on `spawn_fork` — fork sessions must NOT be registered in `CoworkerManager`.

**Issue 7 (test anti-pattern):**
- Rewrote `test_topic_sessions_dedup_returns_existing` to call `handle_session_fork` directly and assert the guard returns `already_exists: true`.

**Issue 8 (async lock on hot path):**
- Introduced `fork_bound_threads: Mutex<HashMap<String, String>>` (sync) in DaemonState. Output binding path in `handle_channel_post` uses this cache instead of acquiring the async `persistent_state` lock. Only fork sessions are in this map, so normal coworker posts hit zero locks.

**Issue 6 (docs):**
- Updated `docs/architecture.md` with "Forked Sessions" subsection under Channel Leads
- Updated `README.md` with `midtown session fork` command and fork description

**Follow-up:** Created task for `--session-id` suggestion (controlling fork session ID upfront).

PR #1413, opened 2026-02-21
docs: Add --channel option to task creation instructions

btucker:
## Summary

Adds documentation for the `--channel` option when creating tasks:

```bash
midtown task create "Fix auth bug" --description "..." --channel auth
```

This routes the coworker's messages to the right channel and lets the channel lead track the work.

## Changes

- `agents/project-lead.md` - Add --channel to task creation examples
- `agents/lead.md` - Add --channel to Task Routing section

## Test plan

- [ ] Verify documentation is clear and examples are correct

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1421, opened 2026-02-21
docs: Clarify thread reply instructions to prevent duplicate messages [Midtown !1704]

btucker:
## Summary
- Adds explicit guidance to `agents/lead.md` that when replying in a thread, the lead should not also write a top-level text response covering the same content
- Top-level text output is auto-posted to the channel, so a thread reply + top-level text produces duplicate messages visible to the user
- Also updates the role description to clarify the user is first priority and the lead should respond concisely before executing tools

## Problem
The lead was replying in threads (via `midtown channel post --thread`) but also writing top-level explanatory text in its response, which auto-posted to the channel as a separate top-level message. Users saw the same information twice — once in the thread, once at the top level.

## Fix
Added a clear warning in the threading section:

> **Important**: Your text output is auto-posted to the channel as a top-level message. When replying in a thread, use `midtown channel post --thread` and **do not also write a top-level text response about the same thing**.

## Test plan
- [x] Verify `agents/lead.md` diff looks correct
- [x] No code changes — documentation/instruction fix only

### DIFF
--- a/agents/channel-lead.md
+++ b/agents/channel-lead.md
@@ -68,7 +68,7 @@ When replying in a thread (nudges include the message ID in the format `sender (
 midtown channel post "reply text" --thread <message-id> --channel {channel_name}
 ```
 
-**Always reply in a thread** when responding to user messages or @mentions — this keeps the channel organized.
+**Always reply in a thread** when responding to user messages or @mentions — this keeps the channel organized. Note: your text output is still auto-posted as a top-level message, so writing text alongside a `--thread` reply produces a duplicate. Keep your text output brief or omit it when the thread reply covers everything.
 
 ## Awareness
 
--- a/agents/lead.md
+++ b/agents/lead.md
@@ -28,7 +28,9 @@ midtown channel post "Your reply" --thread <message-id>
 
 This keeps the channel organized — top-level posts start conversations, replies continue them. If you don't have a message ID (e.g., daemon-generated nudges), post at the top level as usual.
 
-**Important**: Your text output is auto-posted to the channel as a top-level message. When replying in a thread, use `midtown channel post --thread` and **do not also write a top-level text response about the same thing**. Otherwise the user sees a duplicate — once in the thread, once at the top level. Keep your text output brief (e.g., status notes unrelated to the thread) or omit it entirely when the thread reply covers everything.
+<EXTREMELY_IMPORTANT>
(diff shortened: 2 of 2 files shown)


btucker:
### Code review

Found 2 issues:

1. **Structural contradiction between Thread Replies instructions and new warning** -- The "Thread Replies" section (line 23) instructs the lead to "Always reply in a thread" using `midtown channel post --thread`, while the new warning (line 31) tells the lead not to also write top-level text. These two instructions create a tension: the lead must use the CLI tool for thread replies (producing a tool call), but any accompanying text output is auto-posted as a top-level message. The warning patches the symptom rather than resolving the underlying contradiction. Historical context: commit `baa5b02` (PR #1223) deliberately removed explicit `channel post` instructions to avoid this exact duplicate-message pattern. Consider restructuring to make the interaction between auto-posting and thread replies explicit in the section itself (e.g., noting that thread replies are the one case where the lead must suppress or minimize text output).

   Test suggestion: N/A (documentation/instruction issue)

   https://github.com/btucker/midtown/blob/682a88a5a717f81ce4b0596bbfd639e69bf4b327/agents/lead.md#L21-L31

2. **Warning missing from `channel-lead.md` threading section** -- The duplicate-reply warning was added to `lead.md`, but `channel-lead.md` has its own parallel threading section (lines 65-71) that says "Always reply in a thread when responding to user messages or @mentions" with no equivalent caution about auto-posted text output creating duplicates. Channel leads assemble their prompt from `channel-lead.md + lead.md`, so they will eventually see the warning -- but `channel-lead.md`'s own threading section is incomplete without it. The review of PR #1413 established the precedent that guidance relevant to channel leads should appear in the channel-lead-specific sections, not just in the shared base.

   Test suggestion: N/A (documentation/instruction issue)

   https://github.com/btucker/midtown/blob/682a88a5a717f81ce4b0596bbfd639e69bf4b327/agents/channel-lead.md#L65-L71

btucker:
Addressed both review items in 0815ac0:

1. **Structural contradiction** — Restructured the Thread Replies section to integrate the auto-posting caveat directly into the flow using `<EXTREMELY_IMPORTANT>` tags (matching the file's existing emphasis style) instead of a separate `**Important**` callout. The warning now reads as a natural consequence of the preceding instructions rather than a contradicting afterthought.

2. **Missing warning in channel-lead.md** — Added the same auto-posting duplicate warning to `channel-lead.md`'s threading section (lines 71), per the PR #1413 precedent that guidance relevant to channel leads should appear in their own sections.

PR #1424, opened 2026-02-21
feat: Add channel rename command [Midtown !1708]

btucker:
## Summary

- Adds `midtown channel rename <old> <new>` CLI command to rename topic channels
- Implements the full RPC stack: `Channel::rename_channel()` (filesystem), `channel.rename` handler, CLI subcommand, and client method
- Updates all daemon in-memory state on rename: `task_channel` values, `channel_lead_sessions` key, and `headless_sessions` entry for the channel lead
- Shuts down the old channel lead session (will be respawned fresh under the new name when triggered)
- Evicts the stale ChannelRouter cache entry to prevent writes to the old path
- Guards: cannot rename 'midtown', cannot rename to an existing channel name, validates new name format

## Test plan

- [x] 6 unit tests: success path (dir move + state update), nonexistent source, target already exists, 'midtown' forbidden, invalid new name, router cache eviction
- [x] `cargo fmt` clean

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -584,9 +584,27 @@ impl Channel {
             .into());
         }
 
+        // Validate the old name to prevent path traversal (e.g., "../worktrees").
+        // The new name is validated below, but the old name must also be safe since
+        // it's used in Path::join("channels", old).
+        if !Self::is_valid_channel_name(old) {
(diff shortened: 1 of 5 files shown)


btucker:
Addressing all 8 review items now. Will push fixes shortly.

PR #1427, opened 2026-02-22
fix: Prevent repeated session recovery log spam via per-session cooldown [Midtown !1709]

btucker:
## Summary

- Adds a per-session-id `session_recovered` cooldown (30s) that fires after each successful recovery spawn
- Adds `recently_recovered_session_ids: HashSet<String>` to `WorldSnapshot`, populated from cooldowns during snapshot collection
- Guards `dispatch_via_sessions_with_task_lookup` to skip recovery if the session is in `recently_recovered_session_ids`

## Root Cause

When `dispatch_via_sessions_with_task_lookup` successfully recovers a session, it records `SESSION_DISPATCH_COOLDOWN` (2s). The tick runs every 5s (`ORPHAN_CHECK_INTERVAL_SECS`). If the session dies within that window, the next tick sees `is_running=false` and an empty `active_session_ids`, and the 2s cooldown has already expired — so recovery fires again, posting "Session dispatch: recovered task !N" on every tick.

## Fix

A per-session-id "session_recovered" cooldown (30s) is recorded in the `on_success` callback of each recovery spawn. The snapshot pre-computes `recently_recovered_session_ids` by checking this cooldown for all known session IDs. The dispatch function skips recovery if the session is in that set, stopping the log spam.

The 30s window is chosen to exceed `ORPHAN_CHECK_INTERVAL_SECS` (5s) comfortably while staying within `was_failed_resume`'s 30s window, so failed-resume detection can clear the stale session before the cooldown expires.

## Test plan

- [x] `test_session_dispatch_skips_recovery_for_recently_recovered_session` — guards the skip path
- [x] `test_session_dispatch_on_success_includes_session_recovered_cooldown` — verifies the cooldown effect is emitted on success
- [x] All 13 `dispatch_session` tests pass

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -119,7 +119,7 @@ The daemon uses a **session-centric model** where Claude Code sessions (keyed by
 
 **SessionRecord** (`src/daemon/state.rs`): Each session is tracked by a `SessionRecord` containing session ID, task ID, current and preferred names, worktree path, branch, PR number, and running state. Records persist across daemon restarts in `persistent_state.json`.
 
-**Dispatch** (`src/daemon/dispatch.rs`): `dispatch_via_sessions()` is a pure function that examines in-progress tasks with session records. For stopped sessions, it emits `SpawnSession` effects with `resume=true` and the session's preferred name. This replaces the legacy orphan-recovery pattern with a unified session-aware dispatch path.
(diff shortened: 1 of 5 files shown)


btucker:
## Review Status

---
> [\!NOTE]

btucker:
Addressing all 4 review items:

1. **architecture.md** — Adding `recently_recovered_session_ids` to WorldSnapshot fields section and dispatch behavior description
2. **Misleading 30s boundary comment** — Fixing the doc comment to accurately describe the simultaneous expiry (both use 30s, no actual safety margin — tick granularity provides practical protection)
3. **Stale dispatch_via_sessions doc comment** — Adding the recently-recovered cooldown as a second skip condition in Case 2
4. **Snapshot population test** — Adding a dedicated test for the CooldownTracker → `recently_recovered_session_ids` extraction logic

Also fixing the doc comment style on the snapshot field to include the "Pre-evaluated from `state.cooldowns`" pattern per madison's channel note.

PR #1434, opened 2026-02-22
fix: prevent idle shutdown while tools run [Midtown !1710]

btucker:
## Summary
- thread headless tool/subagent flags through `IdleShutdownContext` so coworkers with active tools are protected from idle shutdown
- feed the new context data from the health builder + channel lead tests and trim the idle-shutdown harness/test helpers accordingly
- keep reviewer assignments alive for 30 minutes so reviewers retain idle-shutdown protection during lengthy reviews

## Test plan
- cargo test idle_shutdown
- cargo clippy --all-targets --all-features

### DIFF
--- a/src/daemon/health.rs
+++ b/src/daemon/health.rs
@@ -65,19 +65,6 @@ pub fn check_and_shutdown_idle_coworkers(snap: &snapshot::WorldSnapshot) -> Vec<
             .join(", "),
     );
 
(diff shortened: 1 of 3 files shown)


btucker:
Thanks for the review and for jumping in when the reviewer stalled. I’m going to merge this now so lexington can finish !1711 on top of it; no additional changes needed here.

PR #1454, opened 2026-02-24
fix: exclude project lead from coworkers status sidebar [Midtown !1723]

btucker:
## Summary
- The project lead ('lead') was appearing in the coworkers status sidebar alongside actual coworkers
- Added a case-insensitive name filter in `draw_coworker_status` (board.rs:442) to exclude the lead, alongside the existing idle/done phase filter
- Added a regression test in `board_tests.rs` following the existing `test_coworker_line_map_excludes_idle` pattern

## Root cause
The `active_coworkers` filter in `draw_coworker_status` only excluded coworkers with `phase == "idle"` or `"done"`. Channel leads are excluded upstream by the daemon (not included in the `coworkers` JSON array), but the project lead is present in that array and was not filtered.

## Test plan
- [x] New test `test_coworker_line_map_excludes_project_lead` fails before fix, passes after
- [x] Existing idle exclusion test (`test_coworker_line_map_excludes_idle`) still passes
- [x] Full test suite: 628 passed, 0 failed

### DIFF
--- a/src/bin/midtown/cli/chat/ui/board.rs
+++ b/src/bin/midtown/cli/chat/ui/board.rs
@@ -28,11 +28,18 @@ pub fn draw_board_panel(f: &mut Frame, app: &mut App, area: Rect) -> (Vec<Hyperl
     app.channel_line_map.clear();
     app.coworker_line_map.clear();
 
-    // Split board area vertically: tasks at top, coworkers at bottom
+    // Split board area vertically: tasks at top, coworkers at bottom.
+    // Exclude the project lead (legacy "lead" or repo-named) to keep the height
+    // in sync with the rendering filter in draw_coworker_status.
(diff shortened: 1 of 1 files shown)


btucker:
## Review Status

---
> [\!NOTE]

btucker:
Addressed Vernon's review feedback:

1. **Layout height mismatch** — Added lead exclusion to the `active_coworker_count` filter in `draw_board_panel` to keep height in sync with `draw_coworker_status`. Added regression test `test_draw_board_panel_project_lead_does_not_inflate_height` (failing-first confirmed).

2. **Misleading comment** — Updated the comment to clarify that both project lead and channel leads are already excluded upstream by `build_coworkers_data`; the UI filter is now described as a defensive guard.

Both the legacy literal "lead" and the repo-named lead (via `app.project_name`) are excluded at all layers.

PR #1455, opened 2026-02-24
fix: protect channel leads from idle shutdown [Midtown !1721]

btucker:
## Summary

- `decide_idle_shutdowns()` in `rules.rs` had no exemption for channel leads, causing them to be periodically killed by the idle shutdown loop
- Added `channel_lead_names: &HashSet<String>` field to `IdleShutdownContext`
- Added a guard: if the coworker name is in `channel_lead_names`, skip idle shutdown
- In `health.rs`, derives `channel_lead_names` from `snap.channel_lead_sessions.keys()` (channel names == session names via `channel_lead_session_name()`)

## Root cause

Channel leads (e.g., the "ops" or "daemon" sessions) are long-running domain experts that maintain persistent context for their topic channels. Unlike ordinary coworkers, they should never be idle-shutdown. However, `IdleShutdownContext` had no `channel_lead_names` field, so `decide_idle_shutdowns()` treated them as regular coworkers eligible for shutdown.

The fix mirrors how channel leads are already exempted from orphan recovery (`OrphanRecoveryContext.channel_lead_names`) and task dispatch (`decide_pending_task_action` `is_channel_lead` param).

## Test plan

- [x] `channel_lead_is_not_idle_shutdown` — channel lead in `channel_lead_names` is never idle-shutdown
- [x] `idle_shutdown_protects_channel_leads` — builder-based test with `with_channel_leads()`
- [x] `idle_shutdown_shuts_down_non_channel_lead_with_same_name` — guard is keyed by `channel_lead_names` set, not name pattern alone
- [x] All 1825 lib tests pass, clippy clean

### DIFF
--- a/src/daemon/health.rs
+++ b/src/daemon/health.rs
@@ -85,8 +85,11 @@ pub fn check_and_shutdown_idle_coworkers(snap: &snapshot::WorldSnapshot) -> Vec<
 
     // Pure decision: who should be shut down?
     let to_shutdown = {
-        let channel_lead_names: HashSet<String> =
-            snap.channel_lead_sessions.keys().cloned().collect();
+        let channel_lead_names: HashSet<String> = snap
+            .channel_lead_sessions
+            .keys()
+            .map(|k| k.to_lowercase())
+            .collect();
         let idle_ctx = crate::rules::IdleShutdownContext {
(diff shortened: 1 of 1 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1460, opened 2026-02-24
feat: thread input bar expands to multiple lines [Midtown !1727]

btucker:
## Summary
- Made `calculate_input_bar_height` `pub(super)` in `chat.rs` so sibling UI modules can reuse it
- Updated `draw_thread_panel` in `thread.rs` to compute dynamic input height via `calculate_input_bar_height(&app.thread_input_text, area.width)` instead of the hardcoded `Constraint::Length(3)`
- Added a regression test (`test_thread_input_expands_to_multiple_lines`) verifying 150-char input expands the thread input to height 5 (3 content lines + 2 borders)

The thread input now grows to the same 1–6 content lines as the main channel input, using identical logic.

## Test plan
- [x] New test `test_thread_input_expands_to_multiple_lines` — fails before the fix, passes after
- [x] All existing thread panel tests pass (narrow terminal, cursor color)
- [x] All 165 UI tests pass (`cargo test --bin midtown -- cli::chat::ui`)

### DIFF
--- a/src/bin/midtown/cli/chat/ui/chat.rs
+++ b/src/bin/midtown/cli/chat/ui/chat.rs
@@ -439,7 +439,7 @@ fn draw_chat_messages(f: &mut Frame, app: &mut App, area: Rect) {
 /// Calculate the required height for the input bar based on wrapped text
 ///
 /// Returns total height including borders (2) and content lines (1 minimum, 6 maximum).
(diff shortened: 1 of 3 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1468, opened 2026-02-24
docs: document ClearSessionWorkingDir effect and stale working-dir fallback [Midtown !1740]

btucker:
## Summary

- Document stale `working_dir` validation in the Session-Centric Dispatch section: both dispatch paths now validate `record.working_dir.exists()` before using the path, fall back to a fresh worktree if stale, and emit `Effect::ClearSessionWorkingDir` to prevent retrying the dead path next tick
- Add `Effect::ClearSessionWorkingDir { session_id }` entry (analogous to `ClearSessionForTask` for task bindings)
- Update `MINIMUM_COWORKER_LIFETIME` note in Idle Shutdown Protections to 90s (was 60s), with rationale

These changes document the behavior added in PR #1465.

## Test plan

- [x] Documentation-only change; no code changes


btucker:
Addressed the Codex feedback: the previous version documented `ClearSessionWorkingDir` and dual-path validation from the draft PR #1465, but the code that actually merged (PR #1467) uses a different approach. Updated to reflect reality: only Path 2 validates `working_dir.exists()`, and stale paths are cleared implicitly on the next successful `SpawnSession` by writing the new path back to `record.working_dir` in `effects.rs`.

PR #1469, opened 2026-02-24
feat: create Python SDK for workflow scripts (sdk/python/midtown/) [Midtown !1737]

btucker:
## Summary

- Adds `sdk/python/midtown/__init__.py` with the `MidtownRPC` client and `run()` entry point for workflow scripts
- Adds `sdk/python/pyproject.toml` (`midtown-sdk` package, requires Python ≥ 3.9, zero runtime dependencies)
- Updates `.gitignore` with Python `__pycache__` / `*.pyc` entries

## What's in the SDK

**`MidtownRPC`** — JSON-RPC 2.0 client over a Unix socket:
- `post_to_channel(message, *, channel, sender, thread_parent_id)`
- `create_task(subject, *, description, channel, blocked_by, model)`
- `update_task(task_id, *, owner, status, description, blocked_by, channel, model, pr)`
- `complete_task(task_id)`
- `list_tasks()`
- `spawn_coworker(*, prompt, resume)`
- `nudge_coworker(name, message, *, sender)`

**`run(handler)`** — standard entry point that:
1. Parses `--event`, `--state`, `--socket` CLI args (as the daemon passes them)
2. Decodes the event JSON
3. Loads state from the state file (empty dict on first run)
4. Calls `handler(event, rpc, state)`
5. Atomically saves the mutated state back to disk

**`RpcError`** — raised when the daemon returns a JSON-RPC error object.

## Usage

```python
# workflow.py
from midtown import run, MidtownRPC

def handle(event: dict, rpc: MidtownRPC, state: dict) -> None:
    if event["type"] == "coworker.idle":
        rpc.post_to_channel(f"{event['coworker']} is idle")

if __name__ == "__main__":
    run(handle)
```

## Test plan

- [x] Module imports cleanly
- [x] `run()` parses CLI args, calls handler with correct event dict, persists mutated state
- [x] Existing state file is loaded before handler invocation
- [x] `MidtownRPC` sends correct JSON-RPC requests over a mock Unix socket
- [x] `RpcError` is raised on daemon error responses
- [x] State write is atomic (temp file + rename)
- [x] clippy and fmt checks pass (pre-commit hook)

### DIFF
--- a/sdk/python/midtown/__init__.py
+++ b/sdk/python/midtown/__init__.py
@@ -24,6 +24,7 @@ import os
 import socket
 import sys
 import tempfile
+import uuid
 from pathlib import Path
 from typing import Any, Callable
 
@@ -52,7 +53,6 @@ class MidtownRPC:
 
     def __init__(self, socket_path: str) -> None:
         self._socket_path = socket_path
-        self._next_id = 1
 
     # ------------------------------------------------------------------
     # Transport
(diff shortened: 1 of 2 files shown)


btucker (commented):
Addressed both items:

- **P1 (UUID IDs)**: Replaced the per-process integer counter with `uuid.uuid4()` strings. The comment in `_call` now explains why: the daemon's 60s ID-keyed cache makes a resetting counter a real collision risk across `uv run` invocations.
- **P2 (missing README)**: Removed the `readme = "README.md"` field from `pyproject.toml` — the file doesn't exist yet, so packaging flows would fail.

PR #1482, opened 2026-02-24
fix: suppress insight.report for channel lead sessions [Midtown !1747]

btucker:
## Summary

- Channel lead sessions auto-post all output to their channel already — the `insight.report` RPC path was redundant and caused insights to spill into the main channel
- Added an early-exit check in `handle_insight_report`: if the reporting agent is in `channel_lead_sessions`, return `posted: false / reason: channel_lead` immediately
- Added test `test_insight_suppressed_for_channel_lead` covering this case

## Root cause

`rpc_insight.rs` routed insights using only `headless_sessions` (coworker task-to-channel map). Channel leads are tracked in `channel_lead_sessions` instead, so they had no channel entry and fell through to the main channel default.

## Test plan

- [x] `cargo test test_insight_suppressed_for_channel_lead` passes
- [x] All existing insight tests pass
- [x] `cargo clippy` clean

### DIFF
--- a/src/daemon/rpc_insight.rs
+++ b/src/daemon/rpc_insight.rs
@@ -27,6 +27,25 @@ pub(super) async fn handle_insight_report(
     channel: Option<&str>,
     state: &DaemonState,
 ) -> Response {
+    // Channel leads auto-post all output to their channel already.
+    // Suppress insight.report from channel leads to avoid double-posting.
+    {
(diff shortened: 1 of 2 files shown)


btucker:
Closing as duplicate — PR #1483 implements the same fix for !1747 and is the task owner's PR.

PR #1487, opened 2026-02-24
perf: remove lock contention and cursor disk I/O from TUI message display

btucker:
Fixes !1748.

## Problem

Two issues caused a visible delay between `midtown channel post` completing and the message appearing in the TUI:

### 1. Race condition (up to 500ms blocking sleep)

`tailf` fires a kqueue event when bytes hit disk — but **before** the writer releases its exclusive `flock`. `read_since_cursor()` called `try_lock_shared()`, found the lock held, and slept 50ms per retry (up to 10 retries = 500ms total worst case).

The critical issue: this used `std::thread::sleep` inside the tokio single-thread runtime (`new_current_thread`). Blocking sleeps in this context monopolize the tokio thread — all timers, keyboard events, and redraws stall until the sleep ends. Even a single 50ms hit is noticeable.

### 2. Cursor disk I/O on every tailf tick

Each `refresh()` call did `Cursor::load_or_create()` (disk read) + `cursor.save()` (disk write) — even when no new messages arrived (the common case).

## Fix

**`src/channel.rs`**
- Add `Channel::read_messages_from_position(position: u64)` — reads forward from a byte offset, **no file lock acquired**. `O_APPEND` writes are atomic and we stop at the last complete `\n`, so concurrent reads are safe without locking.
- Rewrite `read_since_cursor()` to delegate to the above — removes all lock retry logic and the blocking sleep.
- Change `set_cursor_to_end()` to return `(u64, Option<String>)` so callers can initialise in-memory state without a second disk read.

**`src/bin/midtown/cli/chat/app.rs`**
- Add `cursor_position: u64` and `cursor_last_message_id: Option<String>` to `App` — the in-memory cursor cache.
- `refresh()` now calls `read_messages_from_position(self.cursor_position)` instead of `read_since_cursor()`.
- Cursor is captured from `set_cursor_to_end()` return value on channel open/switch.
- Add `save_cursor_to_disk()` — persists in-memory cursor on channel switch and app exit so unread counts remain accurate across sessions.

## Result

- **No more blocking sleep** in the hot path — messages appear immediately after the kqueue event fires.
- **No disk I/O on idle ticks** — the common case (no new messages) now does a single `open()` + `seek()` + `read()` with no cursor file involvement.
- All 1877 unit tests pass, clippy clean.

## Test plan

- [ ] Send a message with `midtown channel post "test"` and observe it appears in the TUI without the ~50-500ms delay
- [ ] Switch channels — cursor is saved to disk, unread counts update correctly
- [ ] Exit TUI — cursor persisted, next session starts with correct unread counts
- [ ] `cargo test` passes (1877 unit tests)
- [ ] `cargo clippy --all-targets --all-features -- -D warnings` passes

### DIFF
--- a/src/bin/midtown/cli/chat/app.rs
+++ b/src/bin/midtown/cli/chat/app.rs
@@ -714,34 +714,38 @@ impl App {
             // at the last complete newline, so reading is safe without locking.
             if let Ok((new_messages, new_pos, new_last_id)) =
                 channel.read_messages_from_position(self.cursor_position)
-                && !new_messages.is_empty()
             {
+                // Always advance the position — even past blank/malformed lines —
+                // so we don't re-parse them on the next tick. last_message_id is
+                // only updated when a valid message was actually parsed.
                 self.cursor_position = new_pos;
                 if let Some(id) = new_last_id {
                     self.cursor_last_message_id = Some(id);
                 }
 
-                let added = new_messages.len();
-                let was_at_bottom = self.scroll_offset == 0;
+                if !new_messages.is_empty() {
+                    let added = new_messages.len();
+                    let was_at_bottom = self.scroll_offset == 0;
 
-                // Route thread replies to thread_messages if a thread is open
-                if let Some(ref open_thread_id) = self.thread_parent_id {
-                    for msg in &new_messages {
-                        if msg.thread_parent_id.as_deref() == Some(open_thread_id) {
-                            self.thread_messages.push(msg.clone());
+                    // Route thread replies to thread_messages if a thread is open
+                    if let Some(ref open_thread_id) = self.thread_parent_id {
+                        for msg in &new_messages {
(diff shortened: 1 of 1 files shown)


btucker:
## Review Response (broadway)

Addressed both issues from the Codex review in commit c16307a:

**P1 — Cursor coherence (app.rs:720):** Fixed. `refresh_unread_counts()` now uses `cursor_last_message_id` (in-memory) for the currently open channel when it has been set, bypassing the stale on-disk cursor. When the in-memory cursor is None (channel just opened), falls through to the disk cursor as before.

**P2 — Blank-line advance (channel.rs:885):** Fixed. The read block now always advances `cursor_position` past blank/malformed lines (previously only advanced when `new_messages` was non-empty), preventing redundant re-parsing on the next tick. `cursor_last_message_id` is still only updated when a valid message ID is returned.

All 2513 tests pass, clippy clean.

PR #1516, opened 2026-02-24
fix(web): address 7 unreviewed celebration-effects items from PR #1504 [Midtown !1785]

btucker:
Closes Midtown task !1785. Addresses 7 of 8 items raised in Broadway's review of #1504 that were merged unfixed.

## Changes

1. **`hydrated = $state(false)`** — tells the Svelte compiler to generate getter/setter proxies so the `$effect` dependency-tracking is correct; plain `let` created a silent tracking gap  
2. **`pixel-burst` keyframe** — 0% now has `rotate(0deg) translateY(0px)` matching the 100% transform list; fixes discrete-jump fallback in CSS interpolation  
3. **daemon color documented** — `#585858` for daemon/github/system is intentional (infrastructure noise, not a human voice); was previously an undocumented side-effect of the avenue-colors.js extraction  
4. **TUI palette comment restored** — `// Muted avenue colors matching terminal TUI palette (AVENUE_COLORS from ui.rs)` comment lost during extraction is back  
5. **safe-area-inset padding** — `.celebration-layer` gains `env(safe-area-inset-*)` padding per CLAUDE.md PWA guidelines  
6. **post-teardown guard** — `removed = true` in `onDestroy`; `removeEffect` returns early if destroyed  
7. **`$activeProject` guard** — `<CelebrationEffects>` moved inside `{#if $activeProject}` to match all other interactive components  

## Not in this PR

**Item 8 (before/after screenshots)** requires running the app and triggering a celebration. Screenshots need to be captured manually and added to the original PR #1504 description.

### DIFF
--- a/agents/coworker.md
+++ b/agents/coworker.md
@@ -223,6 +223,38 @@ git push --force origin HEAD:$PR_BRANCH
 
 **Always include the task number in the PR title** using `[Midtown !XXX]` at the end. This makes it easy to trace PRs back to tasks.
 
+### Screenshots for Visual Changes
+
+When your PR includes visual changes to the web UI, include before/after screenshots. Take them with Playwright, then upload via the API:
+
+```bash
+# 1. Take a screenshot with Playwright
+node -e "
+const { chromium } = require('playwright');
+(async () => {
+  const b = await chromium.launch();
+  const p = await b.newPage();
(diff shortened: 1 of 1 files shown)


btucker:
Why is the instruction to coworkers to use the absolute path for images, should they use the path they get back from the upload endpoint?

PR #1527, opened 2026-02-25
fix: re-dispatch PR reviews when associated task is completed

btucker:
## Summary

- When a PR receives a review comment after its task is marked `done`, the daemon now creates a new follow-up task instead of trying to spawn/resume the original coworker with stale session context
- Added PR-based deduplication to `Effect::CreateTask` so multiple review comments don't generate duplicate follow-up tasks
- Both the webhook path (`handle_pr_comment_nudge`) and polling path (`collect_comment_notification_effects`) are fixed

## Root cause

When `midtown task done` completes a task, `pr_author_sessions` is retained in persistent state. So when a review comment arrives on the associated PR:

1. `handle_pr_comment_nudge` finds the task_id via `pr_author_sessions`
2. The existing status check only handled `InProgress` tasks — for `Completed` tasks it fell through
3. The code tried to spawn/resume the original coworker using a stale session ID
4. The spawn could fail silently (expired session), leaving the review unaddressed indefinitely

## Fix

Expanded the task-status branch in `handle_pr_comment_nudge` (and polling equivalent) to detect `Completed` tasks and emit `Effect::CreateTask` instead of spawning the original coworker. The new task is unassigned and picked up by normal dispatch on the next tick.

Added a new pure helper `review_comment_creates_followup(task_status)` in `rules.rs` (returns true for `Completed`) so the decision logic is testable and reusable.

## Test plan

- [x] `test_completed_task_requires_followup_task` — new test verifying the pure decision function returns true for Completed, false for InProgress/Pending
- [x] All existing `daemon::pr` tests pass (67 tests)
- [x] All existing `daemon::effects` tests pass (31 tests)  
- [x] All existing `rules` tests pass (163 tests)
- [x] `cargo fmt` clean

Closes !1794

### DIFF
--- a/src/daemon/effects.rs
+++ b/src/daemon/effects.rs
@@ -2171,6 +2171,24 @@ pub async fn execute_effects(effects: Vec<Effect>, state: &DaemonState) {
                 description,
                 pr,
             } => {
+                // If a PR number is associated, skip creation if a non-completed task
+                // already exists for that PR. This prevents duplicate follow-up tasks
+                // when multiple review comments arrive in quick succession (e.g., after
+                // a daemon restart resets the in-memory cooldown).
+                if let Some(pr_num) = pr {
+                    let existing = crate::tasks::read_tasks_for_repo(Some(&repo_name));
+                    let already_has_task = existing.iter().any(|t| {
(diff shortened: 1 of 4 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1538, opened 2026-02-25
fix: change return to continue in execute_effects BindCoworkerToWorktree collision guard [Midtown !1804]

btucker:
## Summary

- `BindCoworkerToWorktree` handler in `execute_effects` used `return` inside the `for effect in effects` loop when a worktree collision was detected (target worktree already bound to an active coworker). This exited the entire function, silently dropping all remaining effects.
- Fixed by changing `return` to `continue` with an explanatory comment.
- Added `test_is_alive_hook` to `SessionManager` following the existing `test_send_message_to_session_id_hook` pattern, enabling the `is_alive` path to be controlled in tests without spawning real processes.
- Added a batch-level regression test (`bind_coworker_to_worktree_collision_does_not_drop_subsequent_effects`) that directly reproduces the bug: collision guard fires on the first effect, but the subsequent `SaveChannelLeadSession` effect must still run.

Same bug class as the `CreateTask` dedup guard fixed in #1537.

## Test plan

- [x] New test `bind_coworker_to_worktree_collision_does_not_drop_subsequent_effects` fails before the fix (left: `None`) and passes after (`Some("sess-sentinel-99")`)
- [x] All 1938 unit tests pass
- [x] Clippy and fmt clean

### DIFF
--- a/src/daemon/effects.rs
+++ b/src/daemon/effects.rs
@@ -1774,8 +1774,10 @@ pub async fn execute_effects(effects: Vec<Effect>, state: &DaemonState) {
                                     "WORKTREE COLLISION BLOCKED: Refusing to bind {} to worktree {} - already bound to ACTIVE coworker {}",
                                     coworker, worktree_id, current_coworker
                                 );
-                                // Do NOT bind - this would crash both Claude Code sessions
-                                return;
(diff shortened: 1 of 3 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1544, opened 2026-02-25
fix(web): fix channel expand/collapse in sidebar [Midtown !1811]

btucker:
## Summary

Fixes two interacting bugs in `ChannelList.svelte` that made sidebar expand/collapse unreliable:

- **Bug 1:** Clicking the triangle (▶/▼) on an *inactive* channel only toggled task visibility without selecting the channel. The triangle and channel name button are siblings (not nested), so there was no propagation issue — `selectChannel()` was simply never called from the triangle handler.

- **Bug 2:** Clicking the channel name on the *already-active* channel never toggled the task list. `selectChannel()` had an early-exit guard (`channelName !== $activeChannel`) that skipped auto-expand with no toggle fallback.

## Changes

**`channelUtils.js`** — Two new exported pure functions:
- `computeExpandedAfterTriangleClick(channelName, expandedChannels)` — always toggles
- `computeExpandedAfterChannelNameClick(channelName, expandedChannels, activeChannel, kanban)` — auto-expands on channel switch, toggles on re-click of active channel

**`ChannelList.svelte`** — Fixes the event handlers:
- `handleTriangleClick()` captures desired toggle state *before* calling `selectChannel()`, then overrides expansion — undoing any auto-expand interference
- `selectChannel()` now uses `computeExpandedAfterChannelNameClick()` so re-clicking the active channel toggles tasks

**`channelUtils.test.js`** — 10 new unit tests covering:
- Expanding/collapsing via triangle
- Immutability of input set
- Auto-expand on channel switch
- Toggle behavior for active channel (both directions)
- No side-effects on other channels

## Test plan

- [x] All 115 existing tests pass (`vitest run`)
- [x] 10 new unit tests for the new pure functions
- [ ] Manual: click triangle on inactive channel → selects channel + expands
- [ ] Manual: click triangle on expanded active channel → collapses
- [ ] Manual: click channel name while active → toggles task list
- [ ] Manual: switch channels → still auto-expands with active tasks

> Note: Screenshots of the sidebar require a live midtown daemon. The fix is in interaction logic (event handlers), verifiable via unit tests.

### DIFF
--- a/web-app/src/lib/ChannelList.svelte
+++ b/web-app/src/lib/ChannelList.svelte
@@ -5,7 +5,6 @@
   import {
     getChannelTaskCount,
     getChannelCiStatus,
-    getChannelHasActiveTasks,
     computeExpandedAfterTriangleClick,
     computeExpandedAfterChannelNameClick,
   } from './channelUtils.js'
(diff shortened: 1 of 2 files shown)


btucker:
## Review Status

---
> [!NOTE]

PR #1563, opened 2026-02-25
fix(web): clear ghost pending messages on WS reconnect [Midtown !1828]

btucker:
## Summary

Post-merge fixes from !1826 (optimistic message send) review:

- **Ghost pending messages**: `fetchHistory`'s merge-not-replace strategy preserved `pending: true` messages in low-traffic channels when the WS echo was lost during disconnect. Added a pre-merge step to strip all pending entries from existing channels — confirmed messages come back clean via the history response; lost messages are silently discarded.

- **Opacity inconsistency**: `ThreadPanel.svelte` wrapped `MessageRow` in an extra `<div class:opacity-60>` solely for opacity, while `Channel.svelte` applies the class directly on its outer container. Standardized by adding a `class` prop to `MessageRow` (forwarded to its root div), removing the extra wrapper divs from `ThreadPanel`.

## Test plan

- [x] Added regression test: `clears ghost pending messages from channels absent in the history response`
- [x] Added baseline test: `does not leave pending messages when channel is included in history response`
- [x] All 135 existing + new tests pass
- [x] Manual: send message, disconnect, reconnect — pending ghost should not appear

### DIFF
--- a/web-app/src/lib/MessageRow.svelte
+++ b/web-app/src/lib/MessageRow.svelte
@@ -16,6 +16,7 @@
     senderClass = '',
     currentTask = undefined,
     channelName = undefined,
+    class: extraClass = '',
     children = undefined,
   } = $props()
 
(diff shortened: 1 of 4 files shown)


btucker:
### Code Review by midtown

Reviewed the diff — clean fix for the ghost pending message bug. The pre-merge strip in `fetchHistory` is the right approach, and the test coverage is solid (both absent-channel and present-channel scenarios). The ThreadPanel opacity cleanup is a nice bonus.

LGTM — enabling auto-merge.

PR #1574, opened 2026-02-26
feat(web): thread panel — show parent message as first item in scroll stream [Midtown !1840]

btucker:
## Summary
- Renders `parentMessage` as a full `<MessageRow>` at the top of the scrollable stream (avatar, sender, timestamp, markdown content)
- Adds a muted horizontal rule separator showing reply count: "1 reply", "3 replies", or "no replies yet"
- Removes the old centered "No replies yet" empty-state div — the separator handles that state
- Task/PR cards remain at the very top, above the parent message

Both desktop and mobile panels receive the same changes.

## Layout (top → bottom in scroll area)
1. Task/PR metadata cards (if applicable)
2. Parent message as `<MessageRow>` (full markdown, avatar, timestamp)
3. `── N replies ──` separator
4. Replies

## Test plan
- [x] Unit tests pass (140/140)
- [ ] Open a thread — parent message appears at top with avatar + sender header
- [ ] Empty thread shows parent + "── no replies yet ──" separator
- [ ] Thread with replies shows "── 3 replies ──" etc.
- [ ] Task thread without `parentMessage` (task with no message_id) — separator still shows, no crash
- [ ] Mobile: same layout visible in slide-in pane

### DIFF
--- a/web-app/src/lib/api.js
+++ b/web-app/src/lib/api.js
@@ -859,7 +859,11 @@ export function openTaskThread(task, channelName) {
   // Always include the clicked task even if not found above
   if (!tasks.find((t) => t.id === task.id)) tasks.unshift(task)
 
-  const parentMessage = { id: parentMessageId, from: 'lead', content: task.subject }
+  // Use the real channel message if available so the MessageRow gets correct
+  // timestamp, sender, and content.  Fall back to a synthetic stub only when
+  // the message hasn't loaded yet (rare edge case).
+  const parentMessage = channelMsgs.find((m) => m.id === parentMessageId)
+    ?? { id: parentMessageId, from: 'lead', content: task.subject }
   threadData.set({ parentMessage, channelName, messages: [], tasks })
   fetchThread(channelName, parentMessageId).then((replies) => {
     threadData.update((td) => {


btucker:
### Code review

Found 3 issues:

1. Separator renders unconditionally when `parentMessage` is null. The reply-count separator (desktop L265-269, mobile L422-426) is outside the `{#if $threadData.parentMessage}` guard. When a task thread has no backing channel message, the parent `<MessageRow>` is correctly hidden but the separator still renders -- producing a floating "no replies yet" divider with nothing above it. The separator should be wrapped in the same conditional or its own `{#if}` guard.

   Test suggestion: Open a task thread with no `message_id` (synthetic task-only thread) and verify the separator does not render when `parentMessage` is null.

https://github.com/btucker/midtown/blob/f05ff59d2d7ba62ee022b62a758dc3a997519069/web-app/src/lib/ThreadPanel.svelte#L264-L269

2. Parent `<MessageRow>` has no children/slot content, so action, insight, and mermaid messages render as plain text. Every other `<MessageRow>` in the thread panel (e.g. L272-353 for replies) provides slot content with conditional rendering for `isAction`, `isInsight`, and `hasMermaid`. The parent message at L252-261 is self-closing with no children -- if the parent is an action (`/me ...`) or contains a mermaid diagram, it falls back to `renderContent` only, losing the special formatting.

   Test suggestion: Open a thread whose parent message is a `/me` action and verify it renders with action styling (asterisk + sender color), not as plain text.

https://github.com/btucker/midtown/blob/f05ff59d2d7ba62ee022b62a758dc3a997519069/web-app/src/lib/ThreadPanel.svelte#L250-L262

3. Missing before/after screenshots for visual UI changes (CLAUDE.md says "When a PR includes visual changes to the web UI (`web-app/` or `web/`), include before/after screenshots in the PR description.")

https://github.com/btucker/midtown/blob/f05ff59d2d7ba62ee022b62a758dc3a997519069/CLAUDE.md#L120

PR #1575, opened 2026-02-26
feat: add --task flag to midtown channel post for auto-threading [Midtown !1839]

btucker:
## Summary
- Adds `--task <id>` to `midtown channel post` that auto-resolves the thread parent ID from the task's announcement message
- Reuses the existing `task.metadata` RPC (which already returns `message_id`) — no new daemon methods needed
- Updates coworker prompt to show `--task` usage for progress posts and phase transitions
- Updates common.md Threads section to document `--task` as the task shorthand

## Implementation details

The daemon already stores `task_id → message_id` in `persistent_state.task_message_id` when a task is created (populated in `rpc_task.rs`). The new `channel_post_for_task()` client method calls `task.metadata`, extracts `message_id`, then calls `channel_post_in_thread()` with the resolved ID.

Error behavior:
- Unknown task ID → `task.metadata` RPC returns an error (bubbles to CLI)
- Task exists but no announcement message → clear error: "Task !N has no announcement message (may have been created before threading was added)"

`--task` and `--thread` are declared mutually exclusive via clap `conflicts_with`.

## Test plan
- [x] `cargo test` — all 2009+ tests pass, 0 failures
- [x] `cargo fmt --all -- --check` — clean

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,8 @@ The Project Lead is just a Claude Code session, but it's been booted with a [spe
 | `midtown status` | Show system status |
 | `midtown view` | Launch chat UI (use `--attach` to open Project Lead in a split) |
 | `midtown channel post <msg>` | Post to the team channel |
+| `midtown channel post <msg> --task <id>` | Post in the thread for a task (auto-resolves thread ID) |
+| `midtown channel post <msg> --thread <msg-id>` | Post in a specific thread |
 | `midtown channel read` | Read recent messages |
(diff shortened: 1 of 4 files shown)


btucker:
Addressed Codex P2 feedback: `channel_post_for_task` now resolves the channel from `task.metadata` as a fallback when neither `--channel` nor `MIDTOWN_CHANNEL` is set. Priority: explicit `--channel` > `MIDTOWN_CHANNEL` env > task's channel from metadata.

PR #1577, opened 2026-02-26
feat(tui): thread panel shows parent message via render_message with reply-count separator [Midtown !1841]

btucker:
## Summary
- Replaces the custom parent-message renderer in the thread panel with `render_message` / `render_message_with_mermaid`, so the parent message gets the same timestamp gutter, sender header, and code/mermaid highlighting as regular replies
- Separator between parent message and replies now shows reply count: `─── 2 replies ───` or `─── no replies yet ───`
- Passes parent sender as `prev=` to the first reply, enabling correct same-sender grouping across the parent/replies boundary
- Removes `render_header_content_segments` (dead code after the switch) and its 3 associated tests; adds 3 new tests for the reply-count separator and parent sender header

## Test plan
- [x] All existing thread panel tests pass
- [x] `test_thread_separator_shows_reply_count` — separator shows `2 replies` with 2 replies
- [x] `test_thread_separator_no_replies` — separator shows `no replies yet` with 0 replies
- [x] `test_thread_parent_message_rendered_with_sender_header` — parent sender name appears in output


btucker:
## Review Status

---
> [!NOTE]

btucker:
Thanks for the review\! The table-in-mixed-content observation is valid as a known limitation of `render_message_with_mermaid`. However, the same limitation already applied to thread replies before this PR — so the parent message is now consistent with replies rather than being a special case. The previous `render_header_content_segments` did handle this better, but that divergence was part of the old "separate header box" design being replaced. If table support in mermaid-mixed messages becomes a problem, it's worth a targeted fix in `render_message_with_mermaid` itself (affecting both parent and replies uniformly).

PR #1580, opened 2026-02-26
fix(tui): fix double blank line and missing @mention highlights in thread panel [Midtown !1845] [Midtown !1846]

btucker:
## Summary

Follow-up fixes from PR #1577 review — two bugs in the TUI thread panel.

**Bug 1: Double blank line before first reply (task !1845)**

The separator between parent message and replies ends with a blank line. PR #1577 passed `parent_sender` as `prev` to the first reply so same-sender grouping worked across the boundary — but `push_sender_header`'s blank-line logic fires when two non-system senders differ, producing a second blank line.

Fix: pass `None` as `prev` for the first reply. The separator already provides visual separation; the sender grouping blank-line logic is only meaningful within a contiguous message stream, not across a hard visual boundary.

**Bug 2: @mention highlights silently dropped in mermaid/code-block path (task !1846)**

`render_message` calls `apply_mention_highlights` on its rendered content lines. `render_message_with_mermaid` built lines segment-by-segment and never called it — so `@mention` spans in messages with code blocks never received the bold modifier.

Fix: collect text-segment lines locally in `render_message_with_mermaid`, run `apply_mention_highlights` on them, then extend the output.

## Test plan

- [x] `test_first_reply_has_single_blank_line_after_separator` — confirms exactly 1 blank line between separator and first reply sender header (alice parent, bob reply)
- [x] `test_render_message_with_mermaid_applies_mention_highlights` — confirms @alice span has BOLD modifier when message contains code block + @mention
- [x] All 678 existing tests pass

### DIFF
--- a/src/bin/midtown/cli/chat/ui/messages_mermaid.rs
+++ b/src/bin/midtown/cli/chat/ui/messages_mermaid.rs
@@ -93,15 +93,17 @@ pub fn render_message_with_mermaid(
             ContentSegment::Insight(text) => {
                 let insight_style = Style::default().fg(Color::Magenta);
                 let content_lines = wrap_content(text, content_width);
+                let mut segment_lines: Vec<Line<'static>> = Vec::new();
                 for content in &content_lines {
                     let parsed = minimad_ratatui::inline(content, insight_style);
                     if is_first_content_line {
-                        lines.push(build_first_content_line(msg, &ctx, parsed));
+                        segment_lines.push(build_first_content_line(msg, &ctx, parsed));
                         is_first_content_line = false;
                     } else {
-                        lines.push(build_continuation_line(&ctx, parsed));
+                        segment_lines.push(build_continuation_line(&ctx, parsed));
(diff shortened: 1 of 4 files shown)


btucker:
Addressed both review issues:

**Issue 1 — Insight arm missing `apply_mention_highlights`**: Applied the same fix as the Text arm — collect into `segment_lines`, then `lines.extend(apply_mention_highlights(segment_lines))`. Added `test_render_message_with_mermaid_applies_mention_highlights_in_insight` to cover it.

**Issue 2 — Same-sender suppression broken**: Restored `parent_sender` and changed the fix strategy: instead of passing `None` as `prev` (which disabled suppression), I removed the separator's trailing blank line. Now `push_sender_header` provides exactly one blank line for different-sender transitions, and when the sender is suppressed no blank is added — either way, exactly the right spacing. Added `test_first_reply_same_sender_as_parent_suppresses_header` to verify suppression.

PR #1581, opened 2026-02-26
fix(web): delay tool call display clear for 4s after message [Midtown !1849]

btucker:
## Summary

- Tool call items in the activity strip disappeared instantly when a channel lead posted a response, making it impossible to read tool call summaries if the agent responded quickly
- Added a 4-second persistence delay: `agentToolItems` entries are cleared 4s after a sender's channel message arrives, not immediately
- If new `universal_items` events arrive within the delay window (agent is still working), the pending clear is cancelled — keeping the strip live

## Root Cause

In `api.js handleUpdate('channel_message')`, tool items were deleted immediately for any non-lead sender:

```js
delete updated[msg.from.toLowerCase()]
```

Channel leads (e.g., `'web'`) post with their channel name, which coincidentally matches the `agentToolItems` channel key — so their activity strip was cleared the moment their first message arrived.

## Fix

Replaced the immediate delete with a `setTimeout`-based delayed clear (4 seconds). A `Map` tracks pending timeouts per channel key, allowing the `universal_items` handler to cancel any pending clear if the agent continues working.

## Test plan

- [x] New tests cover: (1) items not cleared immediately, (2) items cleared after delay, (3) clear cancelled by new tool activity, (4) unrelated channels unaffected
- [x] All 144 existing + new tests pass

### DIFF
--- a/web-app/src/lib/api.js
+++ b/web-app/src/lib/api.js
@@ -531,22 +531,21 @@ export function handleUpdate(update) {
         // activity strip resets. Only applies to top-level messages; a thread
         // reply mid-task should not clear the coworker's tool activity.
         //
-        // agentToolItems is keyed by channel name. For a channel lead named 'web'
-        // posting to channel 'web', msg.from.toLowerCase() === 'web' === channel key.
+        // Key by channelName (not msg.from): agentToolItems is channel-keyed, so
+        // the correct key to delete is the channel the message was posted to.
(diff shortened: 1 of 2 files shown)


btucker:
Creating follow-up PR to address these review comments — all 3 issues are valid and will be fixed:

1. Add `agentClearTimeouts` cleanup to `switchProject()`
2. Fix the guard condition to use `channelName` semantics instead of `msg.from`
3. Add channel isolation test

PR #1582, opened 2026-02-26
feat(daemon): auto-fork topic channel sessions on new user messages [Midtown !1847]

btucker:
## Summary

- **Daemon auto-fork**: When a user posts a new top-level message to a topic channel, `handle_channel_post` now automatically creates a fork session (bound to the message ID) before nudging the channel lead. The nudge goes directly to the fork, so the fork session exists and handles the conversation from the start.

- **Shared fork helper**: Extracted `create_fork_session` from `handle_session_fork` as a `pub(super)` helper in `rpc_session.rs`. Adds a `channel_hint` parameter so the auto-fork path can provide the known channel name even when session records are stale.

- **Graceful fallback**: If no channel lead session is registered (first message to a new channel), auto-fork is skipped and `NudgeChannelLead` spawns one as before. If spawn fails (e.g., during tests), the "pending" sentinel is cleaned up and the call succeeds.

- **Channel lead instructions updated**: `agents/channel-lead.md` now explains that the daemon auto-forks — leads receive the nudge already in the fork session and just need to respond directly. Manual `session fork` is still documented as a fallback.

## Test plan

- [x] `create_fork_session` returns `Ok((sid, true))` when fork already exists — no spawn triggered
- [x] `create_fork_session` returns `Err` when slot is "pending" (concurrent fork guard)
- [x] `create_fork_session` cleans up "pending" sentinel on spawn failure
- [x] `handle_session_fork` preserves `already_exists: true` RPC response format
- [x] `handle_channel_post` auto-fork: sentinel cleaned up when spawn fails
- [x] `handle_channel_post` auto-fork: skipped gracefully when no channel lead exists
- [x] Thread replies still route to existing fork sessions (existing behavior preserved)
- [x] All 2031+ unit tests pass, clippy clean

### DIFF
--- a/agents/channel-lead.md
+++ b/agents/channel-lead.md
@@ -72,7 +72,7 @@ The daemon now **automatically forks** your session when a new top-level user me
    midtown session fork <message-id>
    ```
 
-   `session fork` is idempotent — calling it when a fork already exists just returns the existing session ID.
+   `session fork` is idempotent — calling it when a fork already exists returns `{already_exists: true, session_id: ...}`. During the daemon's auto-fork spawn window (~30s), it may return `{pending: true}` instead — this means the daemon is already creating the fork. Retry once after a brief wait.
 
 **After forking (or in an auto-forked session):**
 - All your text output is automatically posted to the thread — no `--thread` flag needed.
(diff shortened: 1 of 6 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1586, opened 2026-02-26
feat(start): auto-build web app on midtown start, remove dist from git [Midtown !1851]

btucker:
## Summary

- `midtown start` now automatically runs `npm run build` in `web-app/` when dist is stale or missing, preventing the class of bug where a source update isn't reflected in the running web UI
- Removes `web-app/dist/` from git tracking and adds it to `.gitignore` — built artifacts no longer belong in source control
- Adds Node.js as a prerequisite in README since `npm` is now required at startup

## How it works

`build_web_app_if_needed()` runs during `handle_start()` (after plugin checks, before daemon starts):

1. If `web-app/package.json` doesn't exist (production install without source), skip silently
2. Check staleness: compare `dist/index.html` mtime against source files in `src/`, `public/`, and key config files
3. If stale or dist is missing, run `npm run build` with the progress bar showing "building web app"
4. Warnings are logged on failure but startup never aborts

The staleness check uses `is_dist_fresh()` + `dir_has_newer_file()` for a recursive mtime walk.

## Test plan

- [x] Build compiles cleanly

### DIFF
--- a/src/bin/midtown/cli/daemon.rs
+++ b/src/bin/midtown/cli/daemon.rs
@@ -668,7 +668,12 @@ fn is_dist_fresh(web_app_dir: &Path, dist_index: &Path) -> bool {
         }
     }
 
-    for file_name in &["package.json", "vite.config.js", "svelte.config.js"] {
+    for file_name in &[
+        "index.html",
+        "package.json",
+        "vite.config.js",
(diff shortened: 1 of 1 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1591, opened 2026-02-27
fix: route all agent messages to threads [Midtown !1855]

btucker:
## Summary
- **PostToChannel effect** now resolves `fork_bound_threads` in the executor to auto-apply thread context for forked lead sessions — mirroring the existing RPC path in `rpc_channel.rs`
- **coworker.md** "Replying to Messages" section updated to always use `--task` when replying about task-related work
- All **task-related nudge prompts** (`agents.rs`, `wake_reason.rs`) now include a `--task` reply instruction via shared `task_footer()` helper
- **rpc_session_tests** assertions documented: property-based checks are correct here because `fork_channel_lead_model` reads global config; exact defaults are tested in `helpers_tests.rs`

## Details

Three issues were causing agent messages to land in the main channel instead of threading under their task:

1. The auto-post pipeline (`stream.rs` → `effects.rs`) posted forked lead output to the correct *channel* but lost the *thread*. The fix adds `fork_bound_threads` lookup in the `PostToChannel` effect executor, exactly mirroring how `rpc_channel.rs` already handles this for explicit `midtown channel post` calls.

2. The coworker prompt's "Replying to Messages" section showed every reply as a top-level post with just an @mention. Now all task-related reply examples include `--task <id>`.

3. None of the 20+ nudge paths told recipients to use `--task` when replying. A new `task_footer()` helper centralizes the reply instruction and is used by all task-related prompts in both `agents.rs` (fresh spawns) and `wake_reason.rs` (running session nudges).

### Test assertion clarification

amsterdam's review of PR #1590 flagged `test_fork_channel_lead_model_*` assertions as "weakened." Investigation shows the property-based assertions are correct: `fork_channel_lead_model` goes through the full config stack including `GlobalConfig::load()`, so exact values depend on the machine's `~/.midtown/config.toml` (e.g., `default_model = "large"` → `"opus"` for Claude). Exact assertions for the pure `default_model_for_provider_role` function already exist in `helpers_tests.rs`. Added clear documentation comments explaining this.

## Test plan
- [x] Added `test_task_footer_contains_view_and_reply_instructions` — verifies the shared footer
- [x] Added `test_all_task_prompts_include_reply_instruction` — verifies all 4 prompt types include `--task`
- [x] Added `task_nudges_include_reply_instruction` — verifies all 4 wake reason types include `--task`
- [x] `test_fork_channel_lead_model_*` — property-based assertions with documentation
- [x] All tests pass, clippy clean, fmt clean

### DIFF
--- a/src/daemon/effects.rs
+++ b/src/daemon/effects.rs
@@ -86,10 +86,16 @@ pub enum Effect {
     },
     /// Post a message to the IRC-style channel (and broadcast to WebSocket clients).
     ///
-    /// Channel routing follows a 3-step resolution:
+    /// The executor resolves channel and thread in two phases:
+    ///
+    /// **Channel routing** (3-step):
     /// 1. If `channel` is explicitly provided, use that
     /// 2. Otherwise, extract task ID from message content (e.g., "!42") and route to that task's channel
     /// 3. Fall back to the default project channel if no task ID is found
+    ///
+    /// **Thread resolution**: If the sender has an entry in `fork_bound_threads`,
+    /// the message is posted as a thread reply under the bound thread parent.
+    /// This mirrors the RPC path in `rpc_channel.rs`.
     PostToChannel {
         sender: String,
         message: String,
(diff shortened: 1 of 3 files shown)


btucker:
## Review Status

---
> [!NOTE]

PR #1593, opened 2026-02-28
fix: preserve channel history across jsonl rotation [Midtown !1857]

btucker:
## Summary
- `read_all()` and `read_last_n_messages()` now read from **all** `.jsonl` files in the history directory (date-named archives + `current.jsonl`), not just `current.jsonl`
- Fixes the web UI losing all channel history when daily rotation moves messages to `YYYY-MM-DD.jsonl` archives
- Added `list_all_history_files()` helper that lists archive files in chronological order with `current.jsonl` last
- `read_last_n_messages()` reads files in reverse chronological order and stops early once enough messages are collected (performance optimization)
- Fixed `history_dir()` to derive path from `channel_file` parent so it works for both active and archived channels

## Test plan
- [x] Added `test_read_all_includes_rotated_archives` — verifies read_all returns messages from archive files
- [x] Added `test_read_all_multiple_archive_files_chronological_order` — verifies correct ordering across multiple archive files
- [x] Added `test_read_last_n_includes_archives_when_current_insufficient` — verifies read_last_n spans archive files when current has fewer than N messages
- [x] Added `test_read_last_n_returns_zero_start_when_all_history_fits` — verifies start_position=0 when all history fits in N
- [x] Added `test_read_all_after_rotation_preserves_history` — end-to-end test: write messages, rotate, verify read_all still returns all
- [x] Added `test_list_all_history_files_ignores_rotating_temp` — verifies .rotating temp files are excluded
- [x] All 278 channel tests pass
- [x] Full test suite passes (2053 lib + 688 integration + doctests)
- [x] cargo clippy clean

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -904,15 +904,12 @@ impl Channel {
         Ok(())
     }
 
-    /// Read all messages from the channel
+    /// Read all messages from the channel, including rotated archives.
     ///
-    /// Messages are sorted by timestamp to ensure chronological order,
-    /// regardless of the order they were written to the file.
-    ///
-    /// Uses bounded retries to acquire a shared lock when there's lock contention.
-    /// Retries up to 10 times with 50ms delays (500ms total) to handle transient
-    /// lock contention from writers. Returns an error if the lock can't be acquired
-    /// after 500ms.
+    /// Reads every `.jsonl` file in the history directory (date-named archives
+    /// in ascending order, then `current.jsonl`). Messages are sorted by
+    /// timestamp to ensure chronological order. Each file is locked
+    /// individually via `read_messages_from_file`.
     pub fn read_all(&self) -> Result<Vec<Message>> {
         let history_files = self.list_all_history_files();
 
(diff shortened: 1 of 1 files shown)


btucker:
## P1 Fix: Pagination sentinel breaks callers

**Root cause:** `read_last_n_messages` returned `0`/`1` as sentinels for `start_position`, but `read_messages_before_position` still interpreted `position` as a byte offset in `current.jsonl`. When `position=1`, it would seek to byte 1 and produce garbage results.

**Fix:** Changed `start_position` from byte offsets to **message counts from the tail**:
- `read_last_n_messages(20)` with 100 total → returns 20 messages, `start_position = 20`
- `read_messages_before_position(20, 10)` → reads all history, returns messages at positions `[total-30 .. total-20]`, `new_position = 30`
- When `start_position = 0`, all history is loaded (same as before)

Also rewrote `read_messages_before_position` to be archive-aware — it reads from all history files instead of only `current.jsonl`.

**P2 fix (bonus):** The early-exit loop in `read_last_n_messages` checked `had_enough_before` (before extending the collection), which caused it to read one extra archive file. Now breaks immediately when enough messages are collected.

**Tests:** Added `test_pagination_across_archives` covering multi-page navigation (4 pages, 10 messages across archive + current). All 2054 tests pass. Clippy clean.

PR #1598, opened 2026-02-28
fix: include parent message in thread history API response [Midtown !1863]

btucker:
## Summary
- **Backend**: Added `m.id == parent_id` to the thread history filter in `api_channel_history`, so the parent message is included as the first element in the response — making it self-contained
- **Frontend**: Both `openThread()` and `openTaskThread()` now extract the parent from the API response and upgrade `threadData.parentMessage`, replacing any synthetic stub with real data (correct sender, timestamp, content)
- **Test**: Added unit test `test_thread_history_filter_includes_parent_message` and updated E2E test assertions

## Root cause
The `/channels/history?thread_parent_id=X` endpoint filtered to only return replies (where `thread_parent_id == X`), excluding the parent message itself. The frontend had to look up the parent in the local message store, which failed when history wasn't loaded or had rotated — causing threads to show a synthetic stub or appear empty.

## Test plan
- [x] Unit test verifies thread filter includes parent + replies
- [x] Updated E2E test expects parent as first message in thread response
- [x] All 2056 tests pass, clippy clean, fmt clean

### DIFF
--- a/src/web.rs
+++ b/src/web.rs
@@ -518,13 +518,20 @@ async fn api_channels_create(
     ))
 }
 
+/// Returns true if a message belongs to the given thread: either it is the
+/// parent message itself or a reply to it. Used by `api_channel_history` and
+/// tested directly in unit tests.
+pub(crate) fn is_in_thread(msg: &crate::message::Message, parent_id: &str) -> bool {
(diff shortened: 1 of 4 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1604, opened 2026-02-28
fix: limit channel history API response to prevent iOS Safari crash

btucker:
## Summary
- The `/api/channels/history` endpoint was returning ALL messages from ALL archived history files (105K messages / 27MB for the midtown channel), causing iOS Safari to crash with "A problem repeatedly occurred"
- Switched non-thread queries from `read_all_async()` to `read_last_n_messages_async(limit)` with a default limit of 500 messages
- Added optional `?limit=N` query parameter for clients to request a specific number of messages
- Thread queries (with `thread_parent_id`) still use `read_all_async()` since thread replies are small and may span archives

## Test plan
- [x] `cargo build` compiles successfully
- [x] `cargo test` — all 2085+ tests pass, 0 failures
- [x] `cargo clippy` and `cargo fmt` pass (pre-commit hooks)
- [ ] `curl http://localhost:47024/api/channels/history | python3 -c "import sys,json; data=json.load(sys.stdin); print(len(data))"` — should return ~500 (not 105K)
- [ ] `curl "http://localhost:47024/api/channels/history?limit=10"` — should return 10 messages
- [ ] Load web app on iOS Safari — should show recent messages without crashing

### DIFF
--- a/src/web.rs
+++ b/src/web.rs
@@ -533,6 +533,9 @@ struct ChannelHistoryQuery {
     /// Optional thread parent ID to filter by. If provided, return the parent
     /// message itself plus all replies to it.
     thread_parent_id: Option<String>,
+    /// Maximum number of messages to return (default: 500). Only applies to
+    /// non-thread queries; thread queries always return all replies.
+    limit: Option<usize>,
 }
 
 /// Get channel message history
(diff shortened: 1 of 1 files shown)


btucker:
### Code review

PR #1605, opened 2026-02-28
fix: flush mobile keyboard composition buffer on message submit [Midtown !1870]

btucker:
## Summary
- On iOS Safari and mobile Chrome, the virtual keyboard maintains a composition buffer (for autocorrect/predictive text) that can overwrite the textarea value after Svelte's reactive binding clears it
- Added `blur()` before `focus()` after clearing the textarea value in both Channel.svelte and ThreadPanel.svelte's `handleSubmit` functions to force the browser to finalize/discard pending composition state
- Applied the fix to all three submit code paths: Channel file upload, Channel plain message, and ThreadPanel reply

## Test plan
- [x] Web app builds successfully (`vite build`)
- [ ] On iOS Safari (PWA mode), type a message and tap Send — input should clear completely
- [ ] On mobile Chrome, type a message and tap Send — input should clear completely
- [ ] Thread panel reply also clears correctly on mobile
- [ ] Desktop behavior unchanged (Enter key submit still works)

### DIFF
--- a/web-app/e2e/input-behavior.spec.js
+++ b/web-app/e2e/input-behavior.spec.js
@@ -52,6 +52,51 @@ test.describe('Message input behaviors', () => {
     await expect(input).toHaveValue('')
   })
 
+  test('submit does not programmatically blur the textarea', async ({ page }) => {
+    const input = page.getByTestId('channel-input')
+    await input.fill('Test message')
+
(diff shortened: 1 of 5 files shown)


btucker:
## Review Status

---
> [!NOTE]

PR #1613, opened 2026-02-28
fix: Complete no-PR tasks directly instead of nudging for PR [Midtown !1879]

btucker:
## Summary

- When a coworker reported `midtown state completed` on a task without an open PR, the daemon nudged "please open a PR" but left the task `in_progress`. This caused `dispatch_via_sessions` to respawn the coworker repeatedly (riverside hit 7+ sessions on task !1878 for v0.6.0 release)
- Now the daemon completes the task on disk when the coworker reports completed, even without a PR. This handles legitimate no-PR tasks (release management, ops, investigations)
- Tasks with open PRs still defer to the merge auto-completion path (existing behavior preserved)

## Test plan

- [x] `test_completed_without_pr_marks_task_done` — verifies task is marked completed on disk when reported completed without PR
- [x] `test_completed_with_open_pr_defers_to_merge_path` — verifies task with open PR stays `in_progress` (deferred to merge path)
- [x] `cargo clippy` and `cargo fmt` pass

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -388,6 +388,14 @@ When a coworker calls `midtown pr merge --pr <N>`, the daemon runs three gates b
 
 When a coworker is called in, midtown creates a detached git worktree at the current HEAD. The coworker creates a feature branch and works independently. When the coworker shuts down, worktrees with no commits and no uncommitted changes are automatically cleaned up along with their branches. Worktrees with work in progress are preserved.
 
+## Task Completion
+
+Tasks complete through two paths depending on whether they produce a PR:
+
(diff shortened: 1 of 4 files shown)


btucker:
Addressed all 4 issues in 56be795:

**1. `task_has_open_pr` now checks task.pr on disk** — Added a second source check that reads the task file from disk using the new `read_task_for_repo()` function. This survives daemon restarts where `pr_author_sessions` is empty. Added test `test_completed_with_pr_on_disk_defers_to_merge_path` that sets `task.pr` on disk without `pr_author_sessions` and verifies the task stays `in_progress`.

**2. Doc comment updated** — Rewrote the `task_has_open_pr` doc comment to describe both completion strategies: tasks WITH open PRs defer to merge path, tasks WITHOUT are completed directly.

**3. Test isolation** — Acknowledged as pre-existing pattern. Created follow-up task via `midtown task request` for routing task paths through temp dirs.

**4. Architecture docs** — Added "Task Completion" section to `docs/architecture.md` documenting both completion paths (PR tasks vs no-PR tasks).

PR #1616, opened 2026-02-28
fix: Check disk task.pr field in task_has_open_pr [Midtown !1882]

btucker:
## Summary
- Fix `task_has_open_pr` to verify the on-disk `task.pr` field via GitHub API (`gh pr view`) before trusting it
- Add `is_pr_open()` helper that checks if a PR is actually in OPEN state, returning false for closed/merged/unreachable PRs
- Conservative fallback: API failures treat PR as not-open so tasks complete directly rather than getting stuck

## Context
After a daemon restart, `pr_author_sessions` (in-memory HashMap) is empty. The initial fix (!1879) added a disk fallback using `task.pr`, but the field is never cleared when a PR is closed without merge. This means a closed/superseded PR would incorrectly block task completion. The fix verifies the PR state via `gh pr view` before trusting the disk field.

## Test plan
- [x] `test_completed_with_unverifiable_disk_pr_completes_directly` — disk PR with unverifiable state completes directly
- [x] `test_completed_with_open_pr_defers_to_merge_path` — in-memory path still defers correctly
- [x] `test_is_pr_open_returns_false_for_nonexistent_pr` — API failure returns false
- [x] `test_is_pr_open_returns_false_with_no_repo_path` — no repo path returns false
- [x] Full test suite: 2106 passed, 0 failed

### DIFF
--- a/src/daemon/rpc_coworker.rs
+++ b/src/daemon/rpc_coworker.rs
@@ -691,7 +691,7 @@ pub(super) async fn handle_coworker_questions(id: RequestId, state: &DaemonState
 /// Returns true if the task has an associated open PR, checking two sources:
 ///
 /// 1. **`pr_author_sessions` (in-memory persistent state)**: Presence implies
-///    the PR is still open — closed PRs are cleaned up by `cleanup_closed_pr_state`.
+///    the PR is still open — closed PRs are cleaned up by `cleanup_closed_prs`.
 ///    This mapping is established when a coworker opens a PR and the daemon
 ///    extracts the task ID from the PR title's `[Midtown !XXX]` marker.
 ///


btucker:
## Review Status

---
> [\!NOTE]

btucker:
Both issues addressed:

1. **Stale `task.pr` false positives** — `task_has_open_pr` now verifies the PR is actually open via `gh pr view` before trusting the disk field. If the API reports the PR is closed/merged/unreachable, the disk field is ignored and the task can complete directly. Added `test_completed_with_unverifiable_disk_pr_completes_directly` to cover this case. (See `is_pr_open` helper at rpc_coworker.rs:754)

2. **Doc comment typo** — Fixed `cleanup_closed_pr_state` → `cleanup_closed_prs` in the doc comment.

Rebased onto latest main to pick up the install script commit.

PR #1617, opened 2026-03-01
feat: Add curl | bash install script and document in README [Midtown !1883]

btucker:
## Summary
- Add `install.sh` to the repo root — a POSIX-compatible shell script that detects OS/arch, downloads the latest release binary from GitHub, and installs it to `~/.cargo/bin` (configurable via `MIDTOWN_INSTALL_DIR`)
- Update README.md to show `curl | sh` as the primary (quickest) install method, keeping `cargo install` and source build as alternatives

## Test plan
- [x] Tested locally on macOS arm64 — script downloads v0.6.1 binary and installs correctly
- [x] Verified installed binary runs (`midtown --help`)
- [x] Script handles missing PATH gracefully with a helpful message

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ The quickest way to install:
 curl -fsSL https://raw.githubusercontent.com/btucker/midtown/main/install.sh | sh
 ```
 
-This detects your OS and architecture, downloads the latest release binary, and installs it to `~/.cargo/bin`. Set `MIDTOWN_INSTALL_DIR` to change the install location.
+This detects your OS and architecture, downloads the latest release binary, and installs it to `~/.cargo/bin`. Set `MIDTOWN_INSTALL_DIR` to change the install location. If `~/.cargo/bin` is not already in your PATH, the script will print instructions to add it.
 
 Or install from [crates.io](https://crates.io/crates/midtown):
 
--- a/install.sh
+++ b/install.sh
@@ -53,19 +53,23 @@ echo "Latest version: ${VERSION}"
 
 # ── Download and install ─────────────────────────────────────────────────────
 
-ASSET="midtown-${OS}-${ARCH}-${VERSION}.tar.gz"
+# Normalize version: strip leading "v" to get the bare version number.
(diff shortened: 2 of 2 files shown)


btucker:
### Code review

Found 2 issues:

1. **README omits PATH setup note for non-Rust users.** The README says the script "installs it to `~/.cargo/bin`" but does not mention that users without an existing Rust toolchain may need to add that directory to their PATH. The script itself correctly detects and warns about this at runtime (lines 76-83), but users reading only the README would not know to expect a PATH configuration step. CLAUDE.md says "If your PR changes user-facing behavior, the README should reflect it." Consider adding a brief note like: "If `~/.cargo/bin` is not already in your PATH, the script will print instructions to add it."

   https://github.com/btucker/midtown/blob/3ebce7c23209ab0edb978db1f32644f0e2ff9f3e/README.md#L30-L32

2. **Implicit coupling between install script and publish workflow on version tag format.** The asset filename is constructed as `midtown-${OS}-${ARCH}-${VERSION}.tar.gz` where `$VERSION` is the raw `tag_name` from the GitHub API (e.g., `v0.6.1`). The publish workflow (publish.yml) independently strips the `v` prefix then re-adds it as a literal `-v` separator. Both paths currently produce the same filename, but through different logic -- the install script embeds the full tag, while the workflow concatenates `-v` + bare version. A change to either side (e.g., dropping the `v` prefix from tags, or changing the asset naming template) would silently break downloads. Consider adding a comment documenting this coupling, or normalizing the version format explicitly (e.g., `VERSION="${VERSION#v}"` then `ASSET="midtown-${OS}-${ARCH}-v${VERSION}.tar.gz"`).

   Test suggestion: Add a test that verifies `install.sh` constructs the same asset filename as the publish workflow for a given tag (e.g., parse both naming patterns and assert equality for `v0.6.1`, `v1.0.0`).

   https://github.com/btucker/midtown/blob/3ebce7c23209ab0edb978db1f32644f0e2ff9f3e/install.sh#L55-L58

PR #1621, opened 2026-03-01
feat: Build web-app as part of releases [Midtown !1888]

btucker:
## Summary
- Add a `web-app` build job to the publish workflow that builds the Svelte PWA once using Node 22 and shares it as an artifact
- Modify the `binaries` job to depend on `web-app` and bundle `web-app/dist/` alongside the `midtown` binary in each platform tarball
- Add a `web-builder` stage to the Dockerfile that builds the web-app and copies `dist/` next to the binary in the runtime image
- Update `install.sh` to extract and install the bundled `web-app/` directory alongside the binary

This ensures `resolve_web_dir()` finds the web UI at `exe_dir/web-app/dist` for all distribution methods: release tarballs, Docker images, and the curl installer.

## Test plan
- [x] Workflow YAML validates (proper job dependencies, artifact passing)
- [x] Binary tarballs will contain `midtown` + `web-app/dist/` structure
- [x] Dockerfile multi-stage build adds web-builder stage and copies dist to `/usr/local/bin/web-app/dist`
- [x] install.sh handles presence of `web-app/` directory in tarball
- [ ] CI runs the publish workflow on next release

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -233,6 +233,7 @@ jobs:
 
       - name: Package binary with web-app
         run: |
+          mkdir -p staging
           cp target/${{ matrix.target }}/release/midtown staging/
           cd staging
           tar czf ../${{ matrix.asset }}-v${{ steps.version.outputs.version }}.tar.gz midtown web-app/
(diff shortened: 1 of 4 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1627, opened 2026-03-01
feat: Add midtown update command (auto-update from GitHub releases) [Midtown !1893]

btucker:
## Summary
- Adds `midtown update` command that downloads and installs the latest release from GitHub releases
- Adds `midtown update --check` to check for updates without installing
- Adds background version check on `midtown start` that prints a notice when a newer version is available (e.g., `midtown v0.7.0 is available, run midtown update to upgrade`)
- Stores last-check timestamp in `~/.midtown/update-last-check` to rate limit checks (max once per hour)

### Implementation details
- Uses the GitHub redirect trick from `install.sh` to detect the latest version without API rate limits
- Platform detection uses `cfg!()` compile-time macros for OS (darwin/linux) and arch (amd64/arm64)
- Binary replacement uses atomic swap with rollback: rename current → .old, copy new → current, delete .old
- Web-app directory uses the same atomic swap pattern as `install.sh`
- HTTP client uses `reqwest::blocking` (already a dependency)

## Test plan
- [x] Unit tests for version comparison (`is_newer`) covering equal, older, newer, major/minor bumps
- [x] Unit test for platform detection
- [x] Unit test for last-check file path
- [x] All 695 binary tests pass

### DIFF
--- a/src/bin/midtown/cli/update.rs
+++ b/src/bin/midtown/cli/update.rs
@@ -87,13 +87,21 @@ pub fn check_for_update_notice() -> Option<String> {
         return None;
     }
 
-    // Run the check in a background thread with a short timeout
-    let handle = std::thread::spawn(fetch_latest_version);
+    // Run the check in a background thread with a short timeout.
+    // Use a channel with recv_timeout to enforce a 3-second wall-clock deadline,
+    // rather than join() which would block for the full HTTP timeout (10s).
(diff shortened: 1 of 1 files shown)


btucker:
## Review Status

🔍 Review in progress by riverside...

---
> [\!NOTE]

PR #1625, opened 2026-03-01
test: add E2E test for curl|bash install process [Midtown !1892]

btucker:
## Summary
- Adds `tests/install_e2e.rs` with two tests that verify the `install.sh` script works end-to-end
- Uses a mock HTTP server (raw `TcpListener`) to simulate GitHub's release redirect and tarball download
- Tests the full `curl -fsSL .../install.sh | sh` flow including binary installation, executable permissions, web-app/ directory, and PATH warning
- Also tests the atomic web-app/ replacement path (existing web-app/ → .old → swap → cleanup)
- Adds the test suite to the containerized E2E coordination suite in `e2e-entrypoint.sh`
- Copies `install.sh` into the E2E Docker runtime stage so `include_str!` resolves during container recompilation

## Test plan
- [x] `cargo test --test install_e2e` passes locally (2 tests)
- [x] `cargo clippy --test install_e2e -- -D warnings` clean
- [x] `cargo fmt --all -- --check` clean
- [ ] Containerized E2E suite (`./scripts/e2e-container.sh coordination`) picks up the new `install_e2e` suite

🌃 Co-built with [Midtown](https://github.com/btucker/midtown)

PR #1630, opened 2026-03-01
fix: remove dead zellij_e2e reference and harden install_e2e tests [Midtown !1898]

btucker:
## Summary
- Removed dead `zellij_e2e` entry from `scripts/e2e-entrypoint.sh` — the test file was deleted previously but the entrypoint still referenced it
- Addressed post-merge review feedback on PR #1625 (install_e2e tests):
  - Added `#[ignore]` to both E2E tests so they're skipped during standard `cargo test`
  - Added `--ignored` flag to `install_e2e` invocation in `e2e-entrypoint.sh`
  - Made `read_request()` loop until `\r\n\r\n` instead of relying on a single `read()` call
  - Saved server thread `JoinHandle` and joined after each test completes
  - Fixed misleading comment about fake binary responding to `--version`

## Test plan
- [x] `cargo clippy` passes
- [x] `cargo test --test install_e2e` confirms both tests are `ignored` by default
- [x] `cargo test --test install_e2e -- --ignored` runs the tests when explicitly requested

### DIFF
--- a/scripts/e2e-entrypoint.sh
+++ b/scripts/e2e-entrypoint.sh
@@ -112,12 +112,8 @@ run_coordination_tests() {
             --skip test_mailbox_fallback \
             "${test_args[@]}"
 
-    run_bg "zellij_e2e" \
-        cargo test --release --test zellij_e2e -- --ignored --test-threads=1 \
-            "${test_args[@]}"
-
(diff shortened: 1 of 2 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1642, opened 2026-03-01
feat: Save and restore per-channel/per-thread message drafts [Midtown !1910]

btucker:
## Summary
Incremental improvements to the per-channel/per-thread draft feature (base merged in #1641):

- **Clear drafts on send** — `channelDrafts.delete()` after sending prevents stale drafts from restoring after a message is sent
- **Resize textarea on restore** — `tick().then(() => resizeTextarea())` ensures textarea height adjusts for multi-line restored drafts
- **Persist thread drafts across panel close/reopen** — Move `threadDrafts` Map to `<script module>` scope so it survives `ThreadPanel` unmount/remount cycles; save current draft in `onDestroy` before the component is torn down
- **Clear thread draft on send** — `threadDrafts.delete(parentId)` after sending a thread reply

## Test plan
- [ ] Type multi-line text in #web, switch to #daemon, switch back — textarea height should match the restored text
- [ ] Send a message in #web, switch away and back — input should be empty (not the sent message)
- [ ] Open a thread, type a reply, close the thread panel, reopen the same thread — reply text should be restored
- [ ] Send a thread reply, close/reopen thread — reply input should be empty

### DIFF
--- a/web-app/src/lib/ThreadPanel.svelte
+++ b/web-app/src/lib/ThreadPanel.svelte
@@ -269,7 +269,7 @@
     const parentId = $threadData.parentMessage?.id ?? null
     sendMessage(replyText.trim(), $threadData.channelName, parentId)
     replyText = ''
-    if (parentId) threadDrafts.delete(parentId)
+    if (currentThreadId) threadDrafts.delete(currentThreadId)
     clearMobileTextarea(textareaEl, () => { replyText = '' })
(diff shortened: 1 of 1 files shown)


btucker:
Closing — this PR is superseded by the already-merged PR #1641, which implements the same draft persistence feature with a cleaner single-Map design. All review findings (stale file entries, wrong thread draft key) are already resolved in #1641.

PR #1654, opened 2026-03-02
fix: Resolve --task thread parent from task_thread_id, not task_message_id [Midtown !1920]

btucker:
## Summary
- Fixed `--task <id>` resolution in `midtown channel post` to use `task_thread_id` (conversation root) instead of `task_message_id` (announcement reply) when threading posts
- Added `thread_id` to the `task.metadata` RPC response so the CLI can access it
- When a task is created with `--thread-id <parent>`, the announcement is a reply within that thread — using `message_id` as the thread parent would nest replies incorrectly instead of creating siblings
- Falls back to `message_id` for backward compatibility with tasks created before `task_thread_id` was introduced

### Root cause
PR #1651 correctly threaded task announcements under `--thread-id`, but the CLI's `channel_post_for_task` still resolved thread parent from `message_id` (the announcement reply) rather than `thread_id` (the parent thread). This would cause coworker `--task` posts to create nested replies-to-a-reply instead of sibling replies in the conversation thread.

### Web UI already handled this
The web UI's `openTaskThread()` already walks up `thread_parent_id` to find the root, so it was unaffected.

## Test plan
- [x] Added 3 unit tests documenting `--task` thread resolution semantics (prefer `thread_id`, equivalence without `--thread-id`, fallback to `message_id`)
- [x] All existing tests pass (`cargo test` — 49 suites)
- [x] `cargo clippy` passes with no warnings

### DIFF
--- a/src/bin/midtown/client/mod.rs
+++ b/src/bin/midtown/client/mod.rs
@@ -199,12 +199,15 @@ impl DaemonClient {
         self.channel_post_as(message, &from, channel, None)
     }
 
-    /// Post a message threaded to a task's announcement message.
+    /// Post a message threaded to a task's thread.
     ///
-    /// Resolves the task's announcement message ID via `task.metadata`, then
-    /// posts as a thread reply. Returns an error if the task is not found
-    /// (daemon returns RPC error) or if the task has no announcement message
-    /// (e.g. created before threading was added).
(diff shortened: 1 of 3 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1668, opened 2026-03-02
fix(web): Scope thread tool-clear guard to fork owner and fix thinking-state crosstalk [Midtown !1937]

btucker:
## Summary
- **Bug 1**: Thread-clear guard checked `msg.from !== 'user'` (any non-user sender), so a coworker or lead posting to a thread while a fork was active would prematurely clear its tool display. Now tracks fork ownership via a `threadOwners` Map and only the owning fork's messages trigger the delayed clear.
- **Bug 2**: ThreadPanel's thinking-state `$effect` used `threadItems || channelItems` OR to detect InProgress tool calls. Concurrent channel-lead activity falsely cleared thread thinking dots. Now uses `parentId ? threadItems : channelItems` to scope detection to the correct activity stream.

## Test plan
- [x] 3 new regression tests for Bug 1 (non-fork sender, lead sender, fork-only clear)
- [x] All 171 web-app tests pass
- [x] Rust test suite passes (no backend changes)
- [ ] Manual: open a thread with an active fork, have another sender post — tool display should persist
- [ ] Manual: send a thread reply while channel lead is active — thinking dots should not clear prematurely

### DIFF
--- a/web-app/src/lib/ThreadPanel.svelte
+++ b/web-app/src/lib/ThreadPanel.svelte
@@ -144,16 +144,18 @@
   })
 
   // Clear thinking when real InProgress tool items arrive.
-  // When parentId is set, only check thread-scoped items (fork tool calls); otherwise
-  // fall back to channel-scoped items (for task-only threads / DM channels).
-  // Using an OR of both sources caused channel-lead InProgress items to falsely clear
-  // thread thinking state — a form of crosstalk between unrelated activity streams.
+  // When a fork is handling the thread, its tool calls land in threadToolItems — use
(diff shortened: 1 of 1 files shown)


btucker:
## Review Status

---
> [\!NOTE]

btucker:
Re: missing screenshots — acknowledged. The changes here are to invisible timing logic (when the thinking dots indicator clears), not to visual layout or styling. The dots themselves look identical before and after; the fix just prevents them from clearing prematurely due to crosstalk from unrelated activity streams. A screenshot wouldn't capture the behavioral difference — it would require a screen recording with specific concurrent activity to reproduce. Noted for future PRs where the visual change is more apparent.

PR #1677, opened 2026-03-02
fix: Remove thread-specific dim styling for midtown sender [Midtown !1940]

btucker:
## Summary
- Removed `THREAD_SENDER_OVERRIDES` and `THREAD_DIM_SENDERS` constants from `ThreadPanel.svelte` that forced midtown messages to appear dimmer in threads
- Removed `senderOverrides` and `dimSenders` props from all four `MessageRow` usages (desktop parent, desktop replies, mobile parent, mobile replies)
- Thread messages now use the same color/styling as channel messages (via default `undefined` params)

## Visual change (no screenshot available — headless environment)

**Before:** midtown sender name in threads rendered as dim gray (`#585858`), message body text used `text-muted-foreground`
**After:** midtown sender name in threads uses the standard AVENUE_COLORS gold (`#E3BD3F`), message body text uses `text-foreground` — matching channel rendering

## Test plan
- [x] Web app builds successfully (`vite build`)
- [x] All CI checks green (13/13)
- [ ] Visual verification: midtown messages in threads should have the same text color as in channels

### DIFF
--- a/web-app/src/lib/ThreadPanel.svelte
+++ b/web-app/src/lib/ThreadPanel.svelte
@@ -19,11 +19,6 @@
   import DiffView from './DiffView.svelte'
   import { clearMobileTextarea } from './mobileInput.js'
 
-  const THREAD_SENDER_OVERRIDES = {
-    midtown: '#585858',
-  }
-  const THREAD_DIM_SENDERS = ['midtown']
-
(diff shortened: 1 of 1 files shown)


btucker:
👍 Addressing this now — taking before/after screenshots.

PR #1697, opened 2026-03-03
fix: auth switch --global sends RPC to all running daemons [Midtown !1958]

btucker:
## Summary
- `midtown auth switch` with global scope now broadcasts `auth.switch` to **all** running daemons, not just the current project's daemon
- Adds `enumerate_daemon_sockets()` to scan `~/.local/state/midtown/*/daemon.sock` for all running daemons
- Adds `DaemonClient::connect_to()` for connecting to a daemon at an explicit socket path
- Handles stale sockets gracefully — if all connections fail, falls back to updating config on disk

## Bug
Previously, global auth switch only reached the current project's daemon. Other running daemons (for other projects) kept sessions alive with stale auth credentials.

## Test plan
- [x] Unit tests for `enumerate_daemon_sockets` (4 tests: empty dir, finds sockets, skips dirs without sockets, skips non-dirs)
- [x] Existing auth tests pass
- [x] `cargo clippy` clean
- [x] `cargo fmt` clean

### DIFF
--- a/src/daemon/rpc_auth_tests.rs
+++ b/src/daemon/rpc_auth_tests.rs
@@ -755,3 +755,63 @@ fn test_execution_role_for_coworker() {
         crate::config::ExecutionRole::Coworker
     );
 }
+
+// ============================================================================
+// handle_auth_switch — global fan-out force flag
+// ============================================================================
+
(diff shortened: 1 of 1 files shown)


btucker:
## Review Status

---
> [!NOTE]

PR #1698, opened 2026-03-03
fix: Use project-aware auth profile resolution for fork sessions

btucker:
## Summary
- Fork sessions were using `current_profile_dir_for()` (global resolution) to resolve auth credentials, but after a per-project `auth switch`, only the project config is updated — the global marker stays the same
- Changed `create_fork_session` in `rpc_session.rs` to use `active_profile_dir_for_project_with_provider()` which checks the project config first, matching what the coworker relaunch path in `rpc_auth.rs` already does
- Added regression test verifying that project-aware and global profile resolution diverge after a per-project auth switch

Closes !1960

## Root cause
The auth system has two resolution paths:
- **Global**: `current_profile_dir_for(provider)` — reads `[providers.claude].auth_profile` from global config
- **Project-aware**: `active_profile_dir_for_project_with_provider(project, provider)` — checks project config first, falls back to global

When `midtown auth switch alice` runs (per-project), it updates the project config but not the global marker. The coworker relaunch code in `rpc_auth.rs` correctly uses the project-aware path. But `create_fork_session` was using the global path, so fork sessions launched after a per-project auth switch got the old (pre-switch) credentials → "Not logged in".

## Test plan
- [x] New test `test_fork_auth_uses_project_profile_not_global` verifies the invariant
- [x] `cargo test` — all pass
- [x] `cargo clippy` — clean

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -278,6 +278,8 @@ Channel leads can fork themselves into thread-specific sessions via the `session
 - `SessionRecord.bound_thread_id` (persisted) stores the binding on each session so restarts can rebuild the cache.
 - `name_to_session` / `session_to_name` reverse maps are backfilled in `create_fork_session` since fork sessions (--resume --fork-session) never emit `system/init` and the session ID is pre-assigned via `--session-id`.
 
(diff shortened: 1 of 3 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1699, opened 2026-03-03
fix: Add push notification bell icon to sidebar header [Midtown !1961]

btucker:
## Summary
- The push notification toggle was lost during the sidebar refactor from custom `Sidebar.svelte` to shadcn `SidebarProvider` components
- The orphaned `Sidebar.svelte` had correct Lucide `Bell`/`BellOff` icons, but `App.svelte`'s new layout never included them
- Adds the `Bell`/`BellOff` toggle button to the sidebar header, positioned between the search icon and theme toggle, matching their existing style
- Includes visual feedback: active subscriptions highlight with primary color, denied permissions show reduced opacity

## Test plan
- [x] `vite build` passes cleanly
- [x] `cargo clippy` and `cargo fmt` pass (pre-commit hooks)
- [ ] Verify bell icon appears in sidebar header next to search and theme toggle
- [ ] Click bell to subscribe/unsubscribe push notifications
- [ ] Verify bell icon shows `Bell` when subscribed, `BellOff` when not
- [ ] Verify disabled state when notifications are blocked in browser settings

### DIFF
--- a/web-app/src/App.svelte
+++ b/web-app/src/App.svelte
@@ -30,6 +30,7 @@
     pushSubscribed,
     subscribePush,
     unsubscribePush,
+    checkPushSubscription,
   } from '$lib/push.js'
 
   $effect(() => {
@@ -83,6 +84,9 @@
   })
 
   onMount(async () => {
+    // Initialize push notification state (checks browser support + existing subscription)
+    checkPushSubscription()
+
     // Always in multi-project mode — served from shared gateway on port 47022
     const projectList = await fetchProjects()
 


btucker:
### Code review

Found 3 issues:

1. **Missing before/after screenshots in PR description** (CLAUDE.md says "When a PR includes visual changes to the web UI (`web-app/` or `web/`), include before/after screenshots in the PR description.")

This PR adds a visible bell icon button to the desktop sidebar header. The test plan lists unchecked visual verification items but no screenshots are provided.

https://github.com/btucker/midtown/blob/d119f9ac4e2689d5ef27a1069efe20ca55f02d3d/CLAUDE.md#L121-L122

Test suggestion: N/A (process/docs issue)

2. **Bell icon hidden on unsupported browsers with no fallback UI** (bug due to `{#if $pushSupported}` gate)

The button is wrapped in `{#if $pushSupported}`, which starts as `false` and only becomes `true` after `checkPushSubscription()` confirms browser support. On browsers where push is not supported (iOS < 16.4, non-HTTPS, Firefox with push disabled), the bell icon is permanently hidden with no visual indication that the feature exists. The user reported "cannot see the notification icon at all" -- this gate is the most likely cause. The button should always render, using a `disabled` state with tooltip for unsupported environments instead of hiding entirely.

https://github.com/btucker/midtown/blob/9828a2048f48bdf44939a7dcd2c1024bfa2b236f/web-app/src/App.svelte#L219-L243

Test suggestion: Add a test that sets `pushSupported` to `false` and asserts the bell button still renders (disabled with tooltip) rather than being absent from the DOM.

3. **Race condition in `togglePush` -- no in-flight guard on async subscribe/unsubscribe** (bug due to missing loading state)

The `togglePush` function is async but the button is never disabled while the operation is in progress. Rapid clicks will fire concurrent `subscribePush()`/`unsubscribePush()` calls that can corrupt `pushSubscribed` state. Add a local `toggling` flag that disables the button while the async operation is pending.

https://github.com/btucker/midtown/blob/9828a2048f48bdf44939a7dcd2c1024bfa2b236f/web-app/src/App.svelte#L54-L61

Test suggestion: Add a test that clicks the bell button twice in rapid succession (before the first subscribe resolves) and asserts only one subscribe call is made.

PR #1708, opened 2026-03-03
fix: Retain channel messages when single-channel fetch returns empty [Midtown !1968]

btucker:
## Summary
- Fix intermittent bug where channel content disappears then reappears
- Guard `fetchHistory(channelName)` against replacing non-empty channel data with empty API responses
- Add 3 regression tests covering the empty-response, new-channel, and refresh scenarios

## Root Cause
`fetchHistory(channelName)` unconditionally replaced a channel's messages with the API response. When the backend transiently returned `[]` (e.g., during channel file rotation or archiving race), the channel display was wiped — only new WebSocket messages appeared until the next successful fetch.

The bulk `fetchHistory()` path already used a merge strategy that preserved existing channels, and the `catch` block retained data on network errors. The single-channel path was the only one missing this "retain last-known-good" guard.

## Fix
Skip the store update when the response is empty but the channel already has cached messages. Log a `console.warn` for observability.

## Test plan
- [x] New test: `retains existing channel messages when single-channel fetch returns empty` — confirms the guard
- [x] New test: `allows single-channel fetch to populate an empty channel` — no false positive on first load
- [x] New test: `replaces existing channel messages when single-channel fetch returns non-empty data` — normal refresh still works
- [x] All 193 existing tests pass

### DIFF
--- a/web-app/src/lib/api.js
+++ b/web-app/src/lib/api.js
@@ -235,11 +235,25 @@ export async function fetchHistory(channelName = null) {
         // The backend can transiently return [] when the channel file is briefly
         // missing (rotation, archiving race). Same "retain last-known-good"
         // pattern used in the catch block below.
-        if (channelMsgs.length === 0) {
+        //
+        // Check data.length (raw response), not channelMsgs.length: a response
+        // containing only thread replies is non-empty but annotateThreadReplyCounts
+        // filters them all out, producing channelMsgs=[]. That's real data and
+        // must not be treated as a transient empty response.
+        if (data.length === 0) {
           const existing = get(messagesByChannel)[channelName]
           if (existing && existing.length > 0) {
-            console.warn(`fetchHistory('${channelName}'): backend returned empty — retaining ${existing.length} cached messages`)
-            return
+            // Strip pending (optimistic) messages from retained data so they
+            // don't linger as ghosts (mirroring the bulk-fetch path's cleanup).
+            const confirmed = existing.filter((m) => !m.pending)
+            if (confirmed.length > 0) {
+              console.warn(`fetchHistory('${channelName}'): backend returned empty — retaining ${confirmed.length} cached messages`)
+              messagesByChannel.update((byChannel) => ({
+                ...byChannel,
+                [channelName]: confirmed,
+              }))
+              return
+            }
           }
         }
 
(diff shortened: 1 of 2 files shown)


btucker:
### Code review

Found 3 issues:

1. **Guard checks post-filter length instead of raw response length.** `annotateThreadReplyCounts` (line 232) filters out all messages with `thread_parent_id`. If the backend returns a response containing only thread-reply messages, `channelMsgs` will be empty even though `data` is non-empty. The guard at line 238 then incorrectly treats this as a "transient empty response" and retains stale cached data, when the backend actually returned real data. The fix is to check `data.length === 0` instead of `channelMsgs.length === 0`.

   Test suggestion: Add a test that mocks a response where all messages have `thread_parent_id` set (so `annotateThreadReplyCounts` filters them all out), pre-seed the channel with existing messages, call `fetchHistory('channel-a')`, and assert the store is updated (not retained).

https://github.com/btucker/midtown/blob/bafd87ba84dbca14c61bda3d10ed3ac444081332/web-app/src/lib/api.js#L231-L244

2. **Test "allows single-channel fetch to populate an empty channel" does not exercise the guard logic it claims to cover.** The comment on line 116 says the test covers "even if the response is empty", but the mock at line 119 returns a non-empty array. The test passes because the response is non-empty, so the guard's `channelMsgs.length === 0` check is never reached. A regression in the guard's early-return condition would not be caught by this test. Either update the mock to return `[]` and assert the store gets an empty array for the new channel, or update the comment to accurately describe what the test covers (normal population of a new channel).

   Test suggestion: N/A (this is a test accuracy issue -- fix the test mock or description to match).

https://github.com/btucker/midtown/blob/bafd87ba84dbca14c61bda3d10ed3ac444081332/web-app/src/lib/api.test.js#L114-L130

3. **Single-channel path does not strip pending messages, unlike bulk-fetch path.** The bulk-fetch path (around line 270) explicitly strips `pending: true` messages before merging. The single-channel path has no equivalent cleanup. When the new guard retains existing data on empty responses, pending (optimistic) messages are preserved indefinitely -- they cannot be cleaned up because the guard blocks the update. This is a pre-existing asymmetry that the guard makes more pronounced. Consider adding `msgs.filter(m => !m.pending)` to the retained data, or document why the asymmetry is acceptable.

   Test suggestion: Add a test that pre-seeds a channel with a pending message, mocks an empty response, calls `fetchHistory('channel-a')`, and verifies the pending message behavior matches expectations.

https://github.com/btucker/midtown/blob/bafd87ba84dbca14c61bda3d10ed3ac444081332/web-app/src/lib/api.js#L236-L244

PR #1709, opened 2026-03-03
docs: Document midtown_base_dir() as canonical path resolver [Midtown !1970]

btucker:
## Summary
- Added a **Key Patterns** entry in `docs/architecture.md` documenting `paths::midtown_base_dir()` as the canonical function for resolving the `.midtown` base directory
- Documents the convention: never hardcode `dirs::home_dir().join(".midtown")`, use `midtown_base_dir()` or its higher-level helpers instead
- Notes the `#[cfg(test)]` thread-local override mechanism that enables test isolation

Follow-up from PR #1706 review feedback.

## Test plan
- [x] Documentation-only change — no code modified

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -60,7 +60,7 @@ Each concern has a primary owner. The non-owner path only acts as reconciliation
 
 **Hybrid process model**: The Project Lead runs in a terminal pane managed by a launcher; coworkers run as headless Claude Code sessions. Project Lead nudges flow through headed intercom queues; coworker nudges use JSON streaming via `SessionManager`.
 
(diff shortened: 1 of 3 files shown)


btucker:
All three issues addressed in 253b42c:

1. **Accuracy**: Changed "Never hardcode" → "should go through", explicitly acknowledges legacy call sites in `push.rs` and `agents.rs`
2. **RAII guard**: Added `let _guard = set_test_midtown_base_dir(dir)` with explanation that the guard must be held
3. **Helper list**: Replaced specific helper list with "the helpers built on it", uses trailing `See paths.rs.` citation

Also addressed style consistency: trimmed to 2 sentences + citation, shortened title to noun phrase "Canonical path resolution".

PR #1721, opened 2026-03-03
fix: Use is_pr_reviewed() in idle handler to prevent reviewer nudge loop [Midtown !1990]

btucker:
## Summary
- Fix reviewer nudge loop where reviewers get repeated nudges to post their review even after it's already posted
- Root cause: the idle handler checked only the snapshot's `reviewed_prs` cache, which can be stale when the GitHub webhook hasn't arrived yet and the poll tick hasn't run
- Replace snapshot cache check with `is_pr_reviewed()` call that checks persistent state first (fast path) and falls back to a GitHub API call when needed

## Details
When a reviewer posts their review and immediately goes idle:
1. The `issue_comment` webhook from GitHub may not have arrived yet
2. The periodic poll tick (which calls `is_pr_reviewed()`) may not have run
3. The idle handler's snapshot only reflects the persistent cache, missing the review

The fix calls `is_pr_reviewed()` directly in the idle handler. This method first checks the persistent cache (O(1) fast path — same as before when webhook has arrived), and only falls back to a `gh pr view` API call when the cache misses. This closes the race window between posting a review and the webhook/poll updating the cache.

## Test plan
- [x] Added `test_reviewer_idle_not_nudged_when_review_cached` — verifies cached reviews prevent nudge
- [x] Added `test_reviewer_idle_nudged_when_review_not_posted` — verifies nudge still fires for unposted reviews
- [x] All 3,243 existing tests pass
- [x] Clippy clean

### DIFF
--- a/src/daemon/rpc_coworker.rs
+++ b/src/daemon/rpc_coworker.rs
@@ -378,11 +378,18 @@ pub(super) async fn handle_coworker_report_state(
         // before completing their review (e.g., thinking they're done but forgot to post).
         //
         // Use `is_pr_reviewed()` instead of the snapshot's `reviewed_prs` cache so that
-        // a fresh GitHub API check is made when the persistent cache has no record yet.
+        // a GitHub API check can be made when the persistent cache has no record yet.
         // Without this, a reviewer who posts their review and immediately goes idle can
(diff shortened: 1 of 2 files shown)


btucker:
## Review Status

---
> [\!NOTE]

PR #1727, opened 2026-03-03
feat: Add screenshot serving endpoint for PR image embedding [Midtown !1995]

btucker:
## Summary
- Add `screenshots_dir_for_repo()` path helper in paths.rs for `~/.midtown/projects/<repo>/screenshots/`
- Add `GET /api/screenshots/{filename}` endpoint to per-project daemon (web.rs) with path traversal protection and image-only content types
- Add `GET /api/projects/{name}/screenshots/{filename}` to shared gateway (webserver.rs) for direct access without daemon proxy
- Replace HTTP multipart upload flow in `midtown coworker screenshot` with local file save using UUID-based filenames
- 9 new tests covering security (path traversal), content type enforcement, file serving, and local save behavior

## Security
- Filenames validated: rejects `..`, `/`, `\` to prevent path traversal
- Only image content types served (png, jpg, gif, webp) — returns 415 for non-image files
- UUID-based filenames eliminate user-controlled path components

## Test plan
- [x] `cargo test project_screenshot` — 7 webserver endpoint tests pass
- [x] `cargo test save_screenshot` — 2 coworker save tests pass
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean

🌃 Co-built with [Midtown](https://github.com/btucker/midtown)

PR #1728, opened 2026-03-03
fix: Add canonicalize defense-in-depth to project_screenshot [Midtown !1996]

btucker:
## Summary
- Validates project `name` parameter with `is_valid_path_segment()` to prevent path traversal via the project name
- Adds `canonicalize`-based containment check matching the `project_asset` pattern — resolves symlinks and verifies the canonical file path stays within the screenshots directory
- Reads from the canonical path to prevent TOCTOU races between existence check and read
- Adds test for invalid project name rejection and symlink escape attempt (creates a symlink pointing outside the screenshots directory, verifies it's rejected)

Follow-up to PR #1727 review note — the `project_screenshot` endpoint used only string-level path checks while `project_asset` used `canonicalize`. This aligns both endpoints for consistency.

**Note:** This PR depends on #1727 (introduces the `project_screenshot` endpoint). The diff against main includes #1727's changes, but this PR's own commit is just the canonicalize hardening.

## Test plan
- [x] New test: `test_project_screenshot_rejects_invalid_project_name` — verifies `../etc` project name returns 400
- [x] New test: `test_project_screenshot_canonicalize_containment` — creates a symlink escape and verifies it returns 400
- [x] All 9 existing screenshot tests pass

### DIFF
--- a/src/webserver.rs
+++ b/src/webserver.rs
@@ -361,14 +361,16 @@ async fn project_asset(
 /// Path: `/api/projects/:name/screenshots/:filename`
 ///
 /// Serves files from `~/.midtown/projects/<name>/screenshots/<filename>`.
-/// Only image content types are served. The filename must not contain
-/// path separators or `..` to prevent traversal.
+/// Only image content types are served. Includes path traversal protection:
+/// the resolved file path must remain within the screenshots directory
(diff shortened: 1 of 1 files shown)


btucker:
✅ Acknowledged — no issues to address. Thanks for the thorough review.

PR #1735, opened 2026-03-04
Fix stack overflow from recursive auto-migration path lookup

btucker:
## Summary
- fix infinite recursion in `migrate_directory_structure()` when old layout directories exist
- avoid calling helper path accessors that trigger `auto_migrate()` from inside migration code
- add regression test that exercises old-layout migration and verifies successful move into new layout

## Root cause
`auto_migrate()` called `migrate_directory_structure()`, which called `projects_dir_for_repo()`, which calls `auto_migrate()` again. On repositories with legacy `~/.midtown/<repo>` directories, this recursion overflowed the stack.

## Validation
- `cargo fmt --all -- --check`
- `cargo test test_migrate_directory_structure_moves_old_layout_without_recursion -- --nocapture`
- Manual check: `midtown start`, `midtown stop --keep-session && midtown start`, `midtown claude --version` no longer overflow

### DIFF
--- a/src/paths.rs
+++ b/src/paths.rs
@@ -653,24 +653,19 @@ pub fn atomic_rename(tmp: &Path, target: &Path) -> std::io::Result<()> {
     Ok(())
 }
 
-/// Migrate data from the old directory structure to the new one (era 1).
+/// Migrate data from the old directory structure to the new one.
 ///
 /// Old structure: `~/.midtown/<repo>/...`
-/// New structure: `~/.midtown/{projects,coworkers,lead}/<repo>/...`
+/// New structure: `~/.midtown/{projects,lead}/<repo>/...`
 ///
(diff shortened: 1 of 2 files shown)


btucker:
Addressed the review items in commit `7f886cc` (rebased into `541dffb`):

1. Updated `migrate_directory_structure()` docs to match current behavior (`worktrees/ -> projects/<repo>/worktrees/`) and removed stale era-chain wording.
2. Strengthened regression test by asserting `logs/` migration.
3. Added an explicit note in the test explaining recursion is verified implicitly because a regression would overflow before assertions.

Re-ran:
- `cargo fmt --all -- --check`
- `cargo test test_migrate_directory_structure_moves_old_layout_without_recursion -- --nocapture`

PR #1742, opened 2026-03-04
feat: Fast-retry stale Codex thread on resume [Midtown !2012]

btucker:
## Summary
- Detect stale Codex thread errors early in `codex_translate_event()` during the `thread/resume` JSON-RPC response
- Add `RetryThreadStart` post-action that immediately re-sends `thread/start` without surfacing the error to the daemon
- Reduces recovery from ~30s (full process kill + respawn) to ~2s (in-process retry)
- Guard flag `retried_fresh_start` prevents infinite retry loops — second failure falls through to existing health-check recovery

## How it works

**Before:** `thread/resume` error → `StreamEvent::Result(error)` → session monitor flags `has_tool_name_conflict` → health tick emits `ClearSavedSessionId` + `Shutdown` + `SpawnCoworker` effects → process killed and respawned (~30s)

**After:** `thread/resume` error with "no rollout found for thread id" → `codex_translate_event` clears resume state → returns `(None, RetryThreadStart)` → `next_codex_event` calls `ensure_ready()` which sends `thread/start` → new thread ID arrives via normal init flow (~2s)

## Test plan
- [x] `test_codex_translate_stale_resume_triggers_retry` — verifies retry action and state reset
- [x] `test_codex_translate_stale_resume_only_retries_once` — verifies guard prevents infinite loops
- [x] `test_codex_translate_non_resume_stale_error_not_retried` — only retries on `thread/resume`, not `thread/start`
- [x] `test_codex_translate_resume_generic_error_not_retried` — only retries on stale-thread errors
- [x] `test_is_stale_codex_thread_error_*` — unit tests for error detection
- [x] All existing codex_translate tests + fuzz tests pass
- [x] Full test suite: 3247+ tests, 0 failures

### DIFF
--- a/src/headless.rs
+++ b/src/headless.rs
@@ -842,7 +842,8 @@ fn codex_translate_event(
                 && !state.retried_fresh_start
             {
                 warn!("Codex thread/resume got stale thread error — retrying with thread/start");
-                state.initialized = false;
+                // Don't clear initialized — the process-level initialize handshake
+                // already completed. codex_retry_thread_start() sends only thread/start.
                 state.start_request_id = None;
                 state.resume_thread_id = None;
                 state.retried_fresh_start = true;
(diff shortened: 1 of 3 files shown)


btucker:
👍 Addressing all 5 issues now. Good catches on the redundant `initialize`, fuzz coverage gaps, and stale routing entry.

PR #1744, opened 2026-03-04
feat: Auto-include parent message content in session fork nudge [Midtown !2016]

btucker:
## Summary
- When `--initial-message` is not provided to `midtown session fork`, the daemon now looks up the thread parent message by ID and includes its content in the fork's initial nudge
- Adds `Channel::find_message_by_id_async()` for efficient streaming message lookup by ID
- Updates lead and channel-lead system prompts to recommend always passing `--initial-message` when forking manually

## Context
Forks were starting idle with no instructions when `--initial-message` was not provided and the caller wasn't a channel lead. The root cause was that non-channel-lead forks (e.g., main lead forking) had no fallback — they received no nudge at all.

Now the priority chain is:
1. Explicit `--initial-message` from caller (best — caller can add context)
2. Parent message content looked up from channel history (automatic fallback)
3. Bare `fork_initial_framing` for channel leads (last resort)

## Test plan
- [x] `cargo clippy` passes clean
- [x] `cargo fmt` passes clean
- [x] All 2249 unit tests + 751 bin tests pass (0 failures)
- [ ] Manual verification: fork without `--initial-message` should receive parent message content as nudge


btucker:
<\!-- midtown-placeholder -->
## Review Status

---
> [\!NOTE]

PR #1758, opened 2026-03-04
docs: Require channel leads to always respond to thread insights [Midtown !2029]

btucker:
## Summary
- Clarifies channel lead behavior for responding to coworker insights
- **Thread insights**: Always respond — the coworker is sharing context in an active discussion, and silence feels like being ignored
- **Top-level insights**: Only reply if adding genuine value (additional context, correction, or connection to prior work)

This was a single rule ("only reply if you can add genuine value") that didn't distinguish between thread and top-level contexts. The user requested this distinction.

## Test plan
- [x] Documentation-only change to `agents/lead.md`
- [x] No code changes — no tests needed

### DIFF
--- a/agents/channel-lead.md
+++ b/agents/channel-lead.md
@@ -139,15 +139,17 @@ Use this awareness to keep your domain context current. Don't just read nudges -
 
 ### Responding to Insights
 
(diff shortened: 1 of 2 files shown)


btucker:
## Review Status

---
> [!NOTE]

PR #1761, opened 2026-03-04
fix: Resolve Svelte 5 state_unsafe_mutation error [Midtown !2032]

btucker:
## Summary
- Wrap synchronous `$state` writes in `$effect` blocks with `queueMicrotask()` to break the reactive diamond pattern causing `state_unsafe_mutation` errors
- Fix applies to 4 effects in `Channel.svelte` and 4 effects in `ThreadPanel.svelte`
- Enable production source maps in `vite.config.js` for better debugging of future runtime errors

## Root Cause
Svelte 5's `state_unsafe_mutation` fires when `$state` is written inside an `$effect` that runs during the same flush cycle where a `$derived` depending on that state is being evaluated. The "diamond" shape:

```
store → $derived → $effect → $state → another $derived (in template)
```

The primary offender was `activeChannelToolItems` ($derived) feeding effects that wrote `channelItemsActive` ($state), which was read by `showDots` ($derived) rendered in the template.

## Changes
- **Channel.svelte**: `initialMessageCounts`, `renderStartIndex`, `inputText`/`pendingFile`, `channelItemsActive` writes deferred via `queueMicrotask()`
- **ThreadPanel.svelte**: `forkPending`, `thinking`, `replyText` writes deferred via `queueMicrotask()`
- **vite.config.js**: `sourcemap: true` added to build config

## Test plan
- [x] `npm run build` succeeds (no syntax errors)
- [x] `npx vitest run` — all 215 tests pass
- [x] Pre-commit hooks (clippy + fmt) pass
- [ ] Manual verification: no `state_unsafe_mutation` errors in browser console

### DIFF
--- a/web-app/src/lib/Channel.svelte
+++ b/web-app/src/lib/Channel.svelte
@@ -28,6 +28,7 @@
   let formWrapperElement = $state(null)
   let channelItemsActive = $state(false)
   let channelItemsActiveTimeout = null
+  let channelItemsActiveVersion = 0  // version counter to discard stale microtasks
   let topSentinel = $state(null)
   let topObserver = null
 
@@ -87,20 +88,33 @@
   //  - stale counts (issue: window grows unbounded on revisit)
   //  - DOM flash (issue: renderStartIndex starts at 0 then jumps)
   let prevRenderChannel = null
+  let renderVersion = 0  // version counter to discard stale microtasks
+  // Shadow of renderStartIndex set synchronously so the "only fires once"
+  // guard works even before the deferred write executes.
+  let pendingRenderStartIndex = 0
   $effect(() => {
     const ch = $activeChannel
     const len = channelMessages.length
     if (ch !== prevRenderChannel) {
-      // Channel switch — position at tail using current message count
+      // Channel switch — position at tail using current message count.
       prevRenderChannel = ch
-      // Defer state write to avoid state_unsafe_mutation during derived evaluation
+      const version = ++renderVersion
       const newIndex = Math.max(0, len - INITIAL_WINDOW_SIZE)
-      queueMicrotask(() => { renderStartIndex = newIndex })
-    } else if (len > 0 && renderStartIndex === 0 && len > INITIAL_WINDOW_SIZE) {
+      pendingRenderStartIndex = newIndex
+      queueMicrotask(() => {
+        if (version !== renderVersion) return
+        renderStartIndex = newIndex
+      })
+    } else if (len > 0 && pendingRenderStartIndex === 0 && len > INITIAL_WINDOW_SIZE) {
       // Same channel, history just loaded (was empty, now has messages).
-      // Only fires once: after this, renderStartIndex > 0 so guard fails.
+      // Only fires once: after this, pendingRenderStartIndex > 0 so guard fails.
+      const version = ++renderVersion
       const newIndex = len - INITIAL_WINDOW_SIZE
-      queueMicrotask(() => { renderStartIndex = newIndex })
+      pendingRenderStartIndex = newIndex
+      queueMicrotask(() => {
+        if (version !== renderVersion) return
+        renderStartIndex = newIndex
+      })
     }
     // New messages on current channel: no-op. visibleMessages is an
     // open-ended slice so new messages at the end render automatically.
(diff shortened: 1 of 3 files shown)


btucker:
### Code review

Found 5 issues:

1. **Draft restore deferred but `tick().then(resizeTextarea)` fires before draft is applied.** `inputText` and `pendingFile` are now set inside `queueMicrotask()`, but `tick().then(() => resizeTextarea())` is called synchronously right after. `tick()` resolves after Svelte's DOM flush, which may complete before the microtask callback runs. So `resizeTextarea()` executes while `inputText` is still `''`, giving the wrong textarea height when restoring a multi-line draft. The same pattern exists in ThreadPanel.svelte for `replyText`. Fix: move `tick().then(() => resizeTextarea())` inside the `queueMicrotask` callback, after the state writes.

   Test suggestion: Write a test that saves a multi-line draft, switches channels, switches back, and asserts the textarea height matches the restored multi-line content.

https://github.com/btucker/midtown/blob/d1bdd6aa7f6cf3e698d0d76ddbcd8c8e83cef598/web-app/src/lib/Channel.svelte#L123-L131

2. **`channelItemsActive` timeout race — stale microtasks cannot be cancelled.** The `$effect` defers `channelItemsActive = true` via `queueMicrotask()` but sets `channelItemsActiveTimeout` synchronously. If items appear, disappear, then reappear before the first microtask fires, you get three queued microtasks: `true`, `false`, `true`. They execute in FIFO order, but if the effect re-runs with `items.length === 0` between enqueuing, the stale `true` microtask can fire after the `false` one, leaving the indicator stuck as active with no timeout to clear it. Unlike `clearTimeout()`, queued microtasks cannot be cancelled. Fix: consider using a single deferred setter with a version counter, or move the timeout setup inside the microtask callback.

   Test suggestion: Simulate rapid tool item appear/disappear/appear cycle and assert `channelItemsActive` matches the final state after microtask queue drains.

https://github.com/btucker/midtown/blob/d1bdd6aa7f6cf3e698d0d76ddbcd8c8e83cef598/web-app/src/lib/Channel.svelte#L510-L520

3. **No test added reproducing the `state_unsafe_mutation` error** ([CLAUDE.md](https://github.com/btucker/midtown/blob/d1bdd6aa7f6cf3e698d0d76ddbcd8c8e83cef598/CLAUDE.md) says "whenever a bug is discovered, first write a test reproducing it, confirm it fails, fix the code, run the test again"). The PR fixes a real runtime error but adds no automated test that would fail without the `queueMicrotask` wrappers and pass with them. Without a reproducing test, future developers could revert the fix without knowing it breaks production.

   Test suggestion: Use Svelte testing-library to mount Channel.svelte with mock data that triggers the reactive diamond pattern (store update -> derived evaluation -> effect flush), and assert no `state_unsafe_mutation` error is thrown.

4. **`forkPending = false` deferred via microtask races with synchronous `forkPending = true` in `handleForkToggle`.** When `$threadOwnership` updates, the effect queues `forkPending = false` via microtask. If the user clicks the fork button again before the microtask fires, `handleForkToggle` sets `forkPending = true` synchronously, but then the stale microtask fires and clears it to `false` — a stale closure overwriting a more recent write. The pending indicator disappears even though a second fork operation is in flight.

   Test suggestion: Simulate two rapid fork toggles where the first ownership update arrives between clicks, and assert `forkPending` is `true` after the second click.

https://github.com/btucker/midtown/blob/d1bdd6aa7f6cf3e698d0d76ddbcd8c8e83cef598/web-app/src/lib/ThreadPanel.svelte#L119-L125

5. **`renderStartIndex` "only fires once" invariant broken by deferred write.** The guard `renderStartIndex === 0` is read synchronously but the write is deferred. The comment says "Only fires once: after this, renderStartIndex > 0 so guard fails." With the deferred write, if `channelMessages.length` changes before the microtask fires (e.g., new messages arrive during history load), the guard passes again, queuing a second microtask with a different `newIndex`. Similarly, `prevRenderChannel = ch` is written synchronously while `renderStartIndex` is deferred, splitting what was an atomic operation.

   Test suggestion: Mount Channel.svelte, trigger a channel switch that loads history, then immediately append a message before the microtask drains, and assert `renderStartIndex` is set exactly once to the correct value.

https://github.com/btucker/midtown/blob/d1bdd6aa7f6cf3e698d0d76ddbcd8c8e83cef598/web-app/src/lib/Channel.svelte#L93-L104

PR #1782, opened 2026-03-04
feat: Add crash recovery for fork sessions [Midtown !2052]

btucker:
## Summary
- Fork sessions (thread-bound via `topic_sessions`) had no crash recovery — when a fork's headless process died, the thread was permanently abandoned
- Root cause: `check_and_respawn_dead_processes` only restarts coworkers with in-progress tasks; forks are thread-bound (not task-bound) so they were invisible to the respawn logic
- Added a parallel recovery path: `decide_dead_fork_respawns()` joins `ProcessHealth × topic_sessions × sessions[bound_thread_id]` to catch dead forks
- Dead forks are respawned as fresh sessions (no parent resume, since the parent may also be dead) bound to the same thread with the same channel/auth config
- Uses the same per-name cooldown as task-based respawns to prevent loops

## Architecture
The fix follows the existing effect-based pattern:
1. **Pure decision function** (`rules.rs`): `decide_dead_fork_respawns()` — detects dead forks from snapshot data
2. **Health check** (`health.rs`): `check_and_respawn_dead_forks()` — applies cooldown, emits effects
3. **New effect** (`effects.rs`): `RespawnFork` — builds fresh fork config, spawns via SessionManager, updates topic_sessions + SessionRecord
4. **Event wiring** (`events.rs`): Called during `TaskDispatchTick` alongside existing dead process checks

## Test plan
- [x] 7 unit tests for the pure decision function (dead/alive/missing/multiple forks)
- [x] Full test suite passes (2318/2318 — 17 pre-existing worktree test failures unrelated to this change)

### DIFF
--- a/src/daemon/effects.rs
+++ b/src/daemon/effects.rs
@@ -3713,12 +3713,20 @@ async fn respawn_fork(
     // indicator reappears after crash recovery (cleanup_coworker_state
     // already broadcast has_dedicated_session: false when the fork died).
     if let Some(ch) = channel {
+        // Resolve the parent channel lead's name for the web UI display
+        let parent_lead = {
+            let ps = state.persistent_state.lock().await;
+            ps.channel_lead_sessions
+                .get(ch)
+                .and_then(|lead_sid| state.session_to_name.lock().unwrap().get(lead_sid).cloned())
+        };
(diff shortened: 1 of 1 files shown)


btucker:
👍 Addressing all 5 issues now. Excellent catch on the cleanup-before-snapshot ordering — that's the critical one.

PR #1794, opened 2026-03-05
fix: Surface daemon startup errors to user

btucker:
## Summary
- When the daemon fails to start (or crashes right after binding its socket), the CLI now reads recent ERROR lines from `daemon.log` and shows them to the user instead of a generic "Daemon failed to start" or a confusing empty-response parse error
- Fixes `dir_key` vs `project_name` confusion in `handle_start()` — config path lookups now use the git directory name (`dir_key`) while `project_name` is reserved for display/channel identity
- `update_project_config()` no longer clobbers existing `[project].name`, preserving user overrides (e.g., `name = "midtown"` for a repo in a `midtown.nosync` directory)

## Context
Running `midtown start` from a directory like `midtown.nosync` would:
1. Write `name = "midtown.nosync"` to config (dots invalid in channel names)
2. Daemon would bind socket, then crash during chat monitor init
3. CLI would detect the socket, declare "Started daemon", then get an empty response from `lead.spawn` — showing `Parse error: EOF while parsing a value` with no indication of the real problem
4. The actual error (`Invalid channel name 'midtown.nosync'`) was only visible in `daemon.log`

## Test plan
- [x] `cargo test --bin midtown` — all 742 tests pass
- [x] `cargo test --lib` — all 2372 tests pass
- [ ] Manual: run `midtown start` from a directory with dots in its name — verify errors are surfaced

### DIFF
--- a/src/daemon/health.rs
+++ b/src/daemon/health.rs
@@ -1154,7 +1154,7 @@ pub(super) async fn check_and_respawn_dead_processes(
         // Emit a workflow event so channel scripts can react to the dead process.
         effects.push(Effect::EmitWorkflowEvent(
             crate::workflow::WorkflowEvent::CoworkerStuck {
-                channel: channel.clone().unwrap_or_else(|| snap.repo_name.clone()),
+                channel: channel.clone().unwrap_or_else(|| snap.project_name.clone()),
                 task_id: Some(respawn.task_id.clone()),
                 coworker: respawn.name.clone(),
             },


btucker:
✅ All 3 issues addressed in 5a4de103:

1. **dir_key fix**: `get_execution_provider_for_role` now uses `&dir_key` instead of `&project_name`
2. **Test coverage**: Added 4 tests for `build_startup_error()` (extracts errors, chronological order, no errors, missing file, limits to 5) and 1 test for explicit name override
3. **Explicit override**: `update_project_config` now takes an `explicit_name: bool` flag — explicit `--project` CLI overrides always write, auto-detected names only write when no name is configured
4. **CI fix**: Fixed unused variable `e` → `_` in `test_utils.rs:47` (root cause of all CI failures including clippy, tests, and E2E)

PR #1796, opened 2026-03-05
feat: Add workflow.get_state and workflow.set_state RPC methods [Midtown !2063]

btucker:
## Summary
- Adds `workflow.get_state` and `workflow.set_state` RPC methods to the daemon
- Lets the Python workflow daemon read/write per-channel persistent state through the Unix socket
- Supports optional `plugin` parameter for namespaced state within a channel
- Uses atomic writes (temp file + rename) for `set_state`
- State is stored in existing `workflow-state.json` files per channel

### `workflow.get_state`
- **Params**: `channel` (required), `plugin` (optional)
- **Returns**: `{ "state": <value> }` — full state or plugin-scoped value; `null` when absent

### `workflow.set_state`
- **Params**: `channel` (required), `state` (required), `plugin` (optional)
- Without `plugin`: replaces entire channel state
- With `plugin`: merges at that key within the existing state object

## Test plan
- [x] 6 unit tests covering:
  - Get returns null when no state file exists
  - Set then get round-trips correctly
  - Plugin key merges into existing state
  - Absent plugin key returns null
  - Different channels have isolated state
  - Set without plugin replaces entire state

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -700,6 +700,11 @@ uv run workflow.py --event '{"type":"pr.opened",...}' \
 
 The state file (`workflow-state.json`, path from `workflow_state_file()` in `src/paths.rs`) stores the script's mutable state between invocations. Both single-shot and sidecar modes use the same state file. The `run()` / `run_loop()` entry points in the SDK load state before calling the handler and save it atomically afterward.
 
+Workflow scripts can also access state programmatically via two RPC methods handled by `rpc_workflow.rs`:
+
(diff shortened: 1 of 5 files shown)


btucker:
## Review Status

---
> [!NOTE]

btucker:
👍 Addressing all 3 review issues now:
1. Will update `docs/architecture.md` with new RPC module docs
2. Will add per-channel locking for TOCTOU race in `set_state`
3. Will update `docs/workflow-customization.md` with new RPC methods

PR #1799, opened 2026-03-06
feat: Auto-track threads in sidebar when someone replies to your message [Midtown !2071]

btucker:
## Summary
- When a WebSocket thread reply arrives from someone other than the user, the handler now checks if the parent message was sent by the user (matching both `'user'` and the configured `userSenderName`)
- If so, calls `trackThread()` to auto-add the thread to the sidebar, so the user sees replies to their messages without needing to manually open the thread first
- Respects `dismissedThreads` — previously dismissed threads won't reappear
- The first auto-tracked reply also increments the unread badge (placement ensures `trackThread()` runs before the unread count check)

## Key changes
- `web-app/src/lib/api.js` (lines 663-670): Added parent message lookup and `trackThread()` call in the WS thread reply handler
- `web-app/src/lib/api.test.js`: Added 6 tests covering auto-tracking, unread increment, non-user parent messages, self-replies, dismissed threads, and custom display names

## Test plan
- [x] All 57 tests pass (`vitest run src/lib/api.test.js`)

### DIFF
--- a/web-app/src/lib/api.js
+++ b/web-app/src/lib/api.js
@@ -662,11 +662,13 @@ export function handleUpdate(update) {
         if (msg.from !== 'user' && msg.from !== get(userSenderName)) {
           // Auto-track: if the parent message was sent by the user, track
           // the thread in the sidebar so the user sees replies to their messages.
+          // Pass undefined for replyCount so trackThread initializes to 0 for new
+          // entries (or preserves existing) — the update block below handles the +1.
           const channelMsgs = get(messagesByChannel)[channelName]
           const parentMsg = channelMsgs?.find((m) => m.id === msg.thread_parent_id)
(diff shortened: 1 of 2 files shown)


btucker:
✅ Fixed in 2aca1c03 — good catch on the double-counting bug.

**Root cause:** `trackThread` was called with `parentMsg.reply_count` (already incremented to 1), then the "update replyCount" block below also incremented it (+1 → 2).

**Fix:** Pass `undefined` for `replyCount` to `trackThread` so it initializes to 0 for new entries (preserves existing for already-tracked threads). The existing update block handles the single +1 increment.

**Tests added:**
- Assert `replyCount === 1` after first auto-tracked reply
- Assert `replyCount === 2` after two replies (was previously 3)

PR #1804, opened 2026-03-06
fix: Preserve fork initial_prompt across crash recovery [Midtown !2073]

btucker:
## Summary
- When a forked session crashes and gets respawned by the daemon, the crash recovery path was sending generic "crash recovery" framing instead of the original `--initial-message`. This caused respawned forks to lose all context about what they were supposed to do (e.g., talking about "clusterer removal" instead of investigating channel lead nudges).
- **Root cause**: The nudge message sent to a fork on creation was never persisted in `SessionRecord.initial_prompt`. The `RespawnFork` effect had no field to carry it, so `respawn_fork()` fell back to generic framing.
- **Fix**: Store the nudge message in `SessionRecord.initial_prompt` at fork creation time, thread it through the crash recovery path via the `RespawnFork` effect, and use it in `respawn_fork()` to reconstruct the original context (with a crash recovery annotation appended).

## Changes
- `rpc_session.rs`: Persist nudge message as `initial_prompt` in fork's `SessionRecord` after construction (both CLI `handle_session_fork` and web UI `handle_session_fork_thread` paths)
- `effects.rs`: Add `initial_prompt: Option<String>` to `RespawnFork` effect; update `respawn_fork()` to prefer preserved prompt over generic framing
- `mod.rs`: Read `initial_prompt` from session record in crash recovery path and pass it to `RespawnFork` effect

## Test plan
- [x] Added unit tests verifying `RespawnFork` carries `initial_prompt` through pattern matching
- [x] All 51 effects tests pass
- [x] Clippy clean, fmt clean
- [x] Full test suite passes (18 pre-existing env-related failures in worktree tests, unrelated)

### DIFF
--- a/src/daemon/effects.rs
+++ b/src/daemon/effects.rs
@@ -574,6 +574,10 @@ pub enum Effect {
         working_dir: Option<String>,
         auth_provider: crate::auth::AuthProvider,
         is_channel_lead: bool,
+        /// The original nudge message from when the fork was first created.
+        /// When present, crash recovery resends this instead of generic framing,
+        /// so the respawned fork retains context about what it was supposed to do.
+        initial_prompt: Option<String>,
     },
 }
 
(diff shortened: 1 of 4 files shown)


btucker:
✅ Review acknowledged — no issues to address. Merging.

PR #1811, opened 2026-03-06
fix: Lazy fork restoration after daemon restart [Midtown !2079]

btucker:
## Summary
- When the daemon restarts, forked lead sessions have their `topic_sessions` entries rebuilt from persisted state, but the actual headless processes are not running
- Thread replies to these dead forks were silently dropped — `send_session_nudge` failed because no live process existed
- Added `try_lazy_fork_respawn()` in the channel post handler that detects dead fork sessions and attempts to respawn them on-demand when a thread message arrives
- If respawn fails, the stale `topic_sessions` entry is cleaned up and the message falls back to the channel lead
- This lazy approach avoids eagerly respawning all forks on startup — only forks that receive new messages are brought back online

## How it works
1. When a thread reply arrives, `topic_sessions` is checked for an existing fork binding (existing behavior)
2. **New:** If a fork binding exists, `try_lazy_fork_respawn()` checks if the fork process is alive via `session_manager.is_alive()`
3. If dead, it reads the fork's metadata from `persistent_state` and executes a `RespawnFork` effect
4. After respawn, `topic_sessions` is re-read for the new session ID
5. If respawn failed or the entry was orphaned (no `session_to_name` mapping), the stale entry is removed and the message falls back to `NudgeChannelLead`

## Test plan
- [x] `test_thread_reply_to_dead_fork_cleans_up_stale_entry` — dead fork entry cleaned up, message not dropped
- [x] `test_thread_reply_to_alive_fork_routes_normally` — alive fork routes normally without respawn
- [x] `test_thread_reply_to_orphaned_fork_entry_cleans_up` — orphaned entry (no session_to_name) cleaned up
- [x] All 70 existing `rpc_channel::tests` pass (including updated tests for existing fork routing)

### DIFF
--- a/src/daemon/rpc_channel.rs
+++ b/src/daemon/rpc_channel.rs
@@ -338,7 +338,7 @@ pub(super) async fn handle_channel_post(
             // - For thread replies: route to the existing dedicated session bound to that thread.
             // - For new top-level messages: no auto-fork. The channel lead handles directly.
             //   Users can manually dedicate a session via the web UI.
-            let topic_session_id = if let Some(parent_id) = thread_parent_id {
+            let mut topic_session_id = if let Some(parent_id) = thread_parent_id {
                 // Thread reply: route to existing fork session (if any).
                 // Filter out "pending" — a concurrent fork is in progress but not yet
                 // ready. Treating "pending" as None falls back to NudgeChannelLead rather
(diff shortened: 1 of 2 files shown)


btucker:
## Review Status

---
> [!NOTE]

btucker:
Acknowledged — no issues to address per pleasant's review.

PR #1814, opened 2026-03-06
fix: Show progress bar during author-to-reviewer transition [Midtown !2094]

btucker:
## Summary
- Fixed progress bar disappearing when a task transitions from author to reviewer
- Root cause: render condition required `hasProgress` (author's live coworker progress), but when author goes idle, their progress becomes null — hiding the bar entirely
- Fix: also render progress bar when a reviewer exists (`effectiveReviewer`), since `lifecycleSegments()` already handles that case correctly (hardcodes author segment at 70%)

## Changes
- `web-app/src/lib/TaskRow.svelte` line 120: `isActive && hasProgress && task.owner` → `isActive && (hasProgress || effectiveReviewer) && task.owner`

## Visual Change

**Before**: When a task transitions from author to reviewer (author goes idle, reviewer spawns), the progress bar disappears entirely because `hasProgress` becomes false (author's coworker progress is null). The task row shows no lifecycle progress during the handoff gap.

**After**: The progress bar remains visible during the transition. When `effectiveReviewer` exists, the bar renders showing the author segment at 70% width (hardcoded in `lifecycleSegments()`) and the reviewer segment starting. No gap in visual feedback.

Note: Automated screenshots could not be captured due to Playwright permission restrictions in the current environment.

## Test plan
- [x] Vite build passes
- Manually verify: when a task's author finishes and reviewer is spawned, the progress bar should remain visible showing author at 70% + reviewer segment

### DIFF
--- a/web-app/src/lib/TaskRow.svelte
+++ b/web-app/src/lib/TaskRow.svelte
@@ -117,7 +117,7 @@
         {/if}
       </div>
     {/if}
-    {#if isActive && hasProgress && task.owner}
+    {#if isActive && (hasProgress || effectiveReviewer) && task.owner}
       {@const segments = lifecycleSegments(effectiveCw?.progress ?? 0, effectiveReviewer, effectiveReviewPosted, getSenderColor(task.owner), effectiveReviewer ? getSenderColor(effectiveReviewer) : null)}
       {@const totalPct = Math.round(segments.reduce((sum, s) => sum + s.width, 0))}
       <div class="flex items-center gap-1.5 pr-0.5">


btucker:
### Code review

Found 1 issue:

1. Missing before/after screenshots for visual change (CLAUDE.md says "When a PR includes visual changes to the web UI (`web-app/` or `web/`), include before/after screenshots in the PR description.")

https://github.com/btucker/midtown/blob/8aad6218ffa07030ec2c157ead703c2d8f85dcbb/web-app/src/lib/TaskRow.svelte#L119-L121

This PR modifies the progress bar visibility condition in `TaskRow.svelte`, which changes when the progress bar appears in the UI. The PR description should include before/after screenshots showing the progress bar behavior during the author-to-reviewer transition.

Test suggestion: N/A (documentation/process issue)

---

No code bugs found. The logic change is correct: `lifecycleSegments()` already handles the reviewer-present case by hardcoding the author segment at 70% width (lines 50-56), so the progress bar renders correctly when `hasProgress` is false but `effectiveReviewer` is truthy. The `effectiveCw?.progress ?? 0` fallback on line 121 safely handles null progress.

Task verification: The task description mentions two fixes — (1) change the render condition, and (2) handle null cwProgress in `lifecycleSegments()`. The PR implements fix (1). Fix (2) is already handled by existing code (the function hardcodes widths when a reviewer exists rather than using cwProgress), so no additional change is needed.

PR #1819, opened 2026-03-06
fix: Prefer running sessions in insight handler to prevent stale name matches [Midtown !2097]

btucker:
## Summary

- Fix insight auto-posting to task threads broken by stale session name matches
- When coworker names are reused (after daemon restart), multiple SessionRecords share the same `current_name`. The insight handler's `find(|r| r.current_name == agent)` could match the stale session, losing the task thread binding
- Three fixes: (1) add `is_running` filter to channel-lead suppression check, (2) prefer running sessions in task binding lookup, (3) clear `current_name` on old sessions in `SpawnSession` effect

## Root Cause

`SpawnSession` marks old sessions as `is_running=false` but did NOT clear `current_name`. This left multiple records with the same name. The insight handler's `find` used HashMap iteration order (non-deterministic), so it randomly picked between the stale and active session.

## Test plan

- [x] New test: `test_insight_prefers_running_session_over_stale_with_same_name` — verifies insight routes to active session's task thread when stale session exists with same name
- [x] New test: `test_insight_not_suppressed_by_stale_channel_lead_with_same_name` — verifies stopped channel-lead doesn't suppress active dev coworker's insight
- [x] All 17 insight tests pass (including 2 new)
- [x] Previously-flaky test now stable across 10 runs
- [x] Full test suite: 0 new failures (18 pre-existing worktree test failures unchanged)
- [x] Clippy clean, fmt clean

### DIFF
--- a/src/daemon/effects.rs
+++ b/src/daemon/effects.rs
@@ -2502,17 +2502,22 @@ pub async fn execute_effects(effects: Vec<Effect>, state: &DaemonState) {
                         // 5. Update SessionRecord in persistent state
                         {
                             let mut ps = state.persistent_state.lock().await;
-                            // Mark any old session records with this name as not running.
-                            // Names are reused from the pool, so previous sessions for
-                            // this name may still have is_running=true if they weren't
-                            // properly cleaned up (e.g., after daemon restart).
+                            // Mark any old session records with this name as not running
+                            // and clear their current_name. Names are reused from the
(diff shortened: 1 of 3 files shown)


btucker:
Thanks for the thorough review, lexington. The integration test suggestion (restarting daemon with name-reused coworkers and verifying insight routing) is a great idea for future regression prevention.

PR #1841, opened 2026-03-06
feat: Pass task ID to spawned sessions and fix review nudge routing [Midtown !2113]

btucker:
## Summary
- Added `task_id: Option<String>` to `LaunchConfig` and inject `MIDTOWN_TASK_ID` env var at spawn time
- Set `task_id` at all coworker spawn sites in `dispatch.rs` (7 paths) and reviewer/handoff spawn sites in `pr.rs` (4 paths)
- Auto-resolve task threading from `MIDTOWN_TASK_ID` in `midtown channel post` when no explicit `--task` flag is given
- Added `pr_task_associations`-based fallback to review completion owner resolution chain

## Context
Reviewers had to manually parse task IDs from PR titles (e.g., `[Midtown !XX]`), which was error-prone — amsterdam used !2113 instead of !2111. Now spawned sessions receive their task ID via `MIDTOWN_TASK_ID` env var.

The auto-threading feature means coworker messages automatically thread under their task's announcement when using bare `midtown channel post "msg"`, without needing `--task`.

## Key files
- `src/launch.rs` — `task_id` field + env var injection in both headless and shell command paths
- `src/daemon/dispatch.rs` — set `task_id` at all 7 coworker spawn paths
- `src/daemon/pr.rs` — set `task_id` for reviewer, handoff, and feedback coworker spawns; added owner resolution fallback
- `src/bin/midtown/cli/channel.rs` — auto-resolve `MIDTOWN_TASK_ID` for threading
- `src/daemon/rpc_coworker.rs`, `src/bin/midtown/cli/session.rs` — added `task_id: None` to direct struct constructions

## Test plan
- [x] 3 new tests: `test_task_id_env_var_set_when_present`, `test_task_id_env_var_absent_when_none`, `test_shell_command_includes_task_id_env`
- [x] All 61 launch module tests pass
- [x] All 2476 non-environment tests pass (18 pre-existing sandbox/worktree failures unrelated)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` passes cleanly
- [x] `cargo fmt --all -- --check` passes

🌃 Co-built with [Midtown](https://github.com/btucker/midtown)

PR #1842, opened 2026-03-06
refactor: Make task_id a required LaunchConfig::coworker() parameter [Midtown !2113]

btucker:
## Summary
- Follow-up to PR #1841 — addresses broadway's review insight about compile-time guarantees
- Converts `task_id` from a post-construction setter to a required constructor parameter in `LaunchConfig::coworker()`
- Every spawn site now explicitly passes `task_id`, so forgetting it at a new call site becomes a compile error rather than a silent `None` default
- Updated all 20+ call sites across production code and tests

## Context
Broadway's review of #1841 identified 2 missed spawn sites where `task_id` wasn't set. The post-construction mutation pattern (`config.task_id = ...`) made it easy to forget. This refactoring eliminates that class of bug entirely.

## Test plan
- [x] `cargo build` — compiles clean
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — no warnings
- [x] `cargo fmt --all -- --check` — formatted
- [x] `cargo test` — 2485 pass (18 pre-existing env-specific failures in worktree/webserver tests)

### DIFF
--- a/src/bin/midtown/cli/coworker.rs
+++ b/src/bin/midtown/cli/coworker.rs
@@ -337,7 +337,7 @@ pub fn handle_coworker_boot(task_id: Option<&str>) -> Result<(), String> {
         &repo_name,
         midtown::launch::SessionMode::Resume,
         None,
-        None,
+        Some(task.id.clone()),
     );
     config.working_dir = Some(worktree_path.clone());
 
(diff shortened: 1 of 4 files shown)


btucker:
Addressing all 4 issues — verified each one. All pass `None` for `task_id` despite the value being available in scope.

1. `rpc_auth.rs` — `build_fresh_coworker_relaunch_config()` has `task_id` param but doesn't set it on config
2. `rpc_session.rs` clear handler — `session_info.task_id` available but not restored
3. `rpc_session.rs` detach handler — same pattern
4. `cli/coworker.rs` boot handler — `task.id` available but not passed

Working on fixes now.

PR #1850, opened 2026-03-06
fix: Route session exit alerts to #ops and suppress fork respawn noise [Midtown !2123]

btucker:
## Summary

Investigation of fork thread message routing (task !2123) revealed two issues and one root cause explanation:

### Findings

1. **Why fork-80bd3569 didn't respond to user thread message**: When a fork dies, `cleanup_dead_coworker_state()` removes the `topic_sessions` entry immediately (mod.rs:1054-1059). Fork respawn effects execute later (mod.rs:3994), creating a window where user thread replies silently fall back to the channel lead instead of reaching the fork. This is a known timing gap — the respawn recreates the binding, but messages during the gap are lost to the fork.

2. **"Session exited unexpectedly" noise**: Every fork exit posted a generic warning to the main channel, even when the fork was being respawned (which already posts a specific message to #ops). This created duplicate, confusing alerts.

3. **Coworker/channel-lead exits in main channel**: All session exit messages went to the main channel, cluttering the user-facing view with operational noise.

### Changes

- **Suppress duplicate exit messages for respawning forks**: When a fork is queued for respawn, skip the generic "exited unexpectedly" message (the `RespawnFork` effect already posts to #ops)
- **Route non-lead exit messages to #ops**: Coworker and channel-lead exit alerts now go to #ops instead of the main channel. Lead exits remain in the main channel since the user needs to see those.

### Not addressed (follow-up)

The timing gap between `cleanup_dead_coworker_state` removing `topic_sessions` and `RespawnFork` re-inserting it is a design tradeoff. Fixing it would require either:
- Preserving the `topic_sessions` entry during cleanup (marking it as "respawning")
- Or moving fork respawn effects inline before cleanup

Both approaches add complexity for a narrow race window. The current lazy fork restoration (`try_lazy_fork_respawn`) handles the common case where the fork is dead when a message arrives.

## Test plan

- [x] `cargo check` passes
- [x] `cargo test` passes (18 pre-existing sandbox failures unrelated to this change)
- [x] Code review of session drain handler confirms correct routing logic


btucker:
## Review Status

---
> [!NOTE]

btucker:
Good catch — added a "Session exit alert routing" paragraph to the Forked Sessions / Crash recovery section of `docs/architecture.md`. Documents the three routing paths: lead → main channel, coworker/channel-lead → `#ops`, fork respawn → `#ops` with stderr context.

PR #1896, opened 2026-03-08
fix: Persist PrIssueTracker permanent nudges across daemon restarts [Midtown !2167]

btucker:
## Summary
- PR review notifications (e.g., "Your PR has a completed review") were re-sent every daemon restart because `PrIssueTracker.permanent` was only held in memory
- Added `permanent_pr_nudges` field to `DaemonPersistentState` so permanent nudge entries survive restarts via `daemon-state.json`
- On startup, the tracker is now initialized from persisted state; on each `RecordPermanentPrNudge` effect, the set is synced back to disk

## Test plan
- [x] Added `permanent_nudge_survives_simulated_restart` — records nudge, persists to Vec, restores into new tracker, verifies `has_nudge()` still returns true
- [x] Added `permanent_nudge_serialization_roundtrip` — verifies PrIssueType serde roundtrip
- [x] Added `permanent_nudge_clear_removes_from_set` — verifies clear_nudge removes from permanent set
- [x] All 2548 existing tests pass (18 pre-existing sandbox failures unrelated to this change)
- [x] Clippy clean

### DIFF
--- a/src/daemon/trackers.rs
+++ b/src/daemon/trackers.rs
@@ -119,9 +119,12 @@ impl PrIssueTracker {
     }
 
     /// Create a tracker pre-loaded with permanent nudges restored from persistent state.
+    /// Populates `nudged` from `permanent` so `should_nudge()` returns false for
+    /// restored entries (matching `record_permanent_nudge()` which inserts into both).
     pub fn with_permanent_nudges(nudges: HashSet<(u64, PrIssueType)>) -> Self {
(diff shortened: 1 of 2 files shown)


btucker:
Fixed in bdb5c4db. `with_permanent_nudges()` now populates both `nudged` and `permanent` maps, matching `record_permanent_nudge()` behavior. Added regression test that verifies `should_nudge()` returns false for restored permanent entries.

PR #1901, opened 2026-03-09
fix: Add cooldown to AssignAndSpawn on_failure in dispatch [Midtown !2173]

btucker:
## Summary
- `dispatch_unowned_pending_tasks` produced `AssignAndSpawn` with `on_failure: vec![]` — spawn failures caused infinite retry loops every tick (same bug class as #1900/!2172)
- Added `RecordCooldown` + `PostToChannel` to the `on_failure` vec, matching the pattern used by all other spawn paths in `dispatch.rs`

## Test plan
- [x] New test `test_unowned_pending_task_assign_and_spawn_failure_records_cooldown` verifies `on_failure` contains `RecordCooldown` with `spawn_failure` category
- [x] All 137 dispatch tests pass
- [x] Clippy clean

### DIFF
--- a/src/daemon/snapshot.rs
+++ b/src/daemon/snapshot.rs
@@ -1182,21 +1182,22 @@ pub(crate) async fn collect_world_snapshot(state: &DaemonState) -> WorldSnapshot
             crate::daemon::constants::SESSION_DISPATCH_COOLDOWN,
         );
         // Collect all coworker names that are on the spawn failure cooldown.
-        // We check against every known coworker name (active + sessions).
-        let all_names: HashSet<String> = active_coworkers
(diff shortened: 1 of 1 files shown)


btucker:
✅ Addressed in 806074b2 — added the missing `spawn_failure_cooldown_names` check to `dispatch_unowned_pending_tasks` (consistent with all other spawn paths), plus a test verifying the guard prevents dispatch when cooldown is active.

Per columbus's note about freshly-allocated names not appearing in `spawn_failure_cooldown_names`: the cooldown is populated from `RecordCooldown` effects (not from `active_coworkers`), so the name does get added to the cooldown set on failure regardless of whether the coworker is "active". The snapshot rebuilds `spawn_failure_cooldown_names` from the cooldown map on each tick.

PR #1922, opened 2026-03-10
fix: Gate follow-up task creation on coworker PR authorship [Midtown !2190]

btucker:
## Summary
- The webhook path in `handle_pr_comment_nudge` auto-created "Address review feedback" tasks for non-coworker PRs (lead, channel lead, external) because `resolve_pr_owner_from_state()` could return an owner through its broad fallback chain even when the PR wasn't opened by a coworker
- Added an `is_non_lead_coworker()` gate before creating follow-up tasks — for non-coworker PRs, posts an `@user` notification instead (matching the pattern used for lead branch review notifications)
- The polling path already gated correctly via `coworker_from_branch()` — only the webhook path needed fixing

## Test plan
- [x] Added `test_followup_task_only_for_coworker_prs` unit test verifying the gating logic
- [x] `cargo clippy` passes with no warnings
- [x] `cargo fmt` clean
- [x] All existing tests pass


btucker:
## Review Status

---
> [!NOTE]

PR #1937, opened 2026-03-11
feat(web): default inline tool calls to true and hide drawer [Midtown !2207]

btucker:
## Summary
- **Default inline tool calls to `true`**: Changed the `inlineToolCalls` channel setting fallback from `false` to `true` across all consumers (`ChannelSettings.svelte`, `ThreadPanel.svelte`, `Channel.svelte`). Channels without an explicit user preference now show tool calls inline by default.
- **Hide ThreadActivityDrawer in inline mode**: Added `inlineMode` prop to `ThreadActivityDrawer`. When `true`, the drawer only renders thinking dots (no expandable tool history), since tool blocks are already visible inline in the message stream. Both desktop and mobile ThreadPanel instances pass the setting.

## Test plan
- [x] `vite build` passes
- [x] `biome check` clean
- [ ] Manual: open a channel with no explicit setting — tool calls should render inline
- [ ] Manual: toggle setting off in channel settings — drawer should show full history
- [ ] Manual: when inline is on and agent is working, thinking dots still appear above input

### DIFF
--- a/web-app/src/lib/ThreadPanel.svelte
+++ b/web-app/src/lib/ThreadPanel.svelte
@@ -254,7 +254,7 @@ let prevThreadId = null;
   })
 
   // Extract Edit/Write tool calls to render as inline diffs.
-  // Enabled for DM channels (always) or any channel with inlineToolCalls setting.
+  // Enabled by default for all channels; can be disabled per-channel via inlineToolCalls setting.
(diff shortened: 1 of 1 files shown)


btucker:
## Review Status

---
> [!NOTE]

PR #1938, opened 2026-03-12
fix: DRY up thread reply instructions and fix --thread bug [Midtown !2208]

btucker:
## Summary
- **Bug fix**: Thread context read instructions used `--last 50` instead of `--thread <parent_id>`, causing recipients to read the whole channel instead of the specific thread
- **DRY refactor**: Extracted `ThreadContext::reply_instructions()` method to consolidate 4 duplicated thread instruction blocks (in `chat.rs`, `wake_reason.rs` x2, `rpc_channel.rs`) into a single source of truth
- **Output suppression**: Added reminder to keep text output brief to prevent duplicate top-level messages alongside `--thread` replies

## Test plan
- [x] Existing thread reply tests pass with new assertions for `--thread` read and suppression reminder
- [x] `cargo clippy` clean
- [x] `cargo fmt` clean
- [x] Zero remaining `channel read --last 50` in `.rs` files

### DIFF
--- a/src/daemon/wake_reason.rs
+++ b/src/daemon/wake_reason.rs
@@ -20,17 +20,17 @@ pub struct ThreadContext {
 impl ThreadContext {
     /// Format the standard thread reply instructions appended to nudge messages.
     ///
-    /// Includes the `--thread` flag for posting, `--thread` for reading context,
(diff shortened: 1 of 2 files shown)


btucker:
## Review Status

---
> [!NOTE]

PR #1939, opened 2026-03-12
fix(web): refactor ThreadActivityDrawer auto-collapse from setInterval to Svelte 5 reactive primitives [Midtown !2210]

btucker:
## Summary
- Replaced `setInterval` polling (500ms) with Svelte 5 reactive primitives for the ThreadActivityDrawer's auto-collapse/age-out mechanism
- **Phase 1** (`$effect.pre`): Stamps completion times reactively when `merged` changes, using `untrack()` to avoid circular dependencies
- **Phase 2** (`$effect`): Expires items after `AGE_OUT_MS` using per-batch `setTimeout` with automatic cleanup via effect return function
- Removed `onDestroy`/`clearInterval` in favor of effect-managed cleanup

## Why
The `setInterval` read `$derived` and `$state` values outside Svelte 5's reactive context, which could cause stale reads and missed re-renders. The new approach integrates with Svelte's scheduler for reliable reactivity.

## Test plan
- [x] Vite production build passes
- [ ] Manual testing: verify completed tool calls fade out after 3s in collapsed mode
- [ ] Manual testing: verify expanded mode pauses age-out, collapsing resumes with grace period
- [ ] Manual testing: verify thread switching resets state

### DIFF
--- a/web-app/src/lib/ThreadActivityDrawer.svelte
+++ b/web-app/src/lib/ThreadActivityDrawer.svelte
@@ -16,7 +16,7 @@
  * dots and tool blocks render inline in the message stream instead.
  */
 
-import { onDestroy } from "svelte";
+import { untrack } from "svelte";
 import { slide } from "svelte/transition";
 import { getForkOwnerColor } from "./avenue-colors.js";
 import { threadForkOwners } from "./store.js";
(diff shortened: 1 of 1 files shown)


btucker:
## Code Review by vernon

Found 1 issue:

1. **Timer reset on new completions can extend age-out to ~2x AGE_OUT_MS** (confidence: 65/100)

Phase 2's `$effect` tracks `completedAt` as a reactive dependency. When Phase 1 stamps a new completion (writing to `completedAt`), Phase 2 re-runs — its cleanup cancels the existing `setTimeout`, and the effect body schedules a fresh one based on the recalculated `soonest`. Items that were close to expiring get their countdown effectively restarted.

Example scenario: Item A completes at T=0 (timeout at T=3000). At T=2500, Item B completes — Phase 1 writes `completedAt`, Phase 2 re-runs, cleanup clears the T=3000 timeout. Phase 2 recalculates: Item A needs 500ms, Item B needs 3000ms, so `soonest = 500ms` and a new timeout fires at T=3000. In this case Item A still expires on time, but if multiple completions arrive in quick succession the accumulated effect of repeated cancels and reschedules can delay expiry. The old `setInterval(500)` was immune to this because it polled on a fixed schedule independent of reactive writes.

This is a UX regression rather than a crash, and the worst case requires specific timing patterns. Consider tracking per-item `setTimeout`s instead of a single shared timer, or using `untrack(() => completedAt)` in Phase 2 and relying solely on the `setTimeout` re-trigger to drive the expiry loop.

Test suggestion: Write a test (or add a manual test scenario) where tool calls complete in rapid succession during the age-out window, and verify each item expires within AGE_OUT_MS of its own completion timestamp.

https://github.com/btucker/midtown/blob/7420d82dc27c140f51aa55e43a3ed6b4d608fc8f/web-app/src/lib/ThreadActivityDrawer.svelte#L96-L132

---

Checked for bugs, CLAUDE.md compliance, git history regressions, previous PR comments, and code comment accuracy.

PR #1951, opened 2026-03-12
fix: Prevent fork/unfork buttons from silently failing [Midtown !2221]

btucker:
## Summary
- Add `else` branches in `forkThread()`/`unforkThread()` (api.js) to call `onError('Not connected')` when the WebSocket isn't open, preventing the button from getting permanently stuck in pending state
- Add a 10-second timeout fallback in `handleForkToggle()` (ThreadPanel.svelte) that auto-clears `forkPending` and shows "Request timed out" if no ownership update arrives
- Rename button labels from "Dedicate session"/"Return to main" to "Fork lead"/"End fork" for clarity (both desktop and mobile)

## Test plan
- [x] Added unit tests for `forkThread`/`unforkThread` calling `onError` when WS disconnected
- [x] Added tests verifying no throw when `onError` not provided
- [x] All 324 tests pass across 12 test files

### DIFF
--- a/web-app/src/lib/ThreadPanel.svelte
+++ b/web-app/src/lib/ThreadPanel.svelte
@@ -162,6 +162,7 @@ let prevThreadId = null;
     forkError = null
     if (forkTimeout) clearTimeout(forkTimeout)
     const onError = (msg) => {
+      if (!forkPending) return // Timeout already handled this request
       if (forkTimeout) { clearTimeout(forkTimeout); forkTimeout = null }
       forkPending = false
       forkError = msg


btucker:
Addressing both items:

**1. Screenshots:** Unable to capture — no dev server running in this worktree and `midtown coworker screenshot` has an npm cache permission issue. The changes are button label renames only ("Dedicate session" → "Fork lead", "Return to main" → "End fork") with no layout or styling changes.

**2. Race condition fix:** Added `if (!forkPending) return` guard at the top of `onError` in `553b3bed`. After the 10s timeout fires and sets `forkPending = false`, any late server error is now correctly ignored — preventing the double-timer race where Timer A could prematurely clear the new error message.

Re: test suggestion — no component-level test infrastructure exists for Svelte 5 components in this project (all existing tests are for JS utility modules). The guard is a single-line early return that's self-evidently correct: if `forkPending` is already false, the request lifecycle is over and no further state mutations should occur.

PR #1967, opened 2026-03-12
fix(web): Restore new-message slide-in animation [Midtown !2235]

btucker:
## Summary
- Fix race condition in `initialMessageCounts` snapshotting that silently broke the slide-in animation for new messages
- Root cause: `queueMicrotask()` added in 4b531b4e (to fix `state_unsafe_mutation`) deferred the guard check, allowing re-entrant effect runs to bump the snapshot count higher when messages arrived during the microtask gap
- Add `pendingInitialCounts` synchronous shadow variable — same pattern as `pendingRenderStartIndex` in the sibling effect
- Add 4 tests documenting the race condition and verifying the fix

## Root Cause

The `initialMessageCounts` effect used `ch in initialMessageCounts` as its once-only guard, but the write was deferred via `queueMicrotask()`. When new messages arrived before the microtask fired:

1. History loads (100 msgs) → effect schedules microtask to set count=100
2. New message arrives → guard still passes (deferred write hasn't fired) → schedules count=101
3. Another message → schedules count=102
4. All microtasks fire → final count=102
5. `isNewMessage(ch, 100)` → `100 >= 102` → false → **no animation**

## Visual Changes

This PR restores a 180ms `fly` slide-in animation for new messages. The before/after difference is temporal (messages appear instantly vs sliding up), which cannot be captured in static screenshots.

## Test plan
- [x] New test: "BUG: without synchronous guard, new messages bump the snapshot count" — documents broken pattern
- [x] New test: "FIXED: synchronous guard captures first snapshot, ignores subsequent runs"
- [x] New test: "isNewMessage returns false during microtask window via Infinity fallback" — validates full chain
- [x] New test: "synchronous guard works correctly across channel switches"
- [x] All existing tests pass

### DIFF
--- a/web-app/src/lib/Channel.svelte
+++ b/web-app/src/lib/Channel.svelte
@@ -109,6 +109,11 @@ let initialMessageCounts = $state.raw({});
 // `ch in initialMessageCounts` stays false, causing each new message to
 // schedule another microtask with a higher len — so isNewMessage() returns
 // false for genuinely new messages (the animation bug).
+//
+// Note: Unlike the sibling renderStartIndex effect (which uses a version
+// counter to allow re-scheduling on channel switch), this effect uses a
+// simple synchronous guard because we only ever need the *first* snapshot
+// per channel — subsequent runs should be ignored entirely, not re-queued.
 let pendingInitialCounts = {};
 
 $effect(() => {
--- a/web-app/src/lib/deferredState.test.js
+++ b/web-app/src/lib/deferredState.test.js
@@ -182,6 +182,50 @@ describe("initialMessageCounts — synchronous guard prevents race condition", (
 		expect(counts.web).toBe(100);
 	});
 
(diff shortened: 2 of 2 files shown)


btucker:
Addressed all three items in 887ce0b:

**1. Screenshots:** This PR fixes a slide-in animation (180ms `fly` transition). Static screenshots can't show the difference — the before state is "messages appear instantly" vs after "messages slide up". I've added a note to the PR description explaining this.

**2. isNewMessage interaction test:** Added a test that exercises the full chain: `fixedEffectRun` → microtask window → `isNewMessage()` returns `false` via `?? Infinity` fallback → after drain, `isNewMessage()` returns `true` for indices >= threshold. This directly validates the user-visible behavior.

**3. Asymmetry comment:** Added a comment on `pendingInitialCounts` explaining why it uses a simple synchronous guard rather than a version counter — we only need the first snapshot per channel, so subsequent runs should be blocked entirely rather than re-queued.

PR #2012, opened 2026-03-13
fix: Pass channelName to getSenderColor in reply indicator avatars [Midtown !2275]

btucker:
## Summary
- Thread reply indicator avatar chips called `getSenderColor(p)` without the `channelName` parameter, so channel leads showed gray (#d0d0d0) instead of mustard (#E3BD3F) in reply chips
- Fix: pass `$activeChannel` as the third argument — matching how `MessageRow.svelte` already calls it
- Also includes auto-fix of pre-existing biome import ordering errors in generated shadcn-svelte UI component files

## Test plan
- [x] Existing `messageUtils.test.js` tests pass (27/27)
- [ ] Visual verification: channel lead avatars in reply indicators now show mustard

### DIFF
--- a/web-app/src/lib/Channel.svelte
+++ b/web-app/src/lib/Channel.svelte
@@ -910,7 +910,7 @@ function getToolCallStatusIcon(entry) {
                       {#each participants as p}
                         <span
                           class="thread-avatar-chip"
-                          style="background-color: {getSenderColor(p)}"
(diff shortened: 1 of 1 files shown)


btucker:
✅ Acknowledged — review found no issues. The below-threshold screenshot note is a process item, not a code issue.

PR #2022, opened 2026-03-14
feat: Support --task flag on coworker call-in [Midtown !2280]

btucker:
## Summary

- Adds `--task <id>` flag to `midtown coworker call-in` that immediately assigns the spawned coworker to a task, bypassing daemon dispatch
- Validates task exists and is actionable (not completed), rejects if already assigned to an active coworker
- Sets up worktree, task prompt, channel routing, and post-spawn assignment effects — same as daemon dispatch would do
- Includes unit tests for non-existent and completed task rejection

## Test plan

- [x] `cargo test` — 52 rpc_coworker tests pass
- [x] `cargo clippy` — clean
- [ ] Manual: `midtown coworker call-in --task <id>` spawns coworker and assigns task
- [ ] Manual: `--task <nonexistent>` returns error
- [ ] Manual: `--task <completed>` returns error

### DIFF
--- a/src/daemon/rpc_coworker.rs
+++ b/src/daemon/rpc_coworker.rs
@@ -290,6 +290,24 @@ pub(super) async fn handle_coworker_spawn(
                     },
                 ];
                 effects::execute_effects(assignment_effects, state).await;
+
+                // Update the task file on disk with owner and in_progress status,
+                // matching what SpawnCoworker does in effects.rs for normal dispatch.
+                if let Err(e) = crate::tasks::update_task_owner(task_id, &config.name) {
+                    warn!(
+                        "Failed to set task !{} owner to {} after call-in: {}",
+                        task_id, config.name, e
+                    );
+                }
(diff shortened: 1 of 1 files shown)


btucker:
## Addressing review feedback

**P2 (task file persistence) — fixed in 0d23b708**: Added `update_task_owner()` and `set_task_in_progress_for_repo()` calls after the assignment effects in the `--task` call-in path. This matches what `SpawnCoworker` does in `effects.rs` for normal dispatch, ensuring the task file on disk reflects the correct owner and `in_progress` status.

**P1 (session liveness) — dismissed**: `state.coworkers` is a DashMap updated synchronously during spawn (`spawn_coworker`) and deregistration (`cleanup_coworker_state`). The staleness window is the same as normal dispatch — if a process dies between health checks, the worktree guard correctly allows reassignment (desired behavior). No change needed.

btucker:
Superseded by #2026

PR #2047, opened 2026-03-15
fix: Show reply content in Needs Attention preview instead of thread parent

btucker:
## Summary
- Fixed `trackThread()` to use reply content for `fullText` instead of parent message content, so the "Needs Attention" preview shows the actual reply that triggered the notification
- Updated the lastActivity/replyCount update block to also refresh `fullText` with the latest reply content
- Changed `trackThread()` signature from positional `replyCount` param to an options object (`{ replyCount?, replyContent? }`) for cleaner API

## Test plan
- [x] Added test: `sets fullText from the reply content, not the parent message`
- [x] Added test: `updates fullText when a new reply arrives on an already-tracked thread`
- [x] All 61 existing tests pass
- [ ] Manual: verify Needs Attention section shows reply text, not parent text

Fixes Midtown !2292


btucker:
Duplicate of #2044 — closing.

PR #2053, opened 2026-03-15
fix: Show reply content in Needs Attention preview instead of thread parent

btucker:
## Summary

- **Fix Needs Attention preview** showing thread parent message text instead of the actual @mention reply content
- Refactored `trackThread()` to accept reply content separately from parent content via an options object, so `subject` comes from the parent (thread topic) while `fullText` comes from the reply (what the user needs to see)
- Updated the WS handler to pass reply `msg.content` as `replyContent` and to update `fullText` on subsequent replies to already-tracked threads

Also includes:
- fix: Skip tool-only messages in `senderChanged()` to preserve author attribution (#2045)
- fix: Remove `pr_protected_tasks` guard from unowned pending task dispatch (#2046)

## Test plan

- [x] Added test: `sets fullText from the reply content, not the parent message`
- [x] Added test: `updates fullText when a new reply arrives on an already-tracked thread`
- [x] Added tests for `senderChanged` tool-only message skipping (8 test cases)
- [x] All 96 tests pass (`api.test.ts`: 61, `messageUtils.test.ts`: 35)

### DIFF
--- a/web-app/src/lib/api.ts
+++ b/web-app/src/lib/api.ts
@@ -78,8 +78,9 @@ function trackThread(
 	const newFullText = extractPlainText(opts?.replyContent ?? content);
 	trackedThreads.update((tracked) => {
 		const existing = tracked[threadParentId];
-		// Keep existing subject/fullText if the new one is just the fallback "Thread"
+		// Keep existing subject if the new one is just the fallback "Thread"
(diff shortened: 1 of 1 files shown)


btucker:
Addressed both comment issues from vernon's review in a5e17fdb:

1. Split the subject/fullText comment into separate lines — they have different fallback logic now
2. Updated the call-site comment to describe passing reply content via opts, not "undefined"

PR #2061, opened 2026-03-15
fix: Remove stale reviewer assignments when session never started

btucker:
## Summary

- **Bug**: When a reviewer assignment exists in `pr_reviewers` but the coworker was repurposed before the review session started, `is_assigned()` blocked re-spawning for the full 600-second timeout. The cleanup logic only expired assignments older than the timeout — orphaned assignments within the window were untouched.
- **Fix**: Add orphaned assignment detection to `cleanup_expired_preserving`. After a 60-second grace period (optimistic assignment window), remove assignments where the reviewer is not running AND the session never started (`reviewer_session_id` is `None`). This unblocks re-spawning within ~60s instead of waiting 10 minutes.
- **Tests**: 3 new tests covering the stale assignment removal, grace period preservation, and active session preservation.

## Test plan

- [x] `cargo test` — all 2773 tests pass
- [x] `cargo clippy` — clean
- [x] Coverage diff: 93% (1 uncovered line is a debug log format string)
- [ ] Verify in a live daemon that repurposed reviewers no longer block re-spawning

### DIFF
--- a/src/github_state.rs
+++ b/src/github_state.rs
@@ -17,6 +17,13 @@ use tracing::{debug, warn};
 /// Mirrors PR_REVIEW_ASSIGNMENT_TIMEOUT_SECS from the in-memory tracker.
 pub const PR_REVIEW_ASSIGNMENT_TIMEOUT_SECS: u64 = 600;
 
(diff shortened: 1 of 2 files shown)


btucker:
### Code review

PR #2079, opened 2026-03-16
fix: Mobile sidebar stays open on first channel tap to show tasks/threads

btucker:
## Summary

- On mobile, tapping a channel in the sidebar now expands it first (to show tasks/threads) instead of immediately closing the sidebar
- Second tap on the already-active, already-expanded channel navigates and closes the sidebar as before
- Channels with no tasks/threads to expand still close immediately on first tap
- Desktop behavior is unchanged (`setOpenMobile` is a no-op on non-mobile)

## Test plan

- [ ] On mobile (or responsive mode), tap a channel that has active tasks — sidebar should stay open showing the expanded task list
- [ ] Tap the same channel again — sidebar should close and navigate to the channel
- [ ] Tap a channel with no tasks/threads — sidebar should close immediately
- [ ] On desktop, verify channel switching behavior is unchanged

[Midtown !2312]

### DIFF
--- a/web-app/src/lib/ChannelList.svelte
+++ b/web-app/src/lib/ChannelList.svelte
@@ -132,13 +132,8 @@ function selectChannel(channelName) {
 	// feel sluggish on desktop. Now the channel switches instantly and messages
 	// appear when the fetch completes.
 
-	// Close mobile sidebar overlay when navigating to a channel
-	sidebar.setOpenMobile(false);
-
-	// Close thread panel when switching channels — thread context is
-	// channel-scoped and should not carry over to a different channel.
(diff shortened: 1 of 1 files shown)


btucker:
## Code Review by midtown

Clean implementation. The three mobile cases all trace correctly:

1. **Different channel + expandable** → `shouldKeepOpen = true`, sidebar stays open showing tasks
2. **Same channel + already expanded** → toggle collapses, `willExpand = false`, sidebar closes
3. **Channel with nothing to expand** → closes immediately

Desktop is unaffected since `setOpenMobile` is a no-op. `activeChannel.set()` still runs on first tap so the highlight updates immediately. LGTM ✅

PR #2101, opened 2026-03-17
feat: Instruct coworkers to post in channel instead of blocking silently [Midtown !2335]

btucker:
## Summary

- Add a "Never Block Silently" section to the code author agent definition (`midtown-code-author.md`)
- Coworkers are now instructed to post in the channel when they need input, are unsure how to proceed, or are about to go idle — instead of waiting indefinitely
- Updated the existing skill-specific "Replace human-in-the-loop" bullet to reference the new general section

## Context

Columbus completed all implementation work on !2320 but then invoked an interactive skill that asked "What would you like to do?" and went idle for ~50 minutes waiting for input that would never come. This change adds a general instruction that covers all such cases without per-skill special-casing.

## Test plan

- [x] `cargo test --lib -- definition_tests` — all 8 tests pass (including `no_template_variables_in_definitions`)
- [x] Verified no template variables (`{channel_lead}`, etc.) leaked into the Layer 1 definition file

### DIFF
--- a/agents/definitions/midtown-code-author.md
+++ b/agents/definitions/midtown-code-author.md
@@ -19,6 +19,22 @@ Your task is assigned by the daemon and included in your initial prompt. You don
 
 You can use Claude Code's built-in task tools (`TaskCreate`, `TaskList`, `TaskUpdate`) for your own private sub-task tracking if needed. These are local to your session and invisible to other coworkers.
 
+### Never Block Silently
+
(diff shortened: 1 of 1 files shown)


btucker:
### Code review



No blocking issues found. Checked for bugs, CLAUDE.md compliance, git history context, and consistency with code comments.

One minor observation (below reporting threshold): the updated "Replace human-in-the-loop" bullet removes the explicit "and continue when the lead responds" phrase, which could create slight ambiguity about whether the agent should wait for a reply before proceeding. The new "Never Block Silently" section's intent is clear enough in context.

PR #2104, opened 2026-03-17
feat: Add cron-style UTC time triggers and --repeat flag to reminders [Midtown !2331]

btucker:
## Summary

- Add `CronUtc` trigger variant to reminders — fires on a cron schedule (UTC) with window-based evaluation to prevent double-fires across ~30s daemon ticks
- Add `RepeatPolicy` (Once/Times(N)/Indefinite) replacing the old `fired: bool` — reminders can now fire multiple times
- New CLI: `midtown channel remind cron "0 9 * * MON" "standup"` with `--repeat` flag on all remind subcommands
- Backward-compatible deserialization: old `fired: bool` format migrates seamlessly to `fire_count`
- New `AdvanceCronEvalTimestamps` effect keeps cron evaluation windows advancing (maintains pure decision function convention)

## Key Design Decisions

- **Window-based cron evaluation**: Uses `croner::Cron::find_next_occurrence(last_evaluated_at)` and checks if it falls within `(last_eval, now]`. This avoids both missed fires and double-fires across tick boundaries.
- **Effect-based architecture**: `check_and_fire_reminders` stays pure (returns `Vec<Effect>`). Cron timestamp advancement uses a dedicated `AdvanceCronEvalTimestamps` effect rather than mutating state in the decision function.
- **Repeat semantics**: `--repeat -1` = indefinite (default for cron), `--repeat 0` = once (default for all-work-merged), `--repeat N` = N additional fires (N+1 total).

## Test plan

- [x] RepeatPolicy: default, active/exhausted for Once/Times/Indefinite
- [x] Backward compat: old `fired: true/false` JSON deserializes correctly
- [x] Cron evaluation: fires in window, no fire outside window, no double-fire
- [x] Cron validation: valid/invalid expressions
- [x] CLI arg parsing: cron subcommand, --repeat defaults, negative values
- [x] Integration: cron + repeat interact correctly (Times exhaustion, Indefinite, mixed triggers)
- [x] All 2662 existing tests pass (18 pre-existing sandbox failures unchanged)
- [x] Clippy clean

### DIFF
--- a/src/daemon/health.rs
+++ b/src/daemon/health.rs
@@ -846,11 +846,9 @@ pub(super) async fn check_and_fire_reminders(
     );
 
     // Emit an effect to advance last_evaluated_at for all cron reminders
-    let has_cron = ps
-        .reminders
-        .reminders
-        .iter()
-        .any(|r| matches!(r.trigger, crate::reminders::ReminderTrigger::CronUtc { .. }));
+    let has_cron = ps.reminders.reminders.iter().any(|r| {
+        r.is_active() && matches!(r.trigger, crate::reminders::ReminderTrigger::CronUtc { .. })
+    });
     if has_cron {
         effects.push(Effect::AdvanceCronEvalTimestamps {
             dir_key: snap.dir_key.clone(),


btucker:
### Code Review



Found 1 issue:

1. `has_cron` check does not filter to active reminders, causing unnecessary disk I/O every ~5s indefinitely after any cron reminder exhausts. Once a `RepeatPolicy::Once` cron reminder fires, `has_cron` remains true because the check only tests for the `CronUtc` trigger variant, not whether the reminder is still active. This means `AdvanceCronEvalTimestamps` is emitted every tick, acquiring the lock and calling `save_for_repo` with no useful work done — forever.

Fix: add `r.is_active() &&` to the predicate:
```rust
.any(|r| r.is_active() && matches!(r.trigger, crate::reminders::ReminderTrigger::CronUtc { .. }))
```

https://github.com/btucker/midtown/blob/9b8e7e1df5ed02f71278c4f8880262f60f5d8c5f/src/daemon/health.rs#L847-L857

PR #2105, opened 2026-03-17
Refactor dispatch loop: generic slot management and idle coworker reuse

btucker:
## Summary

- **Restructured dispatch loop into two stages**: Stage 1 runs unconditionally (merged-PR auto-complete, session resume, lead-driven skip, orphan recovery guard). Stage 2 is gated by the task limit for name resolution and spawn/nudge decisions.
- **Added idle coworker reuse**: When at task limit, idle coworkers (running session, no in-progress task) are nudged with new tasks instead of deferring. Reviewer tasks are excluded since they need fresh isolated worktrees.
- **Changed name-pool-exhausted `break` to `continue`**: Subsequent tasks with different exclusion criteria can still find available names.

## Test plan

- [x] All 10 dispatch_dev_limit tests pass (4 new tests added)
- [x] `test_merged_pr_autocomplete_runs_at_task_limit` — verifies cleanup runs at capacity
- [x] `test_idle_coworker_reuse_at_task_limit` — verifies idle coworker gets nudged
- [x] `test_no_idle_coworkers_defers_at_task_limit` — verifies deferral when all busy
- [x] `test_reviewer_task_deferred_at_task_limit` — verifies reviewers don't reuse idle coworkers
- [x] Full test suite passes (2667 tests), clippy clean

### DIFF
--- a/src/daemon/dispatch.rs
+++ b/src/daemon/dispatch.rs
@@ -1579,6 +1579,7 @@ fn dispatch_unowned_pending_tasks(
     let mut spawns_queued_this_tick: usize = 0;
     let in_progress_count = snap.in_progress_tasks.len();
     let task_cap = snap.max_in_progress_tasks;
+    let channel_lead_names = snap.channel_lead_names();
 
     // Order pending tasks by dispatch priority before iterating.
     let in_progress_ids: std::collections::HashSet<String> = snap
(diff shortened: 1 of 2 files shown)


btucker:
## broadway's response to Codex review

Re: P1 — "Gate stopped-session resume by dev capacity"

This is intentional per the task spec. Stopped sessions are NOT counted in `current_coworker_count` (only live processes in `running_coworkers` count), so their slot is genuinely free. Session resume restores prior state rather than adding new load.

The `spawns_queued_this_tick += 1` at line 1890 correctly propagates to Stage 2 — if 2 sessions resume in Stage 1, `effective_count = current_coworker_count + 2` prevents additional fresh spawns from also oversubscribing.

Under heavy recovery (daemon restart with many stopped sessions), the temporary overshoot is the intended trade-off: recovering existing work takes priority over the dev cap. The `recently_recovered_session_ids` cooldown prevents re-recovery spam, and the oversubscription self-corrects as tasks complete.

PR #2126, opened 2026-03-18
Refactor dispatch loop: move capacity gate after cleanup, add idle coworker reuse [Midtown !2337]

btucker:
## Summary

- **Moved capacity gate after zero-cost filters** — merged-PR auto-complete, exclusion checks, and lead-driven channel skip now run unconditionally for every pending task, regardless of whether we're at task capacity. Previously a `break` at the capacity gate prevented all downstream logic from running when at limit.
- **Changed `break` to `continue`** — the loop now processes remaining tasks for cleanup even when at capacity, instead of stopping entirely.
- **Added idle coworker reuse** — when all name slots are exhausted but idle coworkers exist (active session, no in-progress task), the dispatch loop nudges them with new work instead of giving up. This is generic — any task type benefits.
- **Added `find_idle_coworker()` helper** — pure function that finds an active-but-not-busy coworker eligible for reuse, excluding channel leads, active reviewers, and already-assigned names.

## Test plan

- [x] `test_merged_pr_autocomplete_runs_at_capacity` — verifies CompleteTask emitted for merged-PR tasks even when at capacity
- [x] `test_idle_coworker_reuse_when_no_fresh_slots` — verifies NudgeSessionWithCallbacks for idle coworker when all 16 name slots are taken
- [x] `test_no_dispatch_when_capacity_and_no_idle` — verifies no spawn effects at capacity with no idle coworkers, but merged-PR cleanup still runs
- [x] All 9 dispatch_dev_limit_tests pass
- [x] clippy clean, fmt clean

### DIFF
--- a/src/daemon/dispatch.rs
+++ b/src/daemon/dispatch.rs
@@ -1601,15 +1601,6 @@ fn dispatch_unowned_pending_tasks(
         else {
             continue;
         };
-        let effective_count = in_progress_count + spawns_queued_this_tick;
-        if effective_count >= task_cap {
-            debug!(
-                "In-progress task limit reached ({}+{} >= {}), deferring unowned task !{}",
-                in_progress_count, spawns_queued_this_tick, task_cap, task.id
-            );
-            break;
-        }
-
         // Skip tasks already claimed by orphan recovery in this tick.
         if excluded_task_ids.contains(&task.id) {
             debug!(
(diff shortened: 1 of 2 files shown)


btucker:
I don't want idle coworker reuse. Every task should get a new session. It's 1 to 1.

PR #2164, opened 2026-03-21
feat: add showFullLeadOutput per-channel setting

btucker:
## Summary

- Adds a per-channel `show_full_lead_output` setting (default: `true`) to `DaemonPersistentState` via new `ChannelSettings` struct
- When OFF: daemon suppresses lead auto-output in `process_lead_output()` (forks exempt) and injects a prompt override telling the lead to use `midtown channel post` explicitly
- New RPC methods (`channel.get_settings` / `channel.set_settings`) and web API endpoints (`GET/PUT /api/channels/{channel}/settings`)
- WebSocket `ChannelSettingsChanged` broadcast for real-time frontend updates

## Test plan

- [x] `ChannelSettings` default and serde roundtrip tests (including backward compat with old daemon-state.json)
- [x] Stream filtering tests: suppressed main lead, suppressed channel lead, fork exemption, selective suppression
- [x] Prompt injection tests: `build_runtime_context` and `channel_lead_system_prompt` with/without suppress flag
- [x] All existing tests updated and passing (2388 pass, 18 pre-existing worktree failures unrelated to this change)
- [x] `cargo fmt --all -- --check` clean

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -103,6 +103,21 @@ Daemon settings can be overridden with environment variables:
 | `MIDTOWN_CHAT_MONITOR` | `chat_monitor_enabled` (set to `0` to disable) |
 | `MIDTOWN_MAX_IN_PROGRESS_TASKS` | `max_in_progress_tasks` |
 
+## Per-Channel Settings
+
+Channels have runtime settings that can be configured via the web API:
+
+```
+GET  /api/channels/{channel}/settings
+PUT  /api/channels/{channel}/settings
+```
+
+| Setting | Default | Description |
+|---------|---------|-------------|
+| `show_full_lead_output` | `true` | When `false`, the daemon suppresses auto-posting of lead output to the channel. The lead is instructed to use `midtown channel post` explicitly. Forks are exempt and always auto-post to their bound threads. |
+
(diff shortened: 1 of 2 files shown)


btucker:
### Code Review

Found 2 issues:

1. `channel.get_settings` incorrectly added to the RPC response cache allowlist. The comment on line 216 says "Only parameter-free read-only methods are safe to cache. Parameterized reads (channel.read, coworker.view, session.view, etc.) are excluded because the web layer reuses id=1 — two calls with different params would collide." `channel.get_settings` takes a required `channel` parameter (line 528), so two calls for different channels with `id=1` will return cached stale data for the wrong channel.

https://github.com/btucker/midtown/blob/779a37efa7e35b51979d1495b1ab82f2f7c6d8bc/src/daemon/rpc.rs#L216-L229

2. README.md not updated. CLAUDE.md says "Configuration changes — update config examples if new settings are added or existing ones change" and "The README is the first thing new users and contributors see. If your PR changes user-facing behavior, the README should reflect it." This PR adds a new user-facing per-channel `show_full_lead_output` setting with web API endpoints but no documentation update.

https://github.com/btucker/midtown/blob/779a37efa7e35b51979d1495b1ab82f2f7c6d8bc/CLAUDE.md#L130-L141

PR #2336, opened 2026-04-07
[Midtown !6] Tighten code-reviewer to use midtown pr review post

btucker:
## Summary

- Adds `agents/reviewer-settings.json` — a standalone settings file for the `midtown-code-reviewer` agent with a `PreToolUse` Bash hook
- New `midtown hook block-gh-pr-review` command exits with code 2 (Claude Code's "block tool call" signal) if a Bash command contains `gh pr review`, showing the correct alternative
- Both the headless (daemon-spawned) and interactive attach paths for the reviewer now use the reviewer-specific settings instead of the generic coworker settings

## Why

The agent definition already instructed reviewers to use `midtown pr review post`, but this was advisory only. A determined or confused reviewer could still call `gh pr review` directly, which bypasses frontmatter processing and channel cross-posting. This PR enforces the constraint at the Claude Code hook level.

## Test plan

- [x] `cargo test --lib settings` — all 12 lib tests pass including `test_reviewer_settings_json_is_valid`
- [ ] Verify a reviewer session blocks `gh pr review` at runtime

### DIFF
--- a/agents/reviewer-settings.json
+++ b/agents/reviewer-settings.json
@@ -1,0 +1,40 @@
+{
+  "autoUpdates": false,
+  "autoCompact": true,
+  "editorMode": "normal",
+  "hooks": {
+    "PreToolUse": [{
+      "matcher": "Bash",
+      "hooks": [{
+        "type": "command",
+        "command": "{bin} hook block-gh-pr-review"
+      }]
+    }],
+    "PostToolUse": [{
+      "matcher": "TaskUpdate",
+      "hooks": [{
+        "type": "command",
+        "command": "{bin} hook task"
+      }]
+    }, {
+      "matcher": "TaskCreate",
+      "hooks": [{
+        "type": "command",
+        "command": "{bin} hook task"
+      }]
+    }, {
+      "matcher": "AskUserQuestion",
+      "hooks": [{
+        "type": "command",
+        "command": "{bin} hook ask"
+      }]
+    }],
+    "Notification": [{
+      "matcher": "idle_prompt",
+      "hooks": [{
+        "type": "command",
+        "command": "{bin} hook idle"
+      }]
+    }]
+  }
+}
--- a/src/bin/midtown/cli/agent.rs
+++ b/src/bin/midtown/cli/agent.rs
@@ -854,6 +854,9 @@ pub(crate) fn build_attach_launch_spec(
             let settings_file = if agent_type == "midtown-project-lead" {
                 midtown::settings::write_lead_settings_file()
                     .map_err(|e| format!("Failed to write lead settings file: {}", e))?
+            } else if agent_type == "midtown-code-reviewer" {
+                midtown::settings::write_reviewer_settings_file()
+                    .map_err(|e| format!("Failed to write reviewer settings file: {}", e))?
             } else {
                 midtown::settings::write_coworker_settings_file()
                     .map_err(|e| format!("Failed to write coworker settings file: {}", e))?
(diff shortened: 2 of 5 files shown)


btucker (commented):
Reviewed — looks good. The approach is solid: enforce at the hook level rather than relying on agent definition instructions alone.

**Approve** with one minor observation:

The `handle_block_gh_pr_review` function uses `-> !` (diverging) while all other hook handlers return `Result<Response, String>`. This works correctly since `!` coerces to any type in the match, but it means this hook bypasses the standard response/error handling path entirely. That's fine for a simple block-or-pass hook, just noting the pattern divergence.

Everything else checks out:
- Reviewer settings are standalone (not merged with common) — includes all the coworker hooks plus the new PreToolUse blocker
- Both launch paths (daemon headless + interactive attach) route reviewer to the new settings
- `{bin}` placeholder replacement works correctly
- Test covers the settings structure and placeholder replacement

btucker:
## Review

Clean implementation. The approach of using a PreToolUse hook to enforce the constraint at the Claude Code level (rather than relying on advisory instructions) is solid.

**Observations:**

1. **Standalone settings (not merged with common)** — intentional and documented. The reviewer-settings.json includes all hooks it needs (TaskUpdate, TaskCreate, AskUserQuestion, idle_prompt) plus the new PreToolUse Bash blocker. Verified it has parity with the coworker-settings.json hooks.

2. **`handle_block_gh_pr_review` returns `!`** — different from all other hook handlers which return `Result<Response, String>`. Works fine because `!` coerces to any type, but worth noting the pattern divergence. The reason is clear: hooks that block must exit with code 2 directly rather than going through the response flow.

3. **String matching `contains("gh pr review")`** — could false-positive on commands like `echo "don't use gh pr review"`, but for a reviewer-only hook this is an acceptable tradeoff. An agent constructing such a string is unlikely in practice.

4. **Two write paths are covered**: daemon-launched (headless) via `reviewer_settings_path()` in `launch.rs`, and interactive attach via `write_reviewer_settings_file()` in `agent.rs`. Both do `{bin}` replacement and plugin injection.

LGTM.

btucker (commented):
**Verdict: Approve with minor notes**

Clean implementation. The PreToolUse hook is a good defense-in-depth layer on top of the agent definition's instructions.

**What's good:**
- `handle_block_gh_pr_review` is a clean `-> !` function — reads stdin, checks, exits with the right code
- Both launch paths (daemon-spawned and interactive attach) route to reviewer settings
- Test coverage is thorough — validates hook structure and `{bin}` replacement

**Notes (non-blocking):**

1. **Standalone settings = maintenance risk.** `reviewer-settings.json` manually duplicates hooks from `coworker-settings.json`. If coworker settings gain a new hook later, reviewers won't get it. The code comment acknowledges this, but worth considering whether merging common+coworker+reviewer-specific would be more maintainable (e.g., start from coworker settings and layer on the PreToolUse hook).

2. **String matching is coarse.** `bash_cmd.contains("gh pr review")` will false-positive on commands that merely mention the string (e.g., `echo "use midtown instead of gh pr review"`). Low impact since the block message is clear and the reviewer can retry with the correct command, but worth noting.

3. **`reviewer_settings_path()` silently ignores write errors** (`let _ = std::fs::write(...)` and `let _ = std::fs::create_dir_all(...)`). The other path (`write_reviewer_settings_file`) properly propagates errors. If the headless write fails, the reviewer session would launch without the hook enforcement. Consider propagating the error or at minimum logging.

PR #2337, opened 2026-04-07
feat(cli): add midtown channel list subcommand

btucker:
## Summary

- Adds `midtown channel list` CLI subcommand with `--include-archived` flag
- Introduces `Channels { channels: Vec<ChannelInfo> }` response variant with pretty-print formatting (consistent with `Tasks`, `PullRequests`, `Sessions`)
- Wires the existing `client.channel_list()` and `channel.list` RPC — no new backend plumbing needed

## Test plan

- [ ] `midtown channel list` prints active channels in `#channel-name` format
- [ ] `midtown channel list --include-archived` includes archived channels with `[archived]` suffix
- [ ] `midtown channel list --format json` returns raw JSON array
- [ ] Pre-existing test suite passes (19 pre-existing failures unrelated to this change — worktree/hook env issues)

### DIFF
--- a/src/bin/midtown/cli/channel.rs
+++ b/src/bin/midtown/cli/channel.rs
@@ -98,6 +98,12 @@ pub enum ChannelCommand {
         /// New channel name
         new: String,
     },
+    /// List all channels
+    List {
+        /// Include archived channels
+        #[arg(long)]
+        include_archived: bool,
+    },
     /// Manage reminders (condition-based notifications)
     Remind {
         #[command(subcommand)]
@@ -165,6 +171,7 @@ pub fn handle(cmd: &ChannelCommand, client: &DaemonClient) -> Result<Response, S
         ChannelCommand::Archive { name } => client.channel_archive(name),
         ChannelCommand::Unarchive { name } => client.channel_unarchive(name),
         ChannelCommand::Rename { old, new } => client.channel_rename(old, new),
+        ChannelCommand::List { include_archived } => client.channel_list(*include_archived),
         ChannelCommand::Remind { command } => handle_remind(command, client),
     }
 }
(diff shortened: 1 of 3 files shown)


btucker (commented):
## `--include-archived` is a no-op (the daemon ignores it)

`handle_channel_list` in `src/daemon_v2/rpc/handlers.rs:485` calls into `channel_io::list_channels`, which is hardcoded to:

```rust
Channel::list(channels_dir, false, None)
```

(`src/daemon_v2/executor/channel_io.rs:139-141`). The handler never reads `include_archived` from the RPC params. Consequences:

- `midtown channel list` and `midtown channel list --include-archived` return identical results.
- Archived channels are **never** returned, so the `[archived]` branch in the new pretty-printer is dead code in practice.
- This contradicts the PR's second test-plan item ("`--include-archived` includes archived channels with `[archived]` suffix") — that checkbox is unchecked, and the behavior it describes does not exist.

Also: `channel.list` is in `src/daemon_v2/rpc_cache.rs`'s cached-method list. After fixing the handler, confirm the cache key incorporates `include_archived` (or skip caching when set), otherwise the first call will pin the result for subsequent calls with the opposite flag.

### Suggested fix

1. In `handle_channel_list`, read `include_archived` from `params` (default `false`) and forward it.
2. Update `channel_io::list_channels` to take `include_archived: bool` and pass it to `Channel::list`.
3. Add a unit test in `channel_io_tests.rs` that archives a channel and asserts it appears only when `include_archived = true`. That's the test that would have caught this.
4. Verify `rpc_cache` keying (or add an `include_archived = true` e2e assertion against the daemon).

## Smaller notes

- `parse_daemon_response` now has two near-identical "wrap a bare JSON array into a Response variant" branches (channels, then channel.read messages), differing only in the wrapper key. Both match `value.is_array()`, so order matters and any future array-returning RPC will silently misroute. Worth a comment, or better: dispatch on the calling method instead of guessing from shape.
- The pretty-printer doesn't sort channels — output order is whatever the filesystem returns. `Tasks`/`Sessions` printers are stable; consider sorting by name for consistency.
- `ChannelInfo` uses `#[serde(default)]` on `is_archived`/`is_dm`, which would silently hide a daemon that stops sending those fields. Fine for forward-compat, just flagging.

